PR #1, opened 2018-04-24
Escape arguments to curl

chrismwendt:
Previously, whitespace in the GraphQL input could get compressed away.

curl commands now look like this:

```
curl \
   -H 'Authorization: token abcdef' \
   -d \{\"query\":\"query {  currentUser { username } }\",\"variables\":\{}} \
   sourcegraph.com/.api/graphql
```

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,6 +2,12 @@
 
 
 [[projects]]
+  branch = "master"
+  name = "github.com/kballard/go-shellquote"
+  packages = ["."]
+  revision = "cd60e84ee657ff3dc51de0b4f55dd299a3e136f2"
+
+[[projects]]
   name = "github.com/mattn/go-isatty"
   packages = ["."]
   revision = "0360b2af4f38e8d38c7fce2a9f4e702702d73a39"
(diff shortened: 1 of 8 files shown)


slimsag:
Question about your solution here: prior to this change running this command:

```
echo 'query($query: String!) {                                                                                                                                   
  search(query: $query) {
    results {
      resultCount
    }
  }
}' | src api -vars '{"query": "abc"}' --get-curl 'query=abc=foo' 'foo=bar'
```

Would produce:

```
curl \
   -H 'Authorization: token <redacted>' \
   -d '{"query":"query($query: String!) { search(query: $query) { results { resultCount } } }","variables":{"foo":"bar","query":"abc=foo"}}' \
   https://sourcegraph.com/.api/graphql
```

Now it produces (very long line, see horizontal scrollbar):

```
curl \
   -H 'Authorization: token <redacted>' \
   -d '{"query":"query($query: String!) {                                                                                                                                   \n  search(query: $query) {\n    results {\n      resultCount\n    }\n  }\n}\n","variables":{"foo":"bar","query":"abc=foo"}}' \
   https://sourcegraph.com/.api/graphql
```

chrismwendt:
Yeah, unfortunately there is no golang implementation of GraphQL minification (there's a Node.js one https://github.com/rse/graphql-query-compress).

There appears to be a bunch of trailing whitespace on the first line of your query, in particular.

slimsag:
Yeah actually you are right, I had missed the case where someone might have a literal `"   whitespace   "` in their query which would get minified away by my naive implementation. This is much more correct

PR #2, opened 2018-04-24
Remove positional vars

chrismwendt:
Previously there were 2 ways of specifying variables: positional and JSON.

You can still provide vars in JSON format:

```
echo '<query>' | src api -vars '{"var1":"val1, "var2":"val2"}'
```

### DIFF
--- a/cmd/src/api.go
+++ b/cmd/src/api.go
@@ -25,7 +25,7 @@ Examples:
 
   Specify query variables:
 
-    	$ echo '<query>' | src api 'var1=val1' 'var2=val2'
+    	$ echo '<query>' | src api -vars '{"var1":"val1, "var2":"val2"}'
 
   Searching for "Router" and getting result count:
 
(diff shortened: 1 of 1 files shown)


slimsag:
The reason I had added these in is the hope that it would make composing variables from within scripts easier. I'm not sure if it is the best way to do it, so I'd like to hear your feedback. For example, the type of thing I'd like to make easier is writing a script that performs a search:

```
search.sh 'repo:^github.com/gorilla/mux$ error'
```

To write this script with the JSON args, I must do a lot of escaping (`src api -vars="{\"query\": \"$1\"}"`):

```
echo 'query($query: String!) {
	search(query: $query) {
		results {
			resultCount
		}
	}
}' | src api -vars="{\"query\": \"$1\"}"
```

In contrast, with the positional arguments I can write:

```
... | src api "query=$1"
```

What do you think?

chrismwendt:
I just realized that you'd have to json-escape the argument `$1` in the script if you wanted to put it in the `-vars` flag, so that's probably a no-go. Let's leave it as-is.

slimsag:
Good point 👍

PR #4, opened 2018-05-08
Add `src repos list|enable|disable` commands

slimsag:
This PR adds a new top-level command `src repos` (alias: `src repo`) with three new subcommands: `src repos list`, `src repos enable` and `src repos disable`.

# Examples

For command-specific examples see the help flag (`src repos list -h`). In this section I'll write examples of how you can combine these for useful effects.

### List all repositories that are disabled on Sourcegraph

The following will list *all* repositories on Sourcegraph that are disabled, while excluding enabled ones:

```bash
src repos list -first=-1 -disabled -enabled=false
```

(add ` -query='github.com/slimsag/'` to match a query string if you only want a subset, or pipe to grep)

### Enable all disable repositories on Sourcegraph matching a query string

The following will enable any disabled repository on Sourcegraph whose name matches the query string `github.com/slimsag/`

```bash
src repos list -first=-1 -disabled -enabled=false -query='github.com/slimsag/' | xargs src repos enable
```

# Other notes/changes

- `src repos list` takes a variety of arguments, see `src repos list -h` for info.
    - By default, I've set the `-first` limit flag to 1000 to avoid doing too much intensive work by default. Many users may want to specify `-1` for this if they truly want all repositories.
- `src repos enable` and `src repos disable` take a list of repositories to enable/disable, e.g. `src repos enable github.com/my/repo1 github.com/my/repo2`. See `src repos enable -h` for more information.

- A number of refactors have been. Best review this PR as individual commits.
- Added support for nested subcommands (like `src repos list`, whereas before we only had one level deep `src api`) and command aliases (`src repo` is an alias for `src repos`)
- Added support for other non-`src api` commands to invoke the GraphQL API (needed for basically everything we'll want to add in the future).
- Added proper error checking for invalid HTTP responses with non-200 status codes.
- `src api` and other commands now emit a separate exit code `2` any time the actual request went through but GraphQL responded with `{data: ?, errors: ["some errors"]}`. This will let people easily detect GraphQL errors from e.g. bash scripts.

### DIFF
--- a/cmd/src/api.go
+++ b/cmd/src/api.go
@@ -17,6 +17,12 @@ import (
 
 func init() {
 	usage := `
+Exit codes:
+
+  0: Success
+  1: General failures (connection issues, invalid HTTP response, etc.)
+  2: GraphQL error response
+
 Examples:
 
   Run queries (identical behavior):
@@ -35,7 +41,6 @@ Examples:
   Get the curl command for a query (just add '-get-curl' in the flags section):
 
     	$ src api -get-curl -query='query { currentUser { username } }'
-
 `
 
 	flagSet := flag.NewFlagSet("api", flag.ExitOnError)
@@ -97,7 +102,8 @@ Examples:
 				fmt.Println(string(f))
 				return nil
 			},
-			flags: apiFlags,
+			flags:            apiFlags,
+			dontUnpackErrors: true,
 		}).do()
 	}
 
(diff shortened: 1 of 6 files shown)


slimsag:
> I'm not seeing the examples for src repos enable:

They are under `src repos enable -h`

> Am I not using the right subcommand?

Good catch on `enabled` and `disabled` not showing up under there, though. Fixed.

> Also, I'm a bit confused by how -disabled and -enabled interact, but I haven't spent much thought on how I expect it to work. We can refine it later if the need arises.

Let me know if the CLI help doesn't clarify this.

PR #8, opened 2018-08-02
add 'src search' command for using Sourcegraph search from the command line

### LINKED ISSUE #5: Document access token configuration in the README
Since everyone needs to do this, I think it should be documented in the README.

slimsag:
This adds a 'src search' subcommand for two primary use cases:

1. Searching from the command line (the default, `src search 'some query'`).
2. Getting results from our search API without having to specify all the GraphQL fields (via the json flag, `src search -json 'some query'`).

For usage information, consult `src search -h`:

```
Usage of 'src search':
  -explain-json
    	Explain the JSON output schema and exit.
  -get-curl
    	Print the curl command for executing this query and exit (WARNING: includes printing your access token!)
  -json
    	Whether or not to output results as JSON
  -less
    	Pipe output to 'less -R' (only if stdout is terminal) (default true)

Examples:

  Perform a search and get results:

    	$ src search 'repogroup:sample error'

  Perform a search and get results as JSON:

    	$ src search -json 'repogroup:sample error'

Other tips:

  Make 'type:diff' searches have colored diffs by installing https://colordiff.org
    - Ubuntu/Debian: $ sudo apt-get install colordiff
    - Mac OS:        $ brew install colordiff

  Disable color output by setting NO_COLOR=t (see https://no-color.org).

  Force color output on (not on by default when piped to other programs) by setting COLOR=t

  Query syntax: https://about.sourcegraph.com/docs/search/query-syntax/

```

Additionally, it also:

1. Cleans up some existing code.
2. Updates our README to accurately reflect the features this tool has.
3. Fixes #5 by no longer requiring access tokens except on instances that require them, and adding 'just in time' documentation for the user.


chrismwendt:
Wowww this is beautiful

image

PR #9, opened 2018-08-23
add extensions subcommands

sqs:
Adds the following subcommands for managing extensions:

- src extensions list
- src extensions get
- src extensions publish

These are marked as experimental. The API that they use is not yet deployed to Sourcegraph.com.

### DIFF
--- a/cmd/src/extensions_publish.go
+++ b/cmd/src/extensions_publish.go
@@ -6,6 +6,8 @@ import (
 	"flag"
 	"fmt"
 	"io/ioutil"
+	"os"
+	"os/exec"
 	"path/filepath"
 )
 
@@ -22,7 +24,8 @@ Examples:
           "name":      "myextension",
           "publisher": "alice",
           "title":     "My Extension",
-          "url":       "https://example.com/bundled-extension.js"
+          "main":      "dist/myext.js",
+          "scripts":   {"sourcegraph:prepublish": "parcel build --out-file dist/myext.js src/myext.ts"}
         }
     	$ src extensions publish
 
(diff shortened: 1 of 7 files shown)


cmd/src/extensions_publish.go
```diff
@@ -0,0 +1,156 @@
...
+
+  Publish the "alice/myextension" extension described by package.json in the current directory:
+
+    	$ cat package.json
+        {
+          "name":      "myextension",
+          "publisher": "alice",
+          "title":     "My Extension",
+          "url":       "https://example.com/bundled-extension.js"
+        }
```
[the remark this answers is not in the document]
sqs: Yep

cmd/src/extensions_publish.go
```diff
@@ -0,0 +1,156 @@
...
+		forceFlag       = flagSet.Bool("force", false, `Force publish the extension, even if there are validation problems or other warnings.`)
+		apiFlags        = newAPIFlags(flagSet)
+	)
+
+	handler := func(args []string) error {
+		flagSet.Parse(args)
+
+		manifest, err := ioutil.ReadFile(*manifestFlag)
+		if err != nil {
+			return fmt.Errorf("%s\n\nRun this command in a directory with a %s file for an extension.\n\nSee 'src extensions %s -h' for help.", err, *manifestFlag, flagSet.Name())
```
slimsag: Return a `usageError` here instead, and it'll print the help output which is useful here: https://github.com/sourcegraph/src-cli/blob/056c528f8828f7cfbedb4e1b7a7572bb31ccea7f/cmd/src/search.go#L60

Also, what if they just specified the wrong `-manifest` path file? It seems like odd wording in this case. Maybe check if `-manifest` is the default and change wording based on that.

sqs:
Thanks! Will change to ext.

cmd/src/extensions_publish.go
slimsag: Ah, got it, okay. LGTM then

PR #10, opened 2018-09-01
feat: add extensions delete subcommand

sqs:


cmd/src/extensions_delete.go
```diff
@@ -0,0 +1,73 @@
...
+`
+
+	flagSet := flag.NewFlagSet("delete", flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src extensions %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		extensionIDFlag = flagSet.String("id", "", `The ID (GraphQL API ID, not extension ID) of the extension to delete.`)
```
slimsag: Long term, is it going to make sense to have this be the extension ID and not the GraphQL API ID (i.e. was it done this way for now just for a quicker implementation?)

If so, I would say make this something like `-api-id` so that it is future-proof without breaking backwards compat

sqs: This was done for consistency. Many other commands take -id for the GraphQL ID. In the case of extensions, it is confusing because there is such thing as the “extension ID” (this confusion doesn’t exist for the other subcommands).

I propose merging it like this and deciding whether to rename all of those -id flags to -api-id separately. (Feel free to do that!) It would be fairly easy to do this at any point later on because there are no cases where the flag’s arg would be ambiguous.

If you want to do that change right now, then go ahead and I will rename it now here too.

PR #11, opened 2018-09-01
feat: add config subcommand to get/list/edit config

sqs:
This adds the following subcommands:

- src config get (for seeing final merged configuration)
- src config edit (for updating or overwriting settings)
- src config list (for listing all configuration subjects)

### DIFF
--- a/cmd/src/config.go
+++ b/cmd/src/config.go
@@ -1,0 +1,186 @@
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+)
+
+var configCommands commander
+
+func init() {
+	usage := `'src config' is a tool that manages global, organization, and user settings on a Sourcegraph instance.
+
+The effective configuration is computed by shallow-merging the following settings, in order from lowest to highest precedence:
+
+- Global settings (site-wide)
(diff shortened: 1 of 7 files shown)


cmd/src/config.go
```diff
@@ -0,0 +1,186 @@
...
+  }
+}
+`,
+		result: &result,
+	}
+	if err := req.do(); err != nil {
+		return "", err
+	}
+	if result.CurrentUser == nil || result.CurrentUser.ID == "" {
+		return "", errors.New("unable to determine current user ID (ensure the access token you're using to authenticate the CLI is correct, or provide one if none exists; see 'src -h' for more information)")
```
slimsag: In other locations we link to the README for how to authenticate actually, which helps a bit with shortening the wording a bit. See https://gotools.org/github.com/sourcegraph/src-cli/cmd/src#api.go-L220

slimsag:
Awesome!!

cmd/src/config.go
sqs: Thanks, pushed a new commit that links to the README

PR #12, opened 2018-09-02
feat: add users tag subcommand to add/rm user tags, and listing users by tag

sqs:
User tags are used for internal Sourcegraph feature flags to enable certain features for specific users.


slimsag:
LGTM++

PR #21, opened 2018-10-05
Only include tag in the GraphQL query if provided

mdaniel:
fixes: sourcegraph/issues#273

### DIFF
--- a/cmd/src/users_list.go
+++ b/cmd/src/users_list.go
@@ -48,16 +48,26 @@ Examples:
 		if err != nil {
 			return err
 		}
-
+		vars := map[string]interface{}{
+			"first": nullInt(*firstFlag),
+			"query": nullString(*queryFlag),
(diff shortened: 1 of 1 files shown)


cmd/src/users_list.go
```diff
@@ -48,16 +48,26 @@ Examples:
...
+		queryTag := ""
+		if maybeTagVar, ok := vars["tag"].(*string); ok && maybeTagVar != nil {
+			queryTagVar = `$tag: String,`
+			queryTag = `tag: $tag,`
+		}
 		query := `query Users(
   $first: Int,
   $query: String,
-  $tag: String,
+` + queryTagVar + `
```
felixfbecker: I’m a bit confused why we must dynamically create query, as that adds complexity. Why can’t we just pass null?

sqs: Check the issue that this fixes linked from the PR description for why

PR #22, opened 2018-10-05
Allow environment overrides for all config knobs

mdaniel:
Including, naturally, the config location itself. This takes the opportunity to pull some magic strings up into constants to make their usage more obvious

This commit stops shy of wrapping the *132* character long help text line, because maybe that's the golang way, but it could arguably be improved, also

```
--- PASS: TestReadConfig (0.01s)
    --- PASS: TestReadConfig/pure_defaults (0.00s)
    --- PASS: TestReadConfig/accessToken_in_the_environment (0.00s)
    --- PASS: TestReadConfig/accessToken_in_the_-config_flag (0.00s)
    --- PASS: TestReadConfig/accessToken_in_a_file_identified_by_the_SRC_CONFIG_env (0.00s)
    --- PASS: TestReadConfig/accessToken_in_the_env,_superseding_-config_file (0.00s)
    --- PASS: TestReadConfig/accessToken_in_the_env,_superseding_SRC_CONFIG_file (0.00s)
    --- PASS: TestReadConfig/accessToken_in_the_env,_superseding_-config,_with_SRC_CONFIG_set (0.00s)
    --- PASS: TestReadConfig/accessToken_in_-config,_with_SRC_CONFIG_set (0.00s)
    --- PASS: TestReadConfig/endpoint_in_SRC_ENDPOINT (0.00s)
    --- PASS: TestReadConfig/endpoint_in_SRC_ENDPOINT,_superseding_-config (0.00s)
    --- PASS: TestReadConfig/endpoint_in_SRC_ENDPOINT,_and_-config,_-endpoint_wins (0.00s)
    --- PASS: TestReadConfig/endpoint_in_-config,_but_-endpoint_wins (0.00s)
    --- PASS: TestReadConfig/ensure_env-vars_do_not_leak (0.00s)
```

cc: @sqs as mentioned in the other thread
cc: @nicksnyder per our discussion over email (I am expecting there is more coming, but this certainly is something to discuss)

### DIFF
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -11,6 +11,14 @@ import (
 	"strings"
 )
 
+const (
+	AccessTokenEnvVar = "SRC_ACCESS_TOKEN"
+	ConfigEnvVar      = "SRC_CONFIG"
+	ConfigFilename    = "src-config.json"
+	DefaultEndpoint   = "https://sourcegraph.com"
+	EndpointEnvVar    = "SRC_ENDPOINT"
+)
+
 const usageText = `src is a tool that provides access to Sourcegraph instances.
 For more information, see https://github.com/sourcegraph/src-cli
 
@@ -21,7 +29,21 @@ Usage:
 The options are:
 
 	-config=$HOME/src-config.json    specifies a file containing {"accessToken": "<secret>", "endpoint": "https://sourcegraph.com"}
+      You can use "${}" syntax to reference environment variables, even on Windows,
+      but you may wish to use single-quotes to protect that syntax from shell expansion:
+
+      -config=${UserProfile}/src-config.json or
+      -config='${TMPDIR}/my-config.json'
+
+      [Environment Variables]
+      $SRC_CONFIG       can point to the config file
+                        although the -config option is always authoritative
+      $SRC_ACCESS_TOKEN can specify, or supersede, the access token
+
 	-endpoint=                       specifies the endpoint to use e.g. "https://sourcegraph.com" (overrides -config, if any)
+      [Environment Variables]
+      $SRC_ENDPOINT     can specify, or supersede, the value in -config
+                        although the -endpoint option is always authoritative
 
 The commands are:
 
@@ -61,16 +83,22 @@ type config struct {
 	AccessToken string `json:"accessToken"`
 }
 
-// readConfig reads the config file from the given path.
+// readConfig reads the config file from the given path, plus any in-scope environment variables
 func readConfig() (*config, error) {
 	cfgPath := *configPath
-	userSpecified := *configPath != ""
+	userSpecified := cfgPath != ""
 	if !userSpecified {
-		user, err := user.Current()
+		if cfgEnvPath := os.Getenv(ConfigEnvVar); cfgEnvPath != "" {
+			cfgPath = cfgEnvPath
+			userSpecified = true
+		}
+	}
+	if !userSpecified {
+		currentUser, err := user.Current()
 		if err != nil {
 			return nil, err
 		}
-		cfgPath = filepath.Join(user.HomeDir, "src-config.json")
+		cfgPath = filepath.Join(currentUser.HomeDir, ConfigFilename)
 	}
 	data, err := ioutil.ReadFile(os.ExpandEnv(cfgPath))
 	if err != nil && (!os.IsNotExist(err) || userSpecified) {
@@ -84,14 +112,18 @@ func readConfig() (*config, error) {
 	}
 
 	// Apply config overrides.
-	if envToken := os.Getenv("SRC_ACCESS_TOKEN"); envToken != "" {
+	if envToken := os.Getenv(AccessTokenEnvVar); envToken != "" {
 		cfg.AccessToken = envToken
 	}
-	if *endpoint != "" {
-		cfg.Endpoint = strings.TrimSuffix(*endpoint, "/")
+	userEndpoint := *endpoint
+	if envEndpoint := os.Getenv(EndpointEnvVar); userEndpoint == "" && envEndpoint != "" {
+		userEndpoint = envEndpoint
+	}
+	if userEndpoint != "" {
+		cfg.Endpoint = strings.TrimSuffix(userEndpoint, "/")
 	}
 	if cfg.Endpoint == "" {
-		cfg.Endpoint = "https://sourcegraph.com"
+		cfg.Endpoint = DefaultEndpoint
 	}
 	return &cfg, nil
 }
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -12,11 +12,11 @@ import (
 )
 
 const (
-	AccessTokenEnvVar = "SRC_ACCESS_TOKEN"
-	ConfigEnvVar      = "SRC_CONFIG"
-	ConfigFilename    = "src-config.json"
-	DefaultEndpoint   = "https://sourcegraph.com"
-	EndpointEnvVar    = "SRC_ENDPOINT"
+	accessTokenEnvVar = "SRC_ACCESS_TOKEN"
+	configEnvVar      = "SRC_CONFIG"
+	configFilename    = "src-config.json"
+	defaultEndpoint   = "https://sourcegraph.com"
+	endpointEnvVar    = "SRC_ENDPOINT"
 )
 
 const usageText = `src is a tool that provides access to Sourcegraph instances.
@@ -88,7 +88,7 @@ func readConfig() (*config, error) {
 	cfgPath := *configPath
 	userSpecified := cfgPath != ""
 	if !userSpecified {
-		if cfgEnvPath := os.Getenv(ConfigEnvVar); cfgEnvPath != "" {
+		if cfgEnvPath := os.Getenv(configEnvVar); cfgEnvPath != "" {
 			cfgPath = cfgEnvPath
 			userSpecified = true
 		}
@@ -98,7 +98,7 @@ func readConfig() (*config, error) {
 		if err != nil {
 			return nil, err
 		}
-		cfgPath = filepath.Join(currentUser.HomeDir, ConfigFilename)
+		cfgPath = filepath.Join(currentUser.HomeDir, configFilename)
 	}
 	data, err := ioutil.ReadFile(os.ExpandEnv(cfgPath))
 	if err != nil && (!os.IsNotExist(err) || userSpecified) {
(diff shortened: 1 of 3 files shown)


cmd/src/main.go
```diff
@@ -21,7 +29,21 @@ Usage:
 The options are:
 
 	-config=$HOME/src-config.json    specifies a file containing {"accessToken": "<secret>", "endpoint": "https://sourcegraph.com"}
+      You can use "${}" syntax to reference environment variables, even on Windows,
+      but you may wish to use single-quotes to protect that syntax from shell expansion:
+
+      -config=${UserProfile}/src-config.json or
+      -config='${TMPDIR}/my-config.json'
+
+      [Environment Variables]
```
nicksnyder: Couple thoughts:
1. It is hard to scan through and see all the args or environment variables. This might be solved with better indentation, formatting, or grouping (e.g. the `[Environment Variables]` looks like a "everything below here is related" heading, when it isn't). What if env variables were defined on the same line as the params?

2. The tests you have added are great! My concern though is that we need tests in the first place. Values can come from the config, a parameter, or an environment variable. Seems overly complex (and the fact that there is no parameter for accessToken feels inconsistent). I don't think you made any unreasonable decisions here because you were just extending what already existed, just the end state makes me feel a bit uncomfortable. Maybe another followup for @slimsag.

cmd/src/main.go
```diff
@@ -21,7 +29,21 @@ Usage:
 The options are:
 
 	-config=$HOME/src-config.json    specifies a file containing {"accessToken": "<secret>", "endpoint": "https://sourcegraph.com"}
+      You can use "${}" syntax to reference environment variables, even on Windows,
+      but you may wish to use single-quotes to protect that syntax from shell expansion:
```
nicksnyder: Huh, so you are just documenting behavior that already existed here, right?

I am a little bit confused why anyone would want to do this since by default the shell would already do this expansion, right? It is also a bit confusing that this only happens for the config variable.

Not something for you to worry about, but for @slimsag to follow up on.

cmd/src/main_test.go
```diff
@@ -0,0 +1,337 @@
...
+	if testCase.configJsonFlagContents != "" {
+		configFile := writeTempConfigFile(testCase.configJsonFlagContents, testCase.testState)
+		configFilename := configFile.Name()
+		oldConfigPath := configPath
+		configPath = &configFilename
+		cleanupFuncs = append(cleanupFuncs, func() error {
+			// cheat and piggy back on this cleanup
+			configPath = oldConfigPath
+
+			if e := os.Remove(configFilename); e != nil {
```
nicksnyder:
```go
return os.Remove(configFilename)
```

cmd/src/main_test.go
```diff
@@ -0,0 +1,337 @@
...
+		})
+	}
+
+	if testCase.configJsonEnvContents != "" {
+		configFile := writeTempConfigFile(testCase.configJsonEnvContents, testCase.testState)
+		configFilename := configFile.Name()
+		cleanup := pushSetenv(ConfigEnvVar, configFilename)
+		cleanupFuncs = append(cleanupFuncs, cleanup)
+		cleanupFuncs = append(cleanupFuncs, func() error {
+			if e := os.Remove(configFilename); e != nil {
```
nicksnyder:
```go
return os.Remove(configFilename)
```

cmd/src/main_test.go
mdaniel: Damn, you caught me napping on that one :-( it was fallout from some refactoring that moved the code but not the neurons. I'll straighten it out

cmd/src/main.go
mdaniel: That's right, and I did so because I would be gravely surprised by that behavior. The alternative would be to remove the [`os.ExpandEnv`](https://github.com/sourcegraph/src-cli/blob/3.4.0/cmd/src/main.go#L75) which is the source of that behavior and hope [no one's workflow depended on it](https://xkcd.com/1172/)

The counterbalance to "why would anyone want that" is that it _may_ harmonize some Windows behavior, since the env-expansion story is all over the place in that setup (`-config=%FOO%` if cmd, and then presumably `-config="${env:FOO}..."` if running under PS), but even under PS one would need to single-quote to avoid the dollar-brace from being consumed by PS itself

cmd/src/main.go
mdaniel: > My concern though is that we need tests in the first place

Heh, yup, that absolutely crossed my mind. And that underlying sentiment is actually why I said I was prepared for the wontfix -- I can deeply appreciate that this is a polarizing change -- but hey, it makes for great code challenge discussion, right?

But the reasoning breaks out like this:

* testing config permutations
* CI/CD pipelines
* ~OCD is a real thing~

It was turning into a huge PITA to test the sudo tokens work via: edit config, test, edit config, test, ... when there was a super tasty `$SRC_ACCESS_TOKEN` just sitting there taunting me.

Based on my experience reforming the [SourceClear Agent](https://www.sourceclear.com/docs/agent-usage/#environment-variables), the DevSecOps users we interfaced with found the ability to provide hierarchical config values extremely useful. So, the sentiment is not driven entirely by the whims of some random guy on the Internet :-)

And I readily admit just having one env-var was disrupting my sense of harmony.

> and the fact that there is no parameter for accessToken feels inconsistent

It is inconsistent, but purposefully so: it is a grave anti-pattern to even _allow_ users to put access tokens on the command line because ~laziness~ human nature implies that if something is possible, folks will do it and not _think_ about it. Command line credentials leak in two bad places: shell history and `ps -ef`. I am _relatively_ sure that one needs to be either the same uid or root to see env-var values from `ps`.

mdaniel: As for point number 1, I am 100% not wedded to any bit of that change other than I really believe strongly in giving users as much "tl;dr" as possible when they ask for `-h`. It has the very pleasing side-effect of meaning the docs stay next to the code, and it is a grave pet peeve of mine for `-h` to say something like "for more information, visit this almost certainly 404 webpage ..."

mdaniel:
Separately, perhaps also for @slimsag: GoLand whines that `dep ensure` is out of sync with `Gopkg.lock`; and sure enough, running `dep check` surfaces `# Gopkg.lock is out of sync:`

Is that something you'd like straightened out, also?

nicksnyder:
> Is that something you'd like straightened out, also?

Good catch, can you do that in a separate PR?

slimsag:
I can't review yet but saw godep mentioned, let's switch to go modules
instead

cmd/src/main.go
slimsag: Yes, the reason we `os.ExpandEnv` this string is because our docs imply that e.g. `-config=$HOME/src-config.json` will work, among many other situations where we suggest env vars and these won't work on Windows. The other option is to start providing windows-specific help docs, but that seems like a larger change that is unlikely to be maintained going forward. Our windows support is iffy, certainly, this just makes it an ounce nicer.

However, I do not think this needs to be documented behavior, honestly. I doubt anyone is going to be trying to pass a literal `$FOO` as part of the config file path. This feels like documenting something that nobody is ever going to run into.

Suggestion: Please remove this doc addition here OR add it to an extended help flag.

cmd/src/main.go
```diff
@@ -11,6 +11,14 @@ import (
 	"strings"
 )
 
+const (
+	AccessTokenEnvVar = "SRC_ACCESS_TOKEN"
+	ConfigEnvVar      = "SRC_CONFIG"
+	ConfigFilename    = "src-config.json"
+	DefaultEndpoint   = "https://sourcegraph.com"
+	EndpointEnvVar    = "SRC_ENDPOINT"
```
slimsag: Constants like these shouldn't be exported, they're not imported by other code external to this package.

cmd/src/main.go
slimsag: ### Parameter for accessToken

I think it is fine that there is no parameter for `accessToken`. As you stated already, it would be dangerous to provide one. I am strongly against providing one (and I suspect Nick is, too).

### Documentation

By and large the documentation additions in this PR are the number one thing I am opposed to. This adds a lot of information that I estimate 70% of people running `src -h` to get information will not use, and that is bad. When I see CLI tools that spew out a lot of information like this, I immediately think "I should Google how to achieve $X instead, it may be quicker"

I suggest one of two options:

1. Remove this documentation and instead just include a single line at the end stating something like the following (which effectively communicates the same information in far less words):

> Flags can also be specified via environment variables, e.g. -config via SRC_CONFIG.

2. Provide a `src -extended-help` flag, which could include this more detailed information. I am not convinced that we will have enough docs to put in this flag without it being (almost) useless, though, so I am far less in favor of this option.

### Should SRC_ACCESS_TOKEN be documented here?

I do not think it needs to be. The reason I say this is because this is (1) documented in the README and (2) we directly link to the authentication instructions in the README if you encounter an authentication error https://github.com/sourcegraph/src-cli/blob/master/cmd/src/api.go#L219-L221

### Testing

Yeah, we don't have any tests and that sucks. The codebase started out very small and is growing quickly, so it is not surprising.

I think the best way to test the entire CLI would be by building the binary (or refactoring the main command to be callable from tests) and then invoking it in tests against sourcegraph.com and verifying the expected output. But this is a separate / distinct task, and there is a lot we can test in isolation in unit tests without these sort of 'end to end' CLI tests.

I think the tests you added here are fine because any test > no test, but I didn't review them. Testing in this codebase is a bigger task to take on as I mentioned above, so I don't want to block this PR on figuring out what an ideal test suite setup looks like.

cmd/src/main.go
```diff
@@ -61,16 +83,22 @@ type config struct {
...
+		}
+	}
+	if !userSpecified {
+		currentUser, err := user.Current()
 		if err != nil {
 			return nil, err
 		}
-		cfgPath = filepath.Join(user.HomeDir, "src-config.json")
+		cfgPath = filepath.Join(currentUser.HomeDir, ConfigFilename)
 	}
```
slimsag: I think this readConfig function has now become significantly more complex with the addition of the env getters logic.

I think adding utility functions which return first the CLI flag, second the env var would simplify this a ton and likely make the changes to `readConfig` here almost none.

mdaniel:
Had I joined, we would have had a lot more ... discussion ... about your observations. But, with the way things are now, I did as you asked and removed the explanatory `-help` content.

mdaniel:
Transitioning to go modules is waaaaay outside the scope of my golang expertise, so I'll leave those warnings with you

mrnugget:

Issue #3, opened 2018-05-05
Add support for sudo tokens


slimsag:

mdaniel:
I have a Works For Me™ branch, and will put a PR in flight for this

burmudar:
Closing this issue as part of repository maintenance. We are no longer using GitHub issues to track src-cli requests or reports. For future questions or support requests, please contact Sourcegraph Support (mailto:support@sourcegraph.com).

PR #23, opened 2018-10-05
Add support for specifying a sudo token user

mdaniel:
Adds another top-level field to the `src-config.json` entitled "runAs" that serves as both the username as whom the request will run, and therefore also marks the access token as a sudo token. Currently, if `"runAs"` is provided but the token is not actually a sudo token, it will end pretty much as poorly as one might expect

I didn't add any additional sanity checking because the original code didn't have any, but I am cognizant that it appears the cli is currently using the deserialized `string` as is, surfacing header injection attacks if the config payload is untrusted.

fixes: #3

cc: @nicksnyder (less code than #22 but this one is an actual feature, and (heh) was requested, so that might help you more)

### DIFF
--- a/cmd/src/api.go
+++ b/cmd/src/api.go
@@ -138,7 +138,6 @@ func authenticationHeader(cfg *config) (string, string) {
 	} else {
 		headerValue = "token " + cfg.AccessToken
 	}
-
 	return "Authorization", headerValue
 }
 
--- a/cmd/src/api.go
+++ b/cmd/src/api.go
@@ -156,8 +156,7 @@ func curlCmd(cfg *config, query string, vars map[string]interface{}) (string, er
 
 	s := fmt.Sprintf("curl \\\n")
 	if cfg.AccessToken != "" {
-		var headerName, headerValue string
-		headerName, headerValue = authenticationHeader(cfg)
+		headerName, headerValue := authenticationHeader(cfg)
 		if headerName == "" {
 			return "", fmt.Errorf(authHeaderMissingAccessTokenError)
 		}
@@ -227,8 +226,7 @@ func (a *apiRequest) do() error {
 		return err
 	}
 	if cfg.AccessToken != "" {
-		var headerName, headerValue string
-		headerName, headerValue = authenticationHeader(cfg)
+		headerName, headerValue := authenticationHeader(cfg)
 		if headerName == "" {
 			return fmt.Errorf(authHeaderMissingAccessTokenError)
 		}
--- a/cmd/src/api.go
+++ b/cmd/src/api.go
@@ -56,7 +56,9 @@ Examples:
 	)
 
 	handler := func(args []string) error {
-		flagSet.Parse(args)
+		if err := flagSet.Parse(args); err != nil {
+			return err
+		}
 
 		// Build the GraphQL request.
 		query := *queryFlag
--- a/cmd/src/api.go
+++ b/cmd/src/api.go
@@ -123,8 +123,25 @@ func gqlURL(endpoint string) string {
 	return endpoint + "/.api/graphql"
 }
 
+// authenticationHeader returns the header name and header value required to authenticate a request,
+// or `"", ""` if one calls this function when the config has no AccessToken information
+func authenticationHeader(cfg *config) (string, string) {
+	if cfg.AccessToken == "" {
+		return "", ""
+	}
+
+	var headerValue string
+	if cfg.RunAs != "" {
+		headerValue = fmt.Sprintf("token-sudo user=\"%s\",token=\"%s\"", cfg.RunAs, cfg.AccessToken)
+	} else {
+		headerValue = "token " + cfg.AccessToken
+	}
+
+	return "Authorization", headerValue
+}
+
 // curlCmd returns the curl command to perform the given GraphQL query. Bash-only.
-func curlCmd(endpoint, accessToken, query string, vars map[string]interface{}) (string, error) {
+func curlCmd(cfg *config, query string, vars map[string]interface{}) (string, error) {
 	data, err := json.Marshal(map[string]interface{}{
 		"query":     query,
 		"variables": vars,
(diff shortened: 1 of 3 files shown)


cmd/src/api.go
```diff
@@ -134,11 +151,17 @@ func curlCmd(endpoint, accessToken, query string, vars map[string]interface{}) (
 	}
 
 	s := fmt.Sprintf("curl \\\n")
-	if accessToken != "" {
-		s += fmt.Sprintf("   %s \\\n", shellquote.Join("-H", "Authorization: token "+accessToken))
+	if cfg.AccessToken != "" {
+		var headerName, headerValue string
+		headerName, headerValue = authenticationHeader(cfg)
+		if headerName == "" {
+			return "", fmt.Errorf("unable to translate config into authentication header, please file an issue")
```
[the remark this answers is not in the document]
mdaniel: Yup, sorry, I am not in the habit of having an open source repo to point at; I'll straighten that out

cmd/src/api.go
```diff
@@ -123,8 +123,25 @@ func gqlURL(endpoint string) string {
...
+// authenticationHeader returns the header name and header value required to authenticate a request,
+// or `"", ""` if one calls this function when the config has no AccessToken information
+func authenticationHeader(cfg *config) (string, string) {
+	if cfg.AccessToken == "" {
+		return "", ""
+	}
+
+	var headerValue string
+	if cfg.RunAs != "" {
+		headerValue = fmt.Sprintf("token-sudo user=\"%s\",token=\"%s\"", cfg.RunAs, cfg.AccessToken)
```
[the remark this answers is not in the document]
mdaniel: Damn, you caught me napping again

cmd/src/api.go
```diff
@@ -123,8 +123,25 @@ func gqlURL(endpoint string) string {
 	return endpoint + "/.api/graphql"
 }
 
+// authenticationHeader returns the header name and header value required to authenticate a request,
+// or `"", ""` if one calls this function when the config has no AccessToken information
+func authenticationHeader(cfg *config) (string, string) {
```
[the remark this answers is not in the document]
mdaniel: I did actually consider that, but just because the HTTP header is `Authorization:` doesn't change the fact that the header is providing _authentication_ (unless there is some dark corner of RFC7644 I don't know about)

I'm certainly open to changing it if you feel it's distracting, but that's why it's that way

cmd/src/api.go
```diff
@@ -134,11 +155,16 @@ func curlCmd(endpoint, accessToken, query string, vars map[string]interface{}) (
 	}
 
 	s := fmt.Sprintf("curl \\\n")
-	if accessToken != "" {
-		s += fmt.Sprintf("   %s \\\n", shellquote.Join("-H", "Authorization: token "+accessToken))
+	if cfg.AccessToken != "" {
+		headerName, headerValue := authenticationHeader(cfg)
+		if headerName == "" {
+			return "", fmt.Errorf(authHeaderMissingAccessTokenError)
```
slimsag: I don't think returning this error is suitable here. This is a bug in our application so we should be panicking here.

```Go
panic("failed to create authentication header from configuration")
```

cmd/src/api.go
```diff
@@ -200,7 +226,11 @@ func (a *apiRequest) do() error {
 		return err
 	}
 	if cfg.AccessToken != "" {
-		req.Header.Set("Authorization", "token "+cfg.AccessToken)
+		headerName, headerValue := authenticationHeader(cfg)
+		if headerName == "" {
+			return fmt.Errorf(authHeaderMissingAccessTokenError)
```
slimsag: Same thing here.

cmd/src/main.go
```diff
@@ -20,7 +20,7 @@ Usage:
 
 The options are:
 
-	-config=$HOME/src-config.json    specifies a file containing {"accessToken": "<secret>", "endpoint": "https://sourcegraph.com"}
+	-config=$HOME/src-config.json    specifies a file containing {"accessToken": "<secret>", "endpoint": "https://sourcegraph.com", "runAs": "another-user"}
```
slimsag: I do not think that sudo tokens are really going to be used that often. Additionally, this makes it look like this needs to be specified always even when not using sudo tokens.

Instead, please add a section to the README for how to use sudo tokens.

cmd/src/api.go
```diff
@@ -123,8 +127,25 @@ func gqlURL(endpoint string) string {
...
+	}
+
+	var headerValue string
+	if cfg.RunAs != "" {
+		headerValue = fmt.Sprintf(`token-sudo user="%s",token="%s"`, cfg.RunAs, cfg.AccessToken)
+	} else {
+		headerValue = "token " + cfg.AccessToken
+	}
+
+	return "Authorization", headerValue
```
slimsag: nit: Remove blank line before return

cmd/src/api.go
```diff
@@ -123,8 +127,25 @@ func gqlURL(endpoint string) string {
 	return endpoint + "/.api/graphql"
 }
 
+// authenticationHeader returns the header name and header value required to authenticate a request,
+// or `"", ""` if one calls this function when the config has no AccessToken information
+func authenticationHeader(cfg *config) (string, string) {
```
slimsag: This function is _always_ returning just the `Authorization` header, and returning it as a value makes the code significantly more complex.

I think that a better solution would be for this to only return the _value_ of the `Authorization ` header.

mdaniel:
As an FYI, the ab8c6a8 change was pointed out by the latest EAP of GoLand, and it correctly observed that `defer resp.Body.Close()` and `fmt.Fprintf` also returned `error`s that were being swallowed. However, I felt more comfortable that checking `flagSet.Parse`'s would pass your review than fixing either of those other two. I purposefully put my head in the sand about all the other places in the codebase that might also be swallowing `error`.

cmd/src/api.go
mdaniel: For my curiosity, is this a Sourcegraph coding convention? I was sad that `go fmt ./...` didn't say anything about it

slimsag: Hmm, I don't think there is anything formal stating this anywhere. And gofmt doesn't have any formal rules about how code should be separate by whitespace AFAIK. You could e.g. have no blank lines in a function and gofmt wouldn't complain, or you could have tons and likewise it wouldn't complain AFAIK :P

My general rule of thumb is that if the return statement is a simple 1 liner, then no whitespace separating it from the code block above is useful.

You may also appreciate this site if you haven't seen it: https://dmitri.shuralyov.com/idiomatic-go

slimsag:
Yeah, arguably to make linters happy we should be underscore ignoring those errors. In both cases I think there is not anything we can realistically do (or would want to do) with the error, so they are intentionally ignored. Flag parsing error we should definitely handle though, nice catch!

mrnugget:

Issue #17, opened 2018-09-14
add ability to verify e.g. all user's emails en-mass


slimsag:

burmudar:
Closing this issue as part of repository maintenance. We are no longer using GitHub issues to track src-cli requests or reports. For future questions or support requests, please contact Sourcegraph Support (mailto:support@sourcegraph.com).

PR #24, opened 2018-10-05
Add verify-emails to allow mutating in mass

mdaniel:
There seemed to be precedent for using subshells to generate flag values, which was **especially** relevant for this since we need a list of tuples, so a JSON structure seemed like a reasonable choice.

fixes: #17

cc: @nicksnyder (this is the last one I had on my "I think I can do that" list, and hopefully we're getting to a pretty representative sample; it's my first time writing GraphQL, so be gentle :grin: )

### DIFF
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -16,11 +16,12 @@ Usage:
 
 The commands are:
 
-	list       lists users
-	get        gets a user
-	create     creates a user account
-	delete     deletes a user account
-	tag        add/remove a tag on a user
+	list          lists users
+	get           gets a user
+	create        creates a user account
+	delete        deletes a user account
+	tag           add/remove a tag on a user
+	verify-emails verify user emails in bulk
 
 Use "src users [command] -h" for more information about a command.
 `
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -21,7 +21,7 @@ The commands are:
 	create        creates a user account
 	delete        deletes a user account
 	tag           add/remove a tag on a user
-	verify-emails verify user emails in bulk
+	emails        actions related to user emails
 
 Use "src users [command] -h" for more information about a command.
 `
--- a/cmd/src/users_verify.go
+++ b/cmd/src/users_verify.go
@@ -1,0 +1,101 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"strings"
+)
+
+const SubCommand = "verify-emails"
+
+type VerifyEmailJson struct {
+	ID    string `json:"ID"`
+	Email string `json:"Email"`
+}
+
+func init() {
+	usage := `
+Examples:
+
+  Verify one or more user account(s):
+
+    	$ src users ` + SubCommand + ` -verify-json='[{"ID": "ID1", "Email": "email1"},...]'
+
+`
+
+	flagSet := flag.NewFlagSet(SubCommand, flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		verifyJsonFlag = flagSet.String("verify-json", "",
+			`The JSON array of id-email descriptors of those user/emails that should be marked as verified. (required)`)
+		apiFlags = newAPIFlags(flagSet)
+	)
+
+	handler := func(args []string) error {
+		flagSet.Parse(args)
+
+		dec := json.NewDecoder(bytes.NewBufferString(*verifyJsonFlag))
+
+		var verifyList []VerifyEmailJson
+		if err := dec.Decode(&verifyList); err != nil {
+			return err
+		}
+
+		vars := map[string]interface{}{}
+
+		// translate the provided index of `verifyList` into variables for its `id` and `email` payload
+		indexToKeys := func(idx int) (string, string) {
+			return fmt.Sprintf("id%09d", idx), fmt.Sprintf("email%09d", idx)
+		}
+
+		for idx, verifyObj := range verifyList {
+			idKey, emailKey := indexToKeys(idx)
+			vars[idKey] = verifyObj.ID
+			vars[emailKey] = verifyObj.Email
+		}
+
+		query := `mutation VerifyUserEmails(`
+		for k := range vars {
+			varType := "String!"
+			if strings.HasPrefix(k, "id") {
+				varType = "ID!"
+			}
+			query += fmt.Sprintf("$%s: %s,", k, varType)
+		}
+		query += ") {"
+		for idx := range verifyList {
+			idKey, emailKey := indexToKeys(idx)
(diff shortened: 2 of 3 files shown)


cmd/src/users_verify.go
```diff
@@ -0,0 +1,101 @@
...
+			vars[emailKey] = verifyObj.Email
+		}
+
+		query := `mutation VerifyUserEmails(`
+		for k := range vars {
+			varType := "String!"
+			if strings.HasPrefix(k, "id") {
+				varType = "ID!"
+			}
+			query += fmt.Sprintf("$%s: %s,", k, varType)
```
mdaniel: I observed that [`strings.Builder`](https://godoc.org/strings#Builder) is a thing, but it wasn't clear if the code noise is worth the reduction in memory copying; I welcome input about that

cmd/src/users_verify.go
```diff
@@ -0,0 +1,101 @@
...
+		apiFlags = newAPIFlags(flagSet)
+	)
+
+	handler := func(args []string) error {
+		flagSet.Parse(args)
+
+		dec := json.NewDecoder(bytes.NewBufferString(*verifyJsonFlag))
+
+		var verifyList []VerifyEmailJson
+		if err := dec.Decode(&verifyList); err != nil {
```
nicksnyder: it might be simpler to do 
```go
if err := json.Unmarshal([]byte(*verifyJsonFlag), &verifyList); err != nil {
}
```

cmd/src/users_verify.go
```diff
@@ -0,0 +1,100 @@
...
+			return fmt.Sprintf("id%09d", idx), fmt.Sprintf("email%09d", idx)
+		}
+
+		for idx, verifyObj := range verifyList {
+			idKey, emailKey := indexToKeys(idx)
+			vars[idKey] = verifyObj.ID
+			vars[emailKey] = verifyObj.Email
+		}
+
+		query := `mutation VerifyUserEmails(`
```
nicksnyder: all this string concat is a bit hard to follow. [text/template](https://golang.org/pkg/text/template/) might be a better choice.

cmd/src/users_verify.go
nicksnyder: Meh, only if it makes the code more readable or performance is sensitive. I haven't used it before personally. I think my previous comment about text/template is more important and would probably make this question go away.

cmd/src/users_verify.go
```diff
@@ -0,0 +1,100 @@
...
+		var verifyList []VerifyEmailJson
+		if err := dec.Decode(&verifyList); err != nil {
+			return err
+		}
+
+		vars := map[string]interface{}{}
+
+		// translate the provided index of `verifyList` into variables for its `id` and `email` payload
+		indexToKeys := func(idx int) (string, string) {
+			return fmt.Sprintf("id%09d", idx), fmt.Sprintf("email%09d", idx)
```
nicksnyder: This feels weird to me. I would probably prefer two separate maps here for readability but this comment might become obsolete if you use text/template as suggested in a different comment.

cmd/src/users.go
```diff
@@ -16,11 +16,12 @@ Usage:
...
-	get        gets a user
-	create     creates a user account
-	delete     deletes a user account
-	tag        add/remove a tag on a user
+	list          lists users
+	get           gets a user
+	create        creates a user account
+	delete        deletes a user account
+	tag           add/remove a tag on a user
+	verify-emails verify user emails in bulk
```
slimsag: If we do go with this approach, this should be done as an `emails` subcommand not as an `verify-emails` command. This enables us in the future to offer other things like:

- `src users emails add -username=alice alice@gmail.com`
- `src users emails remove -username=alice alice@gmail.com`

etc.

cmd/src/users_verify.go
```diff
@@ -0,0 +1,108 @@
...
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"text/template"
+)
+
+const SubCommand = "verify-emails"
+
+// input = `[]VerifyEmailJson`
```
slimsag: ?

cmd/src/users_verify.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"text/template"
+)
+
+const SubCommand = "verify-emails"
```
slimsag: It is better to hard-code this value inline.

Also, having the name be ambiguous like `SubCommand` means it would conflict with every other `cmd/src/*.go` file. And capitalized `SubCommand` means this is exported which is odd

cmd/src/users_verify.go
```diff
@@ -0,0 +1,108 @@
...
+
+// input = `[]VerifyEmailJson`
+// be careful if you change the vars format, since the vars need to match below
+const setEmailVerifiedTemplate = `
```
slimsag: You could declare this inline and then wouldn't need the comment

cmd/src/users_verify.go
```diff
@@ -0,0 +1,108 @@
...
+	Email string `json:"Email"`
+}
+
+func init() {
+	usage := `
+Examples:
+
+  Verify one or more user account(s):
+
+    	$ src users ` + SubCommand + ` -verify-json='[{"ID": "ID1", "Email": "email1"},...]'
```
slimsag: I apologize for not detailing this in the original issue, but in general with this CLI tool we prefer pipe-based commands over things like passing in JSON because it enables more powerful composition.

To be clear, I think the way this should work is:

Verify a single user's email:

```
src users emails verify -id='USERID'
```

Verify users en-mass:

```bash
src users list -query='myquery' -f='{{.ID}}' | xargs -n 1 -I USERID src users emails verify -id='USERID'
```

This matches the same way we e.g. tell users they can do mass user deletion:

https://github.com/sourcegraph/src-cli/blob/da78eacc42769eecb3d0b3789999dbeafdbff79b/cmd/src/users_delete.go#L20-L22

slimsag: This also simplifies our own code here and means e.g. you don't need to use the GraphQL template composition stuff here

cmd/src/users_verify.go
mdaniel: it was designed to document what `.` represents in the `text/template`, but if you want that huge blob of text inline in the `func` (:cry:) then I guess the documentation need falls away

cmd/src/users_verify.go
mdaniel: but AIUI that would also mean 250 round-trips to the API for verifying 250 users, right? OT1H, CPU and network are conceptually cheap, OTOH, each one of those requests is occupying a Sourcegraph instance that could be doing something else fun with those sockets.

I actually did strongly consider something akin to `src user emails verify -stdin` and letting them pipe JSON to stdin, but I didn't feel like pressing the issue

I'm \[unfortunately\] just a bystander now, so if you'd prefer that then I'll make it happen, but it feels hacky

mdaniel: Separately, just using `-id` is insufficient because a user can have multiple emails; now, don't get me wrong, maybe verifying them all cart blanc is the 99% most often case, but without the ability to specify the explicit email of the UID, you're going to jam up someone's process

mrnugget:

PR #25, opened 2018-10-11
add src repos delete command

ggilmore:
`src repos delete` allows you to delete one or more repositories from your sourcegraph instance. 

@slimsag I will revert my changes that add `multierrror` and `pkg/errors`. I had already written the file using those packages, so I thought I'd throw up a PR first.


cmd/src/repos_delete.go
```diff
@@ -0,0 +1,76 @@
...
+		repoID, err := fetchRepositoryID(repoName)
+		if err != nil {
+			return err
+		}
+
+		query := `mutation DeleteRepository($repoID: ID!){
+		deleteRepository(repository: $repoID) {
+			alwaysNil
+		}
+	}`
```
slimsag: Formatting / indention here seems wonky

cmd/src/repos_delete.go
```diff
@@ -0,0 +1,76 @@
...
+		flagSet.Parse(args)
+		var errs *multierror.Error
+		for _, repoName := range flagSet.Args() {
+			err := deleteRepository(repoName)
+			if err != nil {
+				err = errors.Wrapf(err, "Failed to delete repository %q", repoName)
+				errs = multierror.Append(errs, err)
+			}
+		}
+		return errs.ErrorOrNil()
```
slimsag: I am fine with using multierror / pkg/errors here. I didn't realize your change was so small in this regard.

But for consistency sake, could you update the `repos_disable_enable.go` file to also use this approach?

ggilmore:
@slimsag PTAL, updated

cmd/src/repos_delete.go
```diff
@@ -0,0 +1,75 @@
...
+		repoID, err := fetchRepositoryID(repoName)
+		if err != nil {
+			return err
+		}
+
+		query := `mutation DeleteRepository($repoID: ID!){
+   deleteRepository(repository: $repoID) {
+	   alwaysNil
+    }
+}`
```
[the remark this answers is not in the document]
slimsag: It is not. :(

PR #29, opened 2018-10-28
remove elapsedMillisecond hack now that backend is fixed

lukeautry:
See https://github.com/sourcegraph/sourcegraph/pull/551 for details. Bad value for elapsedMilliseconds should not be coming back from the server on new versions.

Not sure how releases work so I'll leave that part up to you guys.

### DIFF
--- a/cmd/src/search.go
+++ b/cmd/src/search.go
@@ -205,13 +205,6 @@ Other tips:
 					searchResults:       result.Search.Results,
 				}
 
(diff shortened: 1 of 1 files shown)


slimsag:
Thanks :) I'll wait one Sourcegraph version, i.e. when 3.0 is officially released, this'll be merged. This should give folks enough time to update.

slimsag:
Enough time has passed that most instances will have this field working now, merging. There is a separate convo to have about versioning src alongside Sourcegraph

PR #33, opened 2018-12-12
Use new search result interface

attfarhan:
This PR makes commit, diff, and repo search use fields from the generic search result interface merged in https://github.com/sourcegraph/sourcegraph/pull/998. This change is backwards compatible, and only affects results if the Sourcegraph instance connected to the CLI has a build date after Dec 10, 2018.


felixfbecker:
Do you have a screenshot of before and after?

attfarhan:
@felixfbecker 

**Commit search**
_Before_:
image

After:
image


**Diff search**
Before:
image

After:
image

Repo results look identical.

slimsag:
Above screenshots LGTM, I filed https://github.com/sourcegraph/sourcegraph/issues/1366 for you to follow up on some minor things though

cmd/src/config.json
```diff
@@ -0,0 +1 @@
+{"accessToken": "d2086ad0b4f2464647fa38fdc3dedd54f3361dc5", "endpoint": "http://localhost:7080"}
```
slimsag: This shouldn't be in your PR

felixfbecker: @attfarhan 	make sure to revoke that token asap

attfarhan: This was for my local instance. Revoked it though.

cmd/src/search.go
```diff
@@ -299,6 +334,115 @@ func searchHighlightPreview(preview interface{}, start, end string) string {
...
+		// Replace our end-of-match token with the color terminator,
+		// and start all colors that were previously started to the left.
+		line = strings.Replace(line, uniqueEndOfMatchToken, ansiColors["nc"]+strings.Join(left, ""), -1)
+
+		final = append(final, line)
+	}
+	return strings.Join(final, "\n")
+}
+
+func renderResult(searchResult map[string]interface{}) string {
```
slimsag: should declare this inline where it is called (only one call location)

cmd/src/search.go
```diff
@@ -299,6 +334,115 @@ func searchHighlightPreview(preview interface{}, start, end string) string {
...
+	text, err := html2text.FromString(string(renderedHTML), html2text.Options{OmitLinks: true})
+	text = html.UnescapeString(text)
+	if err != nil {
+		return input
+	}
+	return text
+}
+
+// Checks the Sourcegraph instance's build version date to determine if the new search result interface exists.
+func buildVersionHasNewSearchInterface(buildVersion interface{}) bool {
```
slimsag: Make `buildVersion` a `string` here. It should 'just work' still and prevents anyone else from calling this func from passing a wrong argument type.

cmd/src/search.go
```diff
@@ -299,6 +334,115 @@ func searchHighlightPreview(preview interface{}, start, end string) string {
...
+		if isDiff {
+			// We special case diffs because we want to display them with color.
+			return searchHighlightDiffPreview(map[string]interface{}{"value": plainText, "highlights": highlights.([]interface{})})
+		}
+		return searchHighlightPreview(map[string]interface{}{"value": plainText, "highlights": highlights.([]interface{})}, "", "")
+	}
+	return text
+}
+
+func markdownToPlainText(input string) string {
```
slimsag: We talked in person and agreed this makes more sense as the server rendering HTML and us using html2text here. (without rendering the markdown on the src-cli)

cmd/src/search.go
```diff
@@ -299,6 +334,115 @@ func searchHighlightPreview(preview interface{}, start, end string) string {
...
+	renderedHTML := blackfriday.Run([]byte(input))
+	text, err := html2text.FromString(string(renderedHTML), html2text.Options{OmitLinks: true})
+	text = html.UnescapeString(text)
+	if err != nil {
+		return input
+	}
+	return text
+}
+
+// Checks the Sourcegraph instance's build version date to determine if the new search result interface exists.
```
slimsag: Add a test for this function (i.e. call the func with a good build version and a bad build version and check the return value)

cmd/src/search.go
```diff
@@ -299,6 +334,115 @@ func searchHighlightPreview(preview interface{}, start, end string) string {
...
+func buildVersionHasNewSearchInterface(buildVersion interface{}) bool {
+	dateRegex, err := regexp.Compile("(\\w{4}-\\w{2}-\\w{2})")
```
slimsag: Declare this regex as a package-level variable and use `regexp.MustCompile`. This is better because this function gets called _a lot_ and compiling regex isn't free

cmd/src/testdata/search_formatting/basic-commit-new.test.json
```diff
@@ -0,0 +1,183 @@
+{
```
slimsag: You have accidently comitted a `cmd/src/src` binary file to this repo, remove that

PR #41, opened 2019-05-10
Add support for programmatically excluding repos via external service config

slimsag:
This PR adds support for querying repositories by name (e.g. `/secretrepos/`) and passing that list of repositories into a command that adds them to the `exclude` list in the external service configuration.

For example, it is possible to write this script now to add all repositories with `/secretrepos/` in their name to the exclusion list of all external services:

```
# Determine which repositories we will exclude.
repos_to_exclude=$(src -config=./src-config.json repos list -first="-1" -names-without-host -query="/secretrepos/")

# Determine which external service configurations we need to update.
extsvcs=$(src -config=./src-config.json extsvc list -f '{{range .Nodes}}{{.id}} {{end}}')

# edit the configuration to exclude the repositories
echo "$extsvcs" | xargs -n1 -I {} src -config=./src-config.json extsvc edit -id '{}' -exclude-repos "$repos_to_exclude"
```

Previously we maintained backwards compatibility for `setRepositoryEnablement` in the GraphQL API which was used for this purpose, but because it resulted in updating the config once per repository it became VERY slow (unusable slow). This approach is much faster as you can add thousands at once instead of individually.

For https://app.hubspot.com/contacts/2762526/company/407948923

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,13 +58,13 @@ Consult `src -h` and `src api -h` for usage information.
 Some Sourcegraph instances will be configured to require authentication. You can do so via the environment:
 
 ```sh
-SRC_ACCESS_TOKEN="secret" src ...
+SRC_ENDPOINT=https://sourcegraph.example.com SRC_ACCESS_TOKEN="secret" src ...
 ```
 
 Or via the configuration file (`~/src-config.json`):
 
 ```sh
-	{"accessToken": "secret"}
+	{"accessToken": "secret", "endpoint": "https://sourcegraph.example.com"}
 ```
 
 See `src -h` for more information on specifying access tokens.
--- a/cmd/src/extsvc.go
+++ b/cmd/src/extsvc.go
@@ -1,0 +1,39 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+)
+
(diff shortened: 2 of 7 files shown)


keegancsmith:
FYI we support adding things like `{pattern: "/secretrepos/"}` to the exclude list for bitbucketserver now.

tsenart:
The values passed to  `-exclude-repos` must be in a specific format (i.e. not the computed Sourcegraph repo name).

For BitbucketServer: `{projectKey}/{repoSlug}`

See documentation for other code hosts.

slimsag:
> FYI we support adding things like {pattern: "/secretrepos/"} to the exclude list for bitbucketserver now.

Ack, I followed up on this over Slack with you.

> The values passed to -exclude-repos must be in a specific format (i.e. not the computed Sourcegraph repo name).
>
> For BitbucketServer: {projectKey}/{repoSlug}
>
> See documentation for other code hosts.

Yeah, I noticed this. `src repos list -names-without-host` (which was added in this PR) outputs repo names in this format which is sufficient for at least bitbucket/github/gitlab repos.

PR #44, opened 2019-09-20
Add `campaigns add-changesets` subcommand

mrnugget:
This adds the changesets with the given external IDs to the given
campaign.

### Example:

    $ src campaigns add-changesets -campaign-id Q2FtcGFpZ246NA== -repo-name github.com/sourcegraph/sourcegraph 5016 5017 5018
    Created 3 changesets.
    Added changeset to campaign. Changesets now in campaign: 10

### Current downsides

(These are mostly due to how our API works with batches and could be fixed on the API side, collecting these just here for documentation/transparency)

* When one of the changeset already exists, the command fails with:

```
[
  {
    "message": "pq: duplicate key value violates unique constraint \"changesets_repo_external_id_unique\"",
    "path": [
      "createChangesets"
    ]
  }
]
```

* When one of the external IDs specified is invalid (non existant, for example), it creates the rest silently but returns an error message for the single invalid one:

```
[
  {
    "message": "error in GraphQL response: Could not resolve to a PullRequest with the number of 5013.",
    "path": [
      "createChangesets"
    ]
  }
]
```

### DIFF
--- a/cmd/src/campaigns_add_changesets.go
+++ b/cmd/src/campaigns_add_changesets.go
@@ -12,17 +12,17 @@ Add changesets for a given repository to a campaign.
 
 Usage:
 
-	src campaigns add-changesets -campaign-id <id> -repo-name <name> [external changeset IDs...]
+	src campaigns add-changesets -campaign <id> -repo-name <name> [external changeset IDs...]
 
 Examples:
 
-  Add GitHub pull requests #5662 and #5366 for repository "github.com/sourcegraph/sourcegraph" to the Sourcegraph campaign with GraphQL ID "Q2FtcGFpZ246MQ=="
+  Add GitHub pull requests #5662 and #5366 for repository "github.com/sourcegraph/sourcegraph" to the Sourcegraph campaign with GraphQL ID "Q2FtcGFpZ246MQ==":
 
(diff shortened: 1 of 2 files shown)


tsenart:
> When one of the changeset already exists, the command fails with:

What do you think about checking the existing changesets of the given campaign first and reporting to the user exactly those that already exist? This isn't transactionally consistent (since things could change in between reading and writing), but it's improved UX, still safe (because of the database unique constraint) and it can be done here in the src-cli without changing backend code.

cmd/src/campaigns_add_changesets.go
```diff
@@ -0,0 +1,196 @@
...
+
+	return changesetIDs, err
+}
+
+const addChangesetsQuery = `
+mutation AddChangesetsToCampaign($campaign: ID!, $changesets: [ID!]!) {
+  addChangesetsToCampaign(campaign: $campaign, changesets: $changesets) {
+	id
+	changesets {
+      totalCount
```
tsenart: Looks funky.

felixfbecker:
With only this command the user would still have to use the API manually to find out the campaign ID. I think we also need a `campaigns get` command.

mrnugget:
> With only this command the user would still have to use the API manually to find out the campaign ID. I think we also need a `campaigns get` command.

That would list all campaigns then, I assume? I can implement that

felixfbecker:
Yeah, I'd imagine it would list the campaigns with ID, namespace, title (at minimum)

PR #57, opened 2019-10-18
LSIF: automatically generate LSIF upload token at upload time

chrismwendt:
This implements the [roadmap](https://docs.google.com/document/d/1cBsE9801DcBF9chZyMnxRdolqM_1c2pPyGQz15QAvYI/edit#heading=h.dimwsc9ccmwq) item:

> Solution to authorize LSIF data uploads on sourcegraph.com that has good UX and doesn’t feel hacky (e.g. GitHub action).

This runs through the challenge/verification flow on the fly at upload time:

- Get topics
- Get the challenge from Sourcegraph
- Set topics to old+challenge
- Ask Sourcegraph to verify and return a token
- Restore old topics

GitHub actions already have a GITHUB_TOKEN environment variable set https://help.github.com/en/articles/virtual-environments-for-github-actions#github_token-secret which means users only have to set `-upload-token=!github.com`. This doesn't require the user to grant Sourcegraph any permissions. Any other thoughts/ideas? 💭

More context: https://sourcegraph.slack.com/archives/CHXHX7XAS/p1571339621085300

cc @felixfbecker

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -3,6 +3,8 @@ package main
 import (
 	"bufio"
 	"compress/gzip"
+	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -10,6 +12,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"strings"
 
 	"github.com/kballard/go-shellquote"
 	"github.com/mattn/go-isatty"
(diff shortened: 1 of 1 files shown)


efritz:
Might be better to: ask for gitthub access token with public_repo permission and determine if they have push or admin access via the `/user/repo` endpoint rather than modifying the repo for the duration of the challenge.

chrismwendt:
Closing to implement @efritz's idea 👍

PR #59, opened 2019-10-29
Check for status code 202 on LSIF upload

mrnugget:
PR https://github.com/sourcegraph/sourcegraph/pull/6227 changed the
status code from `201` to `202`.

That broke the check here and resulted in an `error: 202 Accepted` being
printed.

And the exit code is checked in
https://github.com/sourcegraph/lsif-upload-action, which runs on every
PR in sourcegrah/sourcegraph

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -135,7 +135,7 @@ Examples:
 		// Our request may have failed before the reaching GraphQL endpoint, so
 		// confirm the status code. You can test this easily with e.g. an invalid
 		// endpoint like -endpoint=https://google.com
-		if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
 			body, err := ioutil.ReadAll(resp.Body)
 			if err != nil {
 				return err


cmd/src/lsif_upload.go
```diff
@@ -135,7 +135,7 @@ Examples:
 		// Our request may have failed before the reaching GraphQL endpoint, so
 		// confirm the status code. You can test this easily with e.g. an invalid
 		// endpoint like -endpoint=https://google.com
-		if resp.StatusCode != http.StatusOK {
+		if resp.StatusCode != http.StatusAccepted {
```
felixfbecker: Why not check if the status is 2XX? Seems like that would have prevented this bug from happening and will prevent it if it ever happens again. Also, it would make the CLI work with older versions of Sourcegraph.

mrnugget: Extended the check to check for `StatusOK` and `StatusAccepted`.

felixfbecker: Is there a reason you don't want to check for `StatusCode >= 200 && StatusCode < 300`? This would exhibit the same bug again if code-intel decides to change it to `201 Created`. Imo all 2xx codes should be considered success.

mrnugget: I usually like to be explicit with checks when I know exactly what to check for. Since, here, we only have two possible "success" status codes, we can check explicitly for those.

Add to that that I try to refrain from anticipating from what could also happen in the future, because my personal success rate with such predictions is pretty bad.

mrnugget: That being said: changed it to your proposed solution, since I don't really mind here

felixfbecker: I see. (The `fetch()` API in the browser has `Response.ok` which is `true` if the code is 2xx, and all other API clients I've used (`got`, PowerShell, etc) throw exceptions on non-2xx, so I'm a little biased on this)

PR #62, opened 2019-11-05
LSIF: apply defaults

chrismwendt:
This applies defaults to `-repo`, `-commit`, `-file`, and `-root`. See https://sourcegraph.slack.com/archives/CHXHX7XAS/p1572901719287700

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -10,6 +10,9 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
 
 	"github.com/kballard/go-shellquote"
 	"github.com/mattn/go-isatty"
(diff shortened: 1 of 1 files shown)


cmd/src/lsif_upload.go
```diff
@@ -36,28 +39,84 @@ Examples:
...
+				os.Exit(1)
+			}
+			*repoFlag, err = parseRemoteURL(strings.TrimSpace(string(remoteURL)))
+			if err != nil {
+				fmt.Println(err)
+				fmt.Println("Set -repo explicitly.")
 				os.Exit(1)
 			}
 		}
+		fmt.Println("Repository: " + *repoFlag)
```
efritz: Do we want these to print or was it part of debugging?

chrismwendt: I'm open to feedback here. Printing them makes the defaults transparent.

PR #68, opened 2019-12-02
fix matching of missing-github-token error

sqs:
The error returned by the server is `must provide github_token` (see https://sourcegraph.com/github.com/sourcegraph/sourcegraph@c46dd67b4682266cae719ac380f5a4ca694509db/-/blob/enterprise/internal/codeintel/lsifserver/proxy/auth.go#L45:47).

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -198,8 +198,8 @@ Examples:
 		// confirm the status code. You can test this easily with e.g. an invalid
 		// endpoint like -endpoint=https://google.com
 		if resp.StatusCode < 200 || resp.StatusCode >= 300 {
-			if resp.StatusCode == http.StatusUnauthorized && string(body) == "Must provide github_token.\n" {
-				return fmt.Errorf("error: you have to provide -github-token with 'public_repo' scope")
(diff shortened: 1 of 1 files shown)


felixfbecker:
The root cause of this was asserting errors messages (which are intended for humans, not machines 🤖). @efritz could LSIF server return a consistent `code` or `name` field for this error in addition to `message`?

efritz:
This isn't an error returned by the lsif-server but from the HTTP proxy. This is currently not returning JSON at all in the case of an error (but we could make it do so).

Having consistent error messages is another reason to move ahead with [RFC 74](https://docs.google.com/document/d/1Vw7kF-BrTGMvdAaVVADIMzfv-wzlurMx6oP5DTmbpcc). ;)

felixfbecker:
Oh, interesting. A consistent data encoding definitely makes sense. Otherwise I would recommend to _only_ check for the 401 (I can't think of a case that this would be a false positive for).

> Having consistent error messages is another reason to move ahead with RFC 74. ;)

Actually it would make this harder - we don't currently don't have any way to return an error code from the GraphQL backend, and status codes are always 200 OK. Error messages can be consistent either way, but those are always bad to assert programmatically.

PR #75, opened 2020-01-09
Update installation instructions

efritz:
This change documents the new API endpoint in https://github.com/sourcegraph/sourcegraph/pull/7600 which redirects the user to the recommended src-cli version for the SG instance.


slimsag:
I changed this a bit so that if you copy+paste the commands, it still works by default. If you're happy with that, LGTM, otherwise we can discuss further

PR #76, opened 2020-01-09
Add version command

efritz:
Changes:

- Supply buildTag value to source via `git describe --tags` at build time in CI
- Add version command that prints current buildTag and hits the current SG instance to display the recommended version

```
$ go build -ldflags="-X 'main.buildTag=`git describe --tags`'" ./cmd/src

$ ./src version
Current version: 3.9.0-19-gfa72eb8e1bf6
./srRecommended Version: <unknown>
This Sourcegraph instance does not support this feature.

$ ./src -endpoint http://localhost:3080 version
Current version: 3.9.0-19-gfa72eb8e1bf6
Recommended Version: 3.9.0
```

Note that "dev" is the buildTag if no ldflags are supplied.


cmd/src/main.go
```diff
@@ -35,6 +35,7 @@ The commands are:
 	extensions,ext  manages extensions (experimental)
 	campaigns       manages campaigns (experimental)
 	lsif            manages LSIF data
+	version         compare the src-cli version against the recommended version for your instance
```
mrnugget:
```suggestion
	version         display and compare the src-cli version against the recommended version for your instance
```

PR #78, opened 2020-01-14
add `src repos get`, `src repos list -f`, show more repo fields

sqs:


cmd/src/repos_list.go
```diff
@@ -14,6 +14,10 @@ Examples:
 
     	$ src repos list
 
+  Print JSON description of repositories list:
+
+		$ src repos list -f '{{.|json}}'
```
mrnugget: I don't think the indentation here matches the lines above. I don't know if this is correct:

```suggestion
    	$ src repos list -f '{{.|json}}'
```

PR #79, opened 2020-01-14
add `src campaign plans create-from-patches`

sqs:


cmd/src/campaign_plans_create_from_patches.go
```diff
@@ -0,0 +1,90 @@
...
+
+Stdin is expected to be a JSON array of {repository: string, baseRevision: string, patch: string}. The repository value is the repository's GraphQL ID (which you can look up given a repository name using 'src repos get -name=...').
+
+Examples:
+
+  Create a campaign plan from my.patch applied to a repository's master branch:
+
+		$ patch=$(jq -M -R -s . < my.patch)
+		$ repo=$(src repo get -name=github.com/sourcegraph/src-cli -f '{{.ID|json}}')
+		$ echo '[{"repository": $repo, "baseRevision": "master", "patch": $patch}]' |
```
mrnugget:
```suggestion
		$ echo '[{"repository": $repo, "baseRevision": "master", "patch": $patch}]' |\
```

cmd/src/campaign_plans_create_from_patches.go
```diff
@@ -0,0 +1,90 @@
...
+    	  src campaign plans create-from-patches
```
mrnugget:
```suggestion
    	  src campaigns plan create-from-patches
```

I actually like `plans` more than `plan`, though. But it's `plan` in the other PRs too.

cmd/src/campaign_plans_create_from_patches.go
```diff
@@ -0,0 +1,90 @@
...
+
+`
+
+	flagSet := flag.NewFlagSet("create-from-patches", flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src campaign plans %s':\n", flagSet.Name())
```
mrnugget:
```suggestion
		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src campaigns plan %s':\n", flagSet.Name())
```

cmd/src/campaign_plans_create_from_patches.go
```diff
@@ -0,0 +1,90 @@
...
+
+	"github.com/mattn/go-isatty"
+	"github.com/pkg/errors"
+)
+
+func init() {
+	usage := `
+Create a campaign plan from a set of patches (in unified diff format) to repository branches.
+
+Stdin is expected to be a JSON array of {repository: string, baseRevision: string, patch: string}. The repository value is the repository's GraphQL ID (which you can look up given a repository name using 'src repos get -name=...').
```
mrnugget: Sorry, but I have to:

```suggestion
STDIN is expected to be a JSON array of {repository: string, baseRevision: string, patch: string}. The repository value is the repository's GraphQL ID (which you can look up given a repository name using 'src repos get -name=...').
```

creachadair: Or, if the issue is clarity rather than aesthetics, you could spell it out: "Standard input is expected…"

Since this is documentation for humans that wouldn't be so bad. 😁

mrnugget: In my case it was purely aesthetics 😉

creachadair: Fair enough! (On aesthetics I'd vote for the original capitalization, personally—"stdin" isn't an acronym)

PR #80, opened 2020-01-14
Add `src actions exec` command to generate arbitrary campaign diffs

sqs:
Usage:

Run:

```
go install ./cmd/src && \
docker build -t sqs1/add-lsif-to-build-pipeline-action actions/add-lsif-to-build-pipeline && \
src actions exec -j 4 -keep-logs -f actions/add-lsif-to-build-pipeline/action.json  | \
  src campaign plan create-from-patches
```

image

This will create a campaign that adds GitHub Action LSIF build steps to repositories matched in the action.json file (see and change the query to match repositories that exist on your instance). **Note:** You should only use this with the test sd9 and sd9org repositories, because if you click the link and press **Create**, it will create PRs on real live repositories!

image

Ping me (@sqs) for help on anything here if you pick up this PR.

TODOs:

- [ ] Format of action.json files and overall model for specifying what to run, which repositories to run it on, etc.
- [ ] CLI UX (it's weird you need to pipe stuff)
- [x] Clean up code and add tests
- [x] Currently only works on Linux, needs to support macOS (the `diff` CLI flags we use only work on GNU diff not BSD diff, and the Docker volume mounting breaks due to macOS's `/var/tmp/folders/...` sandboxing)
- [x] ~Document~ (documentation & guide here: https://github.com/sourcegraph/sourcegraph/pull/8064 and also in the `-help` output of the commands)
- [ ] The terminal library (that lets us print progress bars) does not properly handle colored text. If your terminal is too narrow, the progress bars "climb" up the screen. This is due to the terminal library thinking that a line has (for example) 80 chars but it only has (for example) 65 visible chars plus 15 ANSI escape sequence chars for colors. This appears to be a problem in all Go terminal libs. I started to make my own to fix this on a flight w/o WiFi; if it would be helpful, remind me to push it up.
- [x] The proposed `src campaigns create` command is not implemented
- [x] Remove sample actions in the `actions/` dir. These are good candidates for examples in the docs, but they don't belong in this repo.


mrnugget:
I'm picking this up in 3.13

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+		fmt.Println(usage)
+	}
+	const stdin = "<stdin>"
+	userCacheDir, _ := userCacheDir()
+	if userCacheDir != "" {
+		userCacheDir = filepath.Join(userCacheDir, "action-exec")
+	}
+	displayUserCacheDir := strings.Replace(userCacheDir, os.Getenv("HOME"), "$HOME", 1)
+	var (
+		fileFlag        = flagSet.String("f", "<stdin>", `The action file. (required)`)
```
mrnugget: Note to myself: I think it might be more idiomatic to make `-f` optional and read from `stdin` if it's not set. Or use `-f -` to say "read from stdin"

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+		parallelismFlag = flagSet.Int("j", runtime.GOMAXPROCS(0), "The number of parallel jobs.")
+		cacheDirFlag    = flagSet.String("cache", displayUserCacheDir, `Directory for caching results.`)
+		keepLogsFlag    = flagSet.Bool("keep-logs", false, "Do not remove execution log files when done.")
+	)
+
+	handler := func(args []string) error {
+		flagSet.Parse(args)
+
+		if *cacheDirFlag == displayUserCacheDir {
+			*cacheDirFlag = userCacheDir
```
mrnugget: Note to myself: why not use `userCacheDir` as the default value for `cacheDirFlag` then?

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+		}
+
+		var (
+			actionFile []byte
+			err        error
+		)
+		if *fileFlag == stdin {
+			// TODO!(sqs) easy to forget -f flag and have it hang forever
+			actionFile, err = ioutil.ReadAll(os.Stdin)
```
mrnugget: If we read from `stdin`, we [can check whether it's a pipe](https://coderwall.com/p/zyxyeg/golang-having-fun-with-os-stdin-and-shell-pipes). If not, we exit with an error.

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+		for _, run := range action.Run {
+			if run.Type == "docker" && run.Image != "" {
+				// getDockerImageContentDigest gets the content digest for the image. Note that this
+				// is different from the "distribution digest" (which is what you can use to specify
+				// an image to `docker run`, as in `my/image@sha256:xxx`). We need to use the
+				// content digest because the distribution digest is only computed for images that
+				// have been pulled from or pushed to a registry. See
+				// https://windsock.io/explaining-docker-image-ids/ under "A Final Twist" for a good
+				// explanation.
+				getDockerImageContentDigest := func(ctx context.Context, tag string) (string, error) {
```
mrnugget: Note: extract this

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+		var (
+			lwMu sync.Mutex
+			lw   = uilive.New()
+		)
+		lw.Start()
+		defer lw.Stop()
+
+		spinner := []rune{'⠋', '⠙', '⠹', '⠸', '⠼', '⠴', '⠦', '⠧', '⠇', '⠏'}
+		spinnerI := 0
+		onUpdate := func(reposMap map[ActionRepo]ActionRepoStatus) {
```
mrnugget: Extract this and the printing into its own struct (`ProgressPrinter`?) and pass only small function as callback to executor.

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+			panic(err)
+		}
+		if *verbose {
+			log.Printf("# %d repositories match.", len(repos))
+		}
+		for _, repo := range repos {
+			executor.enqueueRepo(repo)
+		}
+
+		onUpdate(executor.repos)
```
mrnugget: This seems "leaky"

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+					DefaultBranch struct{ Name string }
+					Repository    Repository `json:"repository"`
+				}
+			}
+		}
+	}
+	if err := (&apiRequest{
+		query: query,
+		vars: map[string]interface{}{
+			"query": scopeQuery + " count:999999", // TODO!(sqs)
```
mrnugget: Same issue as in https://github.com/sourcegraph/sourcegraph/issues/7861

cmd/src/actions_exec.go
```diff
@@ -0,0 +1,452 @@
...
+		} else {
+			repo = Repository{
+				ID:            searchResult.ID,
+				Name:          searchResult.Name,
+				DefaultBranch: searchResult.DefaultBranch,
+			}
+		}
+
+		if repo.DefaultBranch.Name == "" {
+			continue
```
mrnugget: If verbose logging is turned on, we should log this

cmd/src/actions_exec_backend.go
```diff
@@ -0,0 +1,130 @@
...
+	if x.opt.onUpdate != nil {
+		go func() {
+			for {
+				select {
+				case <-time.After(50 * time.Millisecond):
+					x.reposMu.Lock()
+					x.opt.onUpdate(x.repos)
+					x.reposMu.Unlock()
+
+					// TODO!(sqs): stop when done
```
mrnugget:
```suggestion
			case <-done:
				return
```

cmd/src/actions_exec_backend_runner.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+	if err != nil {
+		status.Err = err
+		fmt.Fprintf(logWriter, "# ERROR: %s\n", err)
+	}
+	x.updateRepoStatus(repo, status)
+
+	// Add to cache if successful.
+	if err == nil {
+		if err := x.opt.cache.set(ctx, cacheKey, status.Patch); err != nil {
```
mrnugget: Note to myself: does this also write to the cache when `status.Patch == CampaignPlanPatch{}`? Or do we need the `len(patch) > 0` check here too?

cmd/src/actions_exec_backend_runner.go
```diff
@@ -0,0 +1,307 @@
...
+
+	x.updateRepoStatus(repo, ActionRepoStatus{
+		LogFile:   logFile.Name(),
+		StartedAt: time.Now(),
+	})
+
+	// TODO!(sqs)
+	//
+	// ctx, cancel := context.WithTimeout(ctx, 90*time.Second)
+	// defer cancel()
```
mrnugget: Note to myself: command-line flag?

cmd/src/actions_exec_backend_runner.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+	}
+
+	// Compute diff.
+	oldDir, err := unzipToTempDir(ctx, zipFile.Name(), prefix)
+	if err != nil {
+		return nil, err
+	}
+	defer os.RemoveAll(oldDir)
+	// TODO!(sqs): these flags apparently only work on GNU diff not BSD (macOS) diff
```
mrnugget: TODO

cmd/src/actions_exec_backend_runner.go
```diff
@@ -0,0 +1,307 @@
...
+					_ = exec.CommandContext(ctx, "docker", "rm", "-f", "--", string(cid)).Run()
+				}
+			}()
+
+			const workDir = "/work"
+			cmd := exec.CommandContext(ctx, "docker", "run",
+				"--rm",
+				"--cidfile", cidFile.Name(),
+				"--workdir", workDir,
+				// TODO!(sqs): on macOS this fails because the volumeDir is /var/folders/... and the
```
mrnugget: TODO. Should we use a different location for the temp folder, or since the `cmd.Dir = volumeDir` further down, is it possible to mount the current working directory?

cmd/src/actions_exec_backend_runner.go
```diff
@@ -0,0 +1,307 @@
...
+func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
+	u, err := url.Parse(cfg.Endpoint)
+	if err != nil {
+		return nil, err
+	}
+	if token != "" {
+		u.User = url.User(token)
+	}
+	u.Path = path.Join(u.Path, repoName+"@"+rev, "-", "raw/")
+	u.Path += "/"
```
mrnugget: Note to myself: double trailing slash?

cmd/src/actions_exec_backend_runner.go
```diff
@@ -0,0 +1,307 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("unable to fetch archive (HTTP %d from %s)", resp.StatusCode, zipURL)
+	}
+
+	f, err := ioutil.TempFile("", strings.Replace(repoName, "/", "-", -1)+".zip")
```
mrnugget: Note: we should probably clean this up in case of an error

mrnugget:
Note: I think the problem is not `BSD diff` vs `GNU diff` but rather `GNU diff 2.8.1` vs `GNU diff 3.7`.

mrnugget:
Update: I think I'm happy with the code so far and would move on to giving a last shot at fixing the moving/scrolling UI, documentation and adding `campaign create`.

mrnugget:
I broke the UI logic, I think, since it now also prints progress when the output is piped to another command.

mrnugget:
@sqs I can't think of any UX improvements that are hard requirements right now. I actually do like the piping. What were your concerns there?

I started to work on `campaigns create` in this PR here https://github.com/sourcegraph/src-cli/pull/87 which is stacked on top of this one.

mrnugget:
Documentation: https://github.com/sourcegraph/sourcegraph/pull/8064

mrnugget:
I'm going to merge this now because the basic feature set and documentation is there. It works, but can be polished:
- "interactive" mode for creating campaigns that opens the editor
- fix the "scrolling" up of the progress bar
- show progress when CLI is downloading repository archives
- ...

But the PR has gotten quite big and further additions would make it unwieldy. Once it's merged I'll open new PRs to polish things.

@sqs Let me know if you have any specific concerns besides the ones I mentioned, or whether your don't think the things I've mentioned don't have priority right now.

PR #84, opened 2020-01-20
Update .travis.yml to use Go 1.13.x instead of 1.10.x

mrnugget:
I don't see a reason why we can't use the newest version, especially since we already use `go.mod` and `go.sum` files.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 sudo: false
 go:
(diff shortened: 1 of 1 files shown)


unknwon:
Builds on Appveyor are already ran with Go 1.13.x: https://ci.appveyor.com/project/sourcegraph/src-cli/builds/30238251#L11, so LGTM.

PR #86, opened 2020-01-23
Add additional docs on installations and releases.

efritz:
Add more explicit installation instructions and more details on how Sourcegraphers should release new versions of the utility.


README.md
```diff
@@ -98,4 +104,17 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
...
+
+| Sourcegraph version | Recommended src-cli version |
+| ------------------- | --------------------------- | 
+| `3.x`               | `3.y.z`                     |
+| `3.{x-1}`           | `3.{y-1}.w`                 |
+
+If a new feature is added to a new `3.y.{z+1}` release of src-cli and this change requires only features available in Sourcegraph `3.{x-1}`, then this feature should also be present in a new `3.{y-1}.{w+1}` release of src-cli. Similar logic should follow for Sourcegraph `3.{x-2}` and its recommended version of src-cli.
+
+Because a Sourcegraph instance will automatically select the highest patch version, almost all non-breaking changes should increment only the patch version. The process above only necessary if a backwards-compatible change is made _after_ a backwards-incompatible one. If the recommended src-cli version for Sourcegraph `3.{x-1}` was instead `3.y.x` in the example above, there is no additional step required, and the new patch version of src-cli will be available to both Sourcegraph versions.
```
chrismwendt: Is the `x` in `3.y.x` intended to match the `x` in `3.x` under the "Sourcegraph version" column?

efritz: Updated.

efritz:
> It's not clear to me why the minor versions differ (x and y).

They don't match _right now_.

> Would it be easier to remember and work with if the src-cli major.minor always matched the major.minor of Sourcegraph?

Possibly but we don't have a unified release cycle for these two projects. I think that's an orthogonal effort.

chrismwendt:
> > It's not clear to me why the minor versions differ (x and y).
> 
> They don't match _right now_.

Oh, they're so close I thought they were supposed to match and https://sourcegraph.com/github.com/sourcegraph/sourcegraph@727dcb2f6ad74f67294c4c07930e172eb2493c02/-/blob/internal/src-cli/consts.go#L9:7 was just out-of-date.

README.md
```diff
@@ -16,7 +16,13 @@ If there is something you'd like to see Sourcegraph be able to do from the CLI,
 
 ## Installation
 
-**NOTE:** To get the best version for _your_ Sourcegraph instance, simply replace `sourcegraph.com` in the commands below with your own Sourcegraph URL and the latest version compatible with your instance will be provided. If you are running a Sourcegraph version older than 3.12, run the commands below instead.
+The preferred method of installation is to ask _your_ Sourcegraph instance for the latest compatible version. To do this, replace `https://sourcegraph.com` in the commands below with the address of your instance. If you are running a Sourcegraph version older than 3.12, these URLs will not be available and you must install directly from sourcegraph.com (by running the following commands _exactly_), or from one of the published [releases on GitHub](https://github.com/sourcegraph/src-cli/releases).
```
creachadair: While there's a bit more information here, I found the previous wording easier to follow. Specifically I think we've folded together the "preferred" and "exception" cases a little bit too much. I suggest we try separating them explicitly, for example:

~~~suggestion
**For Sourcegraph 3.12 and newer:** The preferred method of installation is to fetch the latest compatible version from _your_ Sourcegraph instance. To do this, replace `https://sourcegraph.com` with the address of _your instance_ in the commands below.

**For Sourcegraph prior to 3.12:** Sourcegraph versions older than 3.12 do not support these installation URLs, and you must install directly from sourcegraph.com by running the commands below _exactly_ as shown.  Alternatively, you may also install one of the published [releases on GitHub](https://github.com/sourcegraph/src-cli/releases).
~~~

README.md
```diff
@@ -98,4 +104,17 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
 3.  Travis will automatically perform the release. Once it has finished, **confirm that the curl commands fetch the latest version above**.
 4.  Update the `MinimumVersion` constant in the [src-cli package](https://github.com/sourcegraph/sourcegraph/tree/master/internal/src-cli/consts.go).
 
-**Note**: The version recommended by a Sourcegraph instance will be the highest patch version with the same major and minor version as the set minimum. This means that patch versions are reserved solely for non-breaking changes and minor bug fixes. This allows us to dynamically release fixes for older versions of src-cli without having to update the instance.
+### Patch releases
+
+The version recommended by a Sourcegraph instance will be the highest patch version with the same major and minor version as the set minimum. This means that patch versions are reserved solely for non-breaking changes and minor bug fixes. This allows us to dynamically release fixes for older versions of src-cli without having to update the instance.
```
creachadair: I think we can simplify this a bit by avoiding passive voice:

```suggestion
A Sourcegraph instance returns the highest patch version with the same major and minor version as the set minimum. Patch versions are reserved solely for non-breaking changes and minor bug fixes. This allows us to dynamically release fixes for older versions of `src-cli` without having to update the instance.
```

I also don't see any definition of "set minimum" anywhere in scope. What I think we are saying is "the same major and minor version as the oldest compatible version" (similar to minimum version selection in Go). Is that correct?

README.md
```diff
@@ -98,4 +104,17 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
...
+
+If you are releasing a bug fix or a new feature that is backwards compatible with one of the previous two minor version of Sourcegraph, the changes should be cherry-picked into a patch branch and re-releases with a new patch version. For example, suppose we have the the recommended versions.
```
creachadair: As above, and also a typo fix:

```suggestion
To release a bug fix or a new feature that is backwards compatible with one of the previous two minor version of Sourcegraph, cherry-pick the changes into a patch branch and re-release with a new patch version. 

For example, suppose we have the recommended versions:
```

README.md
```diff
@@ -98,4 +104,17 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
...
+
+| Sourcegraph version | Recommended src-cli version |
+| ------------------- | --------------------------- | 
+| `3.x`               | `3.y.z`                     |
+| `3.{x-1}`           | `3.{y-1}.w`                 |
+
+If a new feature is added to a new `3.y.{z+1}` release of src-cli and this change requires only features available in Sourcegraph `3.{x-1}`, then this feature should also be present in a new `3.{y-1}.{w+1}` release of src-cli. Similar logic should follow for Sourcegraph `3.{x-2}` and its recommended version of src-cli.
```
creachadair: This might be easier to follow with specific numbers so the reader does not have to evaluate the version arithmetic at read-time.

For example if we use "3.104" and "3.105" for Sourcegraph versions and "3.68.4" and "3.67.18" for recommended versions, this becomes:

> If a new feature is added in the 3.68.5 release of `src-cli` and this change is compatible with Sourcegraph 3.104, then this feature **must** also be present in a new 3.67.19 release of `src-cli`. Similar logic applies for a release of Sourcegraph 3.103 and its recommended version of the `src-cli` tool

…or words to that effect. 

Also: In this suggestion I removed one of the two distinct uses of "feature" (the feature being released vs. features of the existing product) to avert confusion about which feature is the subject. 

I also replaced "should" with "must", since we are making here a rule for ourselves to follow during releases.

README.md
```diff
@@ -98,4 +104,17 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
...
+
+Because a Sourcegraph instance will automatically select the highest patch version, almost all non-breaking changes should increment only the patch version. The process above only necessary if a backwards-compatible change is made _after_ a backwards-incompatible one. If the recommended src-cli version for Sourcegraph `3.{x-1}` was instead `3.y.v` in the example above, there is no additional step required, and the new patch version of src-cli will be available to both Sourcegraph versions.
```
creachadair: Can we remove "almost" here? Are there notable exceptions?

Also: I found it confusing to read all the way through this only to find at the end that we don't have to do this legerdemain except in this condition. I suggest front-loading the condition more clearly, e.g., "If you have made a backward-compatible change _after_ releasing a backward-incompatible change, you must …" or words to that effect.

README.md
efritz: Updated.

README.md
efritz: Updated.

README.md
efritz: Front-loaded.

README.md
```diff
@@ -108,4 +114,21 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
...
+
+A Sourcegraph instance returns the highest patch version with the same major and minor version as `MinimumVersion` as defined in the instance. Patch versions are reserved solely for non-breaking changes and minor bug fixes. This allows us to dynamically release fixes for older versions of `src-cli` without having to update the instance.
+
+To release a bug fix or a new feature that is backwards compatible with one of the previous two minor version of Sourcegraph, cherry-pick the changes into a patch branch and re-releases with a new patch version. 
+
+For example, suppose we have the the recommended versions.
+
+| Sourcegraph version | Recommended src-cli version |
+| ------------------- | --------------------------- | 
+| `3.100`             | `3.90.5`                    |
```
creachadair: FYI I thought the x, y stuff was fine in the table, I just figured it would be nice for reading the example if it used real numbers. (This is fine too, though)

README.md
```diff
@@ -108,4 +114,21 @@ go get -u github.com/sourcegraph/src-cli/cmd/src
...
+| `3.99`              | `3.85.7`                    |
+
+If a new feature is added to a new `3.91.6` release of src-cli and this change requires only features available in Sourcegraph `3.99`, then this feature should also be present in a new `3.85.8` release of src-cli. Similar logic should follow for Sourcegraph `3.98` and its recommended version of src-cli. Because a Sourcegraph instance will automatically select the highest patch version, all non-breaking changes should increment only the patch version.
```
creachadair: I thought the logic was only for one rev +/-, is that not the case?

efritz: Oh yes we've updated it. I'll make the change.

PR #87, opened 2020-01-24
WIP: a8n: Add `campaigns create` command

mrnugget:
Missing:

- namespace support. We should get the current user and take its ID as the namespace ID.
- `-interactive` flag


mrnugget:
This is still WIP. I hit the wrong button.

mrnugget:
Merged into #80

PR #90, opened 2020-01-28
Add 'src actions scope-query' command

mrnugget:
This command is meant to help with creating action definitions. It returns the repositories that are matched by the "scopeQuery" in an action definition.

Example:

    $ src actions scope-query -f ~/run-gofmt-in-dockerfile.json
    github.com/sd9/ava
    github.com/sd9/go-diff
    github.com/sourcegraph/automation-testing

It's based on @sqs' suggestion here: https://github.com/sourcegraph/sourcegraph/issues/7850#issuecomment-577678760

### DIFF
--- a/cmd/src/actions.go
+++ b/cmd/src/actions.go
@@ -19,6 +19,7 @@ Usage:
 The commands are:
 
 	exec              executes an action to produce patches
+	scope-query       list the repositories matched by "scopeQuery" in action
 
 Use "src actions [command] -h" for more information about a command.
 `
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -93,7 +93,7 @@ Format of the action JSON files:
 
 	An action JSON needs to specify:
 
-	- "scopeQuery" - a Sourcegraph search query to generate a list of repositories over which to run the action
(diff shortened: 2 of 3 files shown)


felixfbecker:
What's the difference between this and using the search command of the CLI? My intuitive approach would have been to tell the user that they can test scope queries with the search command, instead of having them remember another command or create the impression that the scope query is somehow different from other repo queries (unless it is?).

mrnugget:
> What's the difference between this and using the search command of the CLI?

1. It can read the `scopeQuery` out of an Action definition in a JSON file. Meaning that you can easily switch between `src action scope-query -f action.json` and `src action exec -f action.json`.
2. It produces a unique list of repositories.

felixfbecker:
Ah, I missed (1). Thanks!
(2) is true for the search command too if only a repo filter is specified:

image

mrnugget:
> (2) is true for the search command too if only a repo filter is specified:

Yep, but what customers want is to also get the repositories from file matches. Example: a customer last week specifically searched for a string that only appears in their code files (think: `package ` to only get Go files, for example). The `type:repo` search wouldn't have worked for them.

felixfbecker:
I see. Should use PowerShell haha
```powershell
search-sourcegraph 'repo:sourcegraph/sourcegraph test' | % repositoryName | sort-object -unique
```

PR #91, opened 2020-01-28
Use $EDITOR to define campaign name/desc if not given

mrnugget:
This opens a template markdown file in $EDITOR if either no `-name` or no `-desc` are given on the command line.

This is heavily inspired by the implementation of GitHub's `hub issue create`, except that it's probably 1/4 of the code.

I also took the little regex from it, to determine whether $EDITOR is a Vim and if so set the filetype to Markdown :)

### DIFF
--- a/cmd/src/campaigns_create.go
+++ b/cmd/src/campaigns_create.go
@@ -236,6 +236,10 @@ func extractContent(file string) (string, error) {
 
 func openInEditor(file string) error {
 	editor := os.ExpandEnv("$EDITOR")
+	if editor == "" {
+		return errors.New("$EDITOR is not set")
+	}
+
 	cmd := exec.Command(editor)
 
 	r := regexp.MustCompile(`\b(?:[gm]?vim)(?:\.exe)?$`)
--- a/cmd/src/campaigns_create.go
+++ b/cmd/src/campaigns_create.go
@@ -16,6 +16,8 @@ import (
 
 func init() {
 	usage := `
+Create a campaign with the given attributes. If -name or -desc are not specified $EDITOR will open a temporary Markdown file to edit both.
+
 Examples:
 
   Create a campaign with the given name, description and campaign plan:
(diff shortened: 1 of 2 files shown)


cmd/src/campaigns_create.go
```diff
@@ -34,7 +41,7 @@ Examples:
 	}
 	var (
 		nameFlag        = flagSet.String("name", "", "Name of the campaign. (required)")
-		descriptionFlag = flagSet.String("desc", "", "Description for the campaign. (required)")
+		descriptionFlag = flagSet.String("desc", "", "Description for the campaign in Markdown. (required)")
```
sqs: `-name` and `-description` are no longer required flags, but the flag help says `(required)`

cmd/src/campaigns_create.go
```diff
@@ -126,3 +151,111 @@ const createcampaignMutation = `mutation CreateCampaign($input: CreateCampaignIn
   }
 }
 `
+
+const (
+	sep    = "------- EVERYTHING BELOW THIS LINE WILL BE IGNORE -------"
```
sqs:
```suggestion
	sep    = "------- EVERYTHING BELOW THIS LINE WILL BE IGNORED -------"
```

cmd/src/campaigns_create.go
```diff
@@ -126,3 +151,111 @@ const createcampaignMutation = `mutation CreateCampaign($input: CreateCampaignIn
...
+	}
+	if err = scanner.Err(); err != nil {
+		return "", err
+	}
+
+	return strings.Join(content, "\n"), nil
+}
+
+func openInEditor(file string) error {
+	editor := os.ExpandEnv("$EDITOR")
```
sqs: why not `os.Getenv("EDITOR")`?

mrnugget: Leftover code! Fixed.

cmd/src/campaigns_create.go
```diff
@@ -126,3 +153,111 @@ const createcampaignMutation = `mutation CreateCampaign($input: CreateCampaignIn
...
+The first line of text is the name and the rest is the description.`
+)
+
+type CampaignEditor struct {
+	Name        string
+	Description string
+}
+
+func (e *CampaignEditor) EditAndExtract() (string, string, error) {
+	f, err := ioutil.TempFile("", "new-campaign")
```
keegancsmith: can't you just give this file the suffix `.md` so all editors highlight it correctly?

mrnugget: Saw this too late: https://github.com/sourcegraph/src-cli/pull/92

PR #96, opened 2020-01-29
a8n: add branch flag for creating campaigns

kzh:
This PR adds a flag to specify a branch name when creating campaigns.

### DIFF
--- a/cmd/src/campaigns_create.go
+++ b/cmd/src/campaigns_create.go
@@ -19,10 +19,11 @@ Create a campaign with the given attributes. If -name or -desc are not specified
 
 Examples:
 
-  Create a campaign with the given name, description and campaign plan:
+  Create a campaign with the given name, branch, description and campaign plan:
 
 		$ src campaigns create -name="Format Go code" \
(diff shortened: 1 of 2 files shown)


mrnugget:
Idea as suggested by @sqs in https://github.com/sourcegraph/sourcegraph/issues/7850#issuecomment-577674931:

It would be cool if the branch name could also be specified in the Markdown file by using a comment, such as:

```
Title of the campaign

Description

// branch = my-branch-name
```

There's precedent here: sometimes blog engines or static site generators allow you to add "front matter" to a page, in the form of YAML/TOML/JSON (see Hugo's explanation of it here: https://gohugo.io/content-management/front-matter/)

I don't think we need to fully support that, but some way to declare a line as "meta" would be great.

mrnugget:
Update on my previous comment here: it's more important to get this merged with the basic functionality (`-branch` flag!) than the additonal functionality of specifying the branch name in the markdown file.

mrnugget:
Merging this now since https://github.com/sourcegraph/sourcegraph/pull/8111 has been merged.

I also added a commit that documents this new flag and prints it in the friendly output message.

PR #99, opened 2020-02-04
Make diff flag check work for GNU diffutils 3.3

mrnugget:
It seems that some point after GNU diffutils 2.8.1 the output changed from using a backtick to a single quote.

That's why the check fails even if the diff version supports it.

### DIFF
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -13,6 +13,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"strings"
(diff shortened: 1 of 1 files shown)


felixfbecker:
Could this have been prevented with unit tests in CI?

mrnugget:
> Could this have been prevented with unit tests in CI?

I bet it _could_, but not sure whether it _would_ have been prevent 😄 It would require installing multiple versions of `diff` and I'm honestly not sure whether I would've installed 3.3 (which doesn't support the option but also doesn't use backticks in its output).

felixfbecker:
Yeah I agree 😄

PR #101, opened 2020-02-05
Properly report when command reached timeout

mrnugget:
Before this we just reported that the process was killed, but not that the kill signal was triggered by the timeout.

### DIFF
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -86,7 +86,7 @@ func (x *actionExecutor) do(ctx context.Context, repo ActionRepo) (err error) {
 		}
 	}
 	if err != nil {
-		if errors.Cause(err) == context.DeadlineExceeded {
+		if reachedTimeout(runCtx, err) {
 			err = &errTimeoutReached{timeout: x.opt.timeout}
 		}
 		status.Err = err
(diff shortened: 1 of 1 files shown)


cmd/src/actions_exec_backend_runner.go
```diff
@@ -112,6 +112,16 @@ func (e *errTimeoutReached) Error() string {
 	return fmt.Sprintf("Timeout reached. Execution took longer than %s.", e.timeout)
 }
 
+func reachedTimeout(cmdCtx context.Context, err error) bool {
+	if ee, ok := errors.Cause(err).(*exec.ExitError); ok {
+		if ee.String() == "signal: killed" && cmdCtx.Err() == context.DeadlineExceeded {
```
ryanslade: I think there's a chance of returning a false positive here.

We could have received a kill signal and THEN the timeout happened. We're only checking the value of `cmdCtx.Err()` after we already know `err` isn't nil

mrnugget: Yeah, I know. But is there a reliable way of checking that the timeout is what caused the process to be killed? Even if I turn the check around and check `cmdCtx.Err()` as the first thing, there's still a chance that the process was killed by a user-signal and right after we ran into the timeout, right?

ryanslade: I also couldn't think of a reliable way to check but wanted to point it out in case it was a deal breaker

mrnugget: Yep, appreciated. I think this is better than not reporting anything, though.

PR #121, opened 2020-02-13
Only check for -branch for Sourcegraph >= 3.13, dev or custom build

mrnugget:
This changes the behavior of `campaign create` to only check for the
`-branch` flag if the Sourcegraph instance's version is either

- >= 3.13
- dev
- a custom build date

I'm aware that the "custom build date" check is flaky, but what we want
at the moment is simply to say: only check for this flag if your
Sourcegraph version is newer than `3.12.x`.

### DIFF
--- a/cmd/src/campaigns_create.go
+++ b/cmd/src/campaigns_create.go
@@ -92,7 +92,7 @@ Examples:
 			if err != nil {
 				return err
 			}
-			needsBranch, err := sourcegraphVersionCheck(version, ">= 3.13")
+			needsBranch, err := sourcegraphVersionCheck(version, ">= 3.13", "2020-02-13")
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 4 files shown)


mrnugget:
Also added support for build date.

cmd/src/campaigns_create.go
```diff
@@ -260,3 +274,51 @@ func openInEditor(file string) error {
...
+		result: &sourcegraphVersion,
+	}).do()
+	if err != nil {
+		return "", err
+	}
+
+	return sourcegraphVersion.Site.ProductVersion, nil
+}
+
+func sourcegraphVersionCheck(version, constraint, minDate string) (bool, error) {
```
efritz: lol just shell out to basic-code-intel

mrnugget: You wish, buddy, you wish

PR #122, opened 2020-02-13
Add link for update workflow

eseliger:
image

### DIFF
--- a/cmd/src/format.go
+++ b/cmd/src/format.go
@@ -98,11 +98,14 @@ func parseTemplate(text string) (*template.Template, error) {
 		"friendlyCampaignPlanCreatedMessage": func(campaignPlan CampaignPlan) string {
 			var buf bytes.Buffer
 			fmt.Fprintln(&buf)
-			fmt.Fprintln(&buf, color.HiGreenString("✔  Campaign plan saved."), "To preview and run the campaign (and create branches and changesets):")
+			fmt.Fprintln(&buf, color.HiGreenString("✔  Campaign plan saved."), "\n\nTo preview and run the campaign (and create branches and changesets):")
 			fmt.Fprintln(&buf)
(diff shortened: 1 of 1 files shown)


cmd/src/format.go
```diff
@@ -98,11 +98,14 @@ func parseTemplate(text string) (*template.Template, error) {
...
 			var buf bytes.Buffer
 			fmt.Fprintln(&buf)
-			fmt.Fprintln(&buf, color.HiGreenString("✔  Campaign plan saved."), "To preview and run the campaign (and create branches and changesets):")
+			fmt.Fprintln(&buf, color.HiGreenString("✔  Campaign plan saved."), "\n\nTo preview and run the campaign (and create branches and changesets):")
 			fmt.Fprintln(&buf)
 			fmt.Fprintln(&buf, " ", color.HiCyanString("▶ Web:"), campaignPlan.PreviewURL, color.HiBlackString("or"))
 			cliCommand := fmt.Sprintf("src campaigns create -plan=%s -branch=DESIRED-BRANCH-NAME", campaignPlan.ID)
 			fmt.Fprintln(&buf, " ", color.HiCyanString("▶ CLI:"), cliCommand)
+
+			fmt.Fprintln(&buf, "\nTo update an existing campaign:")
```
mrnugget:
```suggestion
			fmt.Fprintln(&buf, "\nTo update an existing campaign to use this campaign plan:")
```

cmd/src/format.go
```diff
@@ -98,11 +98,14 @@ func parseTemplate(text string) (*template.Template, error) {
...
+			fmt.Fprintln(&buf, "\n ", color.HiCyanString("▶ Web:"), strings.Replace(campaignPlan.PreviewURL, "/new", "/update", 1))
```
mrnugget: Let's make a mental note that we should probably add this to the API once we've settled on a workflow 😄

eseliger: absolutely :D

PR #126, opened 2020-02-14
Only show "update campaign" message if it's supported

mrnugget:
Quoting my comment here: I know that it's hacky to do this in the
formatting helper, but for now I think this is the easiest change to
make without introducing a whole lot of global state and/or mixing the
update property into the CampaignPlan.

If someone knows of a better way, let me know.

### DIFF
--- a/cmd/src/format.go
+++ b/cmd/src/format.go
@@ -95,7 +95,7 @@ func parseTemplate(text string) (*template.Template, error) {
 		"renderResult":                      searchTemplateFuncs["renderResult"],
 
 		// `src campaign plans create-from-patches`
-		"friendlyCampaignPlanCreatedMessage": func(campaignPlan CampaignPlan) string {
+		"friendlyCampaignPlanCreatedMessage": func(campaignPlan CampaignPlan, supportsUpdate bool) string {
 			var buf bytes.Buffer
 			fmt.Fprintln(&buf)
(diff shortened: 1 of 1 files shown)


cmd/src/format.go
```diff
@@ -104,8 +104,24 @@ func parseTemplate(text string) (*template.Template, error) {
...
-			fmt.Fprintln(&buf, "\n ", color.HiCyanString("▶ Web:"), strings.Replace(campaignPlan.PreviewURL, "/new", "/update", 1))
+			// Hacky to do this in a formatting helper, but better than
+			// globally querying the version and only using it here for now.
+			version, err := getSourcegraphVersion()
+			if err != nil {
+				// We ignore the error and return what we have
+				return buf.String()
+			}
+
+			supportsUpdatingPlan, err := sourcegraphVersionCheck(version, ">= 3.13", "2020-02-14")
```
eseliger: the Sourcegraph valentines edition 🌹

efritz:
If you wanted to make it a little cleaner you could just have it queried as part of the startup flow and stash it in a global variable (there won't be an error to check in the formatter then)

mrnugget:
> If you wanted to make it a little cleaner you could just have it queried as part of the startup flow and stash it in a global variable (there won't be an error to check in the formatter then)

Yeah, I considered that. But then you'd have this additional request even if you use `-f {{.}}` as the formatter.

PR #132, opened 2020-02-19
Remove vendor folder and ignore it in gitignore

mrnugget:
I have to admit that I'm not an expert on Go modules, but I do think that we don't need this folder ever since the switch to Go modules?

git history also says that this folder was last changed 14 months ago and when I run `go mod vendor` locally the diff of changes is _huge_


ryanslade:
Good thing we don't get paid by lines added

mrnugget:
I hope to someday be paid by lines removed, though.

PR #133, opened 2020-02-19
Take Docker image content digest into cache key

### LINKED ISSUE #112: a8n: Steps should not be cached if the Docker image has been rebuilt. SHA needs to be checked, not just the tag
My development workflow was to re-build a Docker image with the same tag (latest) and if a repository had a diff previously created, it would skip running the steps over that repository, unless I changed the tag for the Docker image.

mrnugget:
This fixes #112 by fixing a regression I introduced when taking over the
`src actions exec` functionality.

I lower-cased `imageContentDigest` so that users cannot set it. That had
the side effect of `src` itself not writing the `imageContentDigest`
to the cache key (because `json` needs the fields to be exported in
order to marshal/unmarshal them)

The fix here is to upper-case the field and add a separate validation
step to make sure that users don't set the cache field.

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -47,8 +47,8 @@ type ActionStep struct {
 	CacheDirs  []string `json:"cacheDirs,omitempty"`
 	Args       []string `json:"args,omitempty"`
 
-	// imageContentDigest is an internal field that should not be set by users.
-	imageContentDigest string
+	// ImageContentDigest is an internal field that should not be set by users.
+	ImageContentDigest string
 }
 
 type CampaignPlanPatch struct {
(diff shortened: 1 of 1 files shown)


cmd/src/actions_exec.go
```diff
@@ -346,6 +351,30 @@ Format of the action JSON files:
...
+			if step.Dockerfile == "" && step.Image == "" {
+				return fmt.Errorf("docker run step has to specify either 'image' or 'dockerfile'")
+			}
+
+			if step.Dockerfile != "" && step.Image != "" {
+				return fmt.Errorf("docker run step may specify either image (%q) or dockerfile, not both", step.Image)
+			}
+
+			if step.ImageContentDigest != "" {
+				return errors.New("Setting the ImageContentDigest field of a docker run step is not allowed")
```
unknwon: nit:
```suggestion
				return errors.New("setting the ImageContentDigest field of a docker run step is not allowed")
```

PR #138, opened 2020-02-20
Use Go's archive/zip to unzip repository archives

### LINKED ISSUE #109: Unzip failed with interactive message
**unzip version**: `UnZip 6.00 of 20 April 2009, by Info-ZIP.  Maintained by C. Spieler.`
**OS**: `macOS Catalina 10.15.3`
Error log:

```
Unzipping the ZIP archive failed: unzip failed: exit status 1: replace /tmp/action-sd9-akita694954415/angular/playground/projects/ng-entity-service/src/lib/ng-entity-service-notifier.ts? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
(EOF or read error, treating as "[N]one" ...)
```

```
# Repository github.com/sd9/akita @ refs/heads/master (1 steps)
# ERROR: Unzipping the ZIP archive failed: unzip failed: exit status 1: replace /tmp/action-sd9-akita391331449/angular/playground/projects/ng-entity-service/src/lib/ng-entity-service-notifier.ts? [y]es, [n]o, [A]ll, [N]one, [r]ename:  NULL
(EOF or read error, treating as "[N]one" ...)
```

action file: 
```
{
  "scopeQuery": "repo:github.com/sd9/* repohasfile:^package.json",
  "steps": [
    {
      "type": "command",
      "args": ["sed", "-i", "", "s/\"main\"/\"es2015\"/", "package.json"]
    }
  ]
}
```

mrnugget:
This fixes #109 by using Go's archive/zip package to unzip files.

(The code is taken from here and slightly modified: https://golangcode.com/unzip-files-in-go/)

It fixes the "interactive" problem, because it silently ignores
case-sensitive naming conflicts, which is what you get when using macOS'
Finder's unzip utility and unzip on the CLI. "Silently ignores" sounds
bad but I think it's what you expect to happen on macOS.

It also handles the chinese characters correctly.

@eseliger Can you test this with your repositories in #109?

### DIFF
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"archive/zip"
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
@@ -308,12 +309,7 @@ func unzipToTempDir(ctx context.Context, zipFile, prefix string) (string, error)
 	if err != nil {
 		return "", err
 	}
-	unzipCmd := exec.CommandContext(ctx, "unzip", "-qq", zipFile, "-d", volumeDir)
-	if out, err := unzipCmd.CombinedOutput(); err != nil {
-		os.RemoveAll(volumeDir)
-		return "", fmt.Errorf("unzip failed: %s: %s", err, out)
-	}
-	return volumeDir, nil
+	return volumeDir, unzip(zipFile, volumeDir)
 }
 
 func fetchRepositoryArchive(ctx context.Context, repoName, rev string) (*os.File, error) {
@@ -362,4 +358,51 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
 	}
 	u.Path = path.Join(u.Path, repoName+"@"+rev, "-", "raw")
 	return u, nil
+}
+
+func unzip(zipFile, dest string) error {
+	r, err := zip.OpenReader(zipFile)
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+
+	for _, f := range r.File {
+		fpath := filepath.Join(dest, f.Name)
+
+		// Check for ZipSlip. More Info: http://bit.ly/2MsjAWE
+		if !strings.HasPrefix(fpath, filepath.Clean(dest)+string(os.PathSeparator)) {
+			return fmt.Errorf("%s: illegal file path", fpath)
+		}
+
+		if f.FileInfo().IsDir() {
+			os.MkdirAll(fpath, os.ModePerm)
+			continue
+		}
+
+		if err = os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
+			return err
+		}
+
+		outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
+		if err != nil {
+			return err
+		}
+
+		rc, err := f.Open()
+		if err != nil {
+			return err
+		}
+
+		_, err = io.Copy(outFile, rc)
+
+		outFile.Close()
+		rc.Close()
+
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
 }
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -376,11 +376,14 @@ func unzip(zipFile, dest string) error {
 		}
 
 		if f.FileInfo().IsDir() {
-			os.MkdirAll(fpath, os.ModePerm)
+			err := os.MkdirAll(fpath, os.ModePerm)
+			if err != nil {
+				return err
+			}
 			continue
 		}
 
-		if err = os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
+		if err := os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
 			return err
 		}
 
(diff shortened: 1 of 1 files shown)


cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,50 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+		if err != nil {
+			return err
+		}
+
+		rc, err := f.Open()
+		if err != nil {
+			return err
+		}
+
+		_, err = io.Copy(outFile, rc)
```
creachadair: We shouldn't ignore the error on closing `outFile`, since it's open for writing.

```go
_, err := io.Copy(outFile, rc)
rc.Close()
_, cerr := outFile.Close()
if err != nil {
  return err
} else if cerr != nil {
  return cerr
}
```

Otherwise we may silently drop data in uncommon-but-therefore-very-hard-to-debug cases.

mrnugget: Ahh, nice. Thank you!

cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,50 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+		}
+
+		outFile, err := os.OpenFile(fpath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, f.Mode())
+		if err != nil {
+			return err
+		}
+
+		rc, err := f.Open()
+		if err != nil {
+			return err
```
creachadair:
```suggestion
			outFile.Close()
			return err
```

To avoid orphaning a filehandle.

cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,50 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+		if !strings.HasPrefix(fpath, filepath.Clean(dest)+string(os.PathSeparator)) {
+			return fmt.Errorf("%s: illegal file path", fpath)
+		}
+
+		if f.FileInfo().IsDir() {
+			os.MkdirAll(fpath, os.ModePerm)
+			continue
+		}
+
+		if err = os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
```
creachadair:
```suggestion
		if err := os.MkdirAll(filepath.Dir(fpath), os.ModePerm); err != nil {
```

we don't really need to shadow here.

cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,50 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+	for _, f := range r.File {
+		fpath := filepath.Join(dest, f.Name)
+
+		// Check for ZipSlip. More Info: http://bit.ly/2MsjAWE
+		if !strings.HasPrefix(fpath, filepath.Clean(dest)+string(os.PathSeparator)) {
+			return fmt.Errorf("%s: illegal file path", fpath)
+		}
+
+		if f.FileInfo().IsDir() {
+			os.MkdirAll(fpath, os.ModePerm)
```
creachadair: We don't care if this fails?

mrnugget:
@creachadair I've tried to address all of your feedback. Can you take another look? (Also: your thoroughness earned you a spot on my ask-for-review list 😎)

eseliger:
fixes all reported nits from #109 :)

cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,62 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+	for _, f := range r.File {
+		fpath := filepath.Join(dest, f.Name)
+
+		// Check for ZipSlip. More Info: http://bit.ly/2MsjAWE
+		if !strings.HasPrefix(fpath, filepath.Clean(dest)+string(os.PathSeparator)) {
+			return fmt.Errorf("%s: illegal file path", fpath)
+		}
+
+		if f.FileInfo().IsDir() {
+			err := os.MkdirAll(fpath, os.ModePerm)
```
creachadair: This is of course fine as you wrote it, but I'd just one-line it as elsewhere:

```suggestion
			if err := os.MkdirAll(fpath, os.ModePerm); err != nil {
```

cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,62 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+	if err != nil {
+		return err
+	}
+	defer r.Close()
+
+	for _, f := range r.File {
+		fpath := filepath.Join(dest, f.Name)
+
+		// Check for ZipSlip. More Info: http://bit.ly/2MsjAWE
+		if !strings.HasPrefix(fpath, filepath.Clean(dest)+string(os.PathSeparator)) {
```
creachadair: I'd advise hoisting the prefix assembly out of the loop:

```go
outputBase := filepath.Clean(dest)+string(os.PathSeparator)
// …
if !strings.HasPrefix(fpath, outputBase) {
```

Not so much for performance but to be self-documenting.

cmd/src/actions_exec_backend_runner.go
```diff
@@ -363,3 +359,62 @@ func repositoryZipArchiveURL(repoName, rev, token string) (*url.URL, error) {
...
+			return err
+		}
+
+		rc, err := f.Open()
+		if err != nil {
+			outFile.Close()
+			return err
+		}
+
+		if _, err = io.Copy(outFile, rc); err != nil {
```
creachadair: I would recommend doing this differently. As written, we're still missing a close of `outFile` if `rc.Close` fails. And we actually don't _care_ if closing `rc` fails, because it's only open for reading so there is no risk to data. At best we might want to log that.

What I would write here is:

```go
_, err := io.Copy(outFile, rc)
rc.Close()
_, cerr := outFile.Close()
// Now we have safely closed everything that needs it, and can check errors
if err != nil {
   return fmt.Errorf("copying %q: %v", f.Name, err)
}
if cerr != nil {
  return fmt.Errorf("closing: %v", cerr)
}
```

I know this pattern seems a little weird at first, but it's a lot easier to be sure we haven't missed a case if we do error checking as a separate phase. We _always_ want to close, regardless of what fails, and only after we've done that do we care whether something broke.

mrnugget: Ah, _now_ I get it. Thank you!

mrnugget:
> What have I done? 😆

Don't worry, I appreciate it. "This is my jam", as they say :)

Issue #137, opened 2020-02-19
Diff produced by src actions exec doesn't respect gitignore


ryan-blunden:
It seems that auto-generated files that are ignored by `.gitignore` such as [schema files](https://github.com/sourcegraph/sourcegraph/blob/master/.gitignore#L104) still make there way into a campaign plan.

To reproduce:

```
git clone https://github.com/sourcegraph/campaign-examples.git
cd campaign-examples/rxjs-upgrade
make exec # You may want to change the scopeQuery to just be sourcegraph
make campaign-plan
```

Follow the link to view the campaign plan in the UI and observe schema files in the diff [( see this example campaign on sgdemo)](https://demo.sgdev.org/campaigns/Q2FtcGFpZ246Ng==).

mrnugget:
To clarify: the bug is "diff doesn't respect .gitignore", right?

eseliger:
I think so. At least that is _one_ bug or at least a source for confusion

mrnugget:
While `diff` (from `GNU diffutils` at least) accepts an `--exclude-from=<file-with-ignore-patterns>` flag, there's two problems preventing us from simply using `--exclude-from=fresh-folder/.gitignore`:

1. The patterns in `.gitignore` are often relative to the folder in which they're specified (Example: `./foobar` in `/.gitignore` only ignores `/foobar` but not `/vendor/foobar`). When we're running `diff`, we're on level up and the patterns don't match anymore. AFAICT there's no way to tell `diff` to respect `.gitignore` files natively
2. `git` allows `.gitignore` in every subfolder, which doesn't seem transferrable to `--exclude-from`

PR #140, opened 2020-02-24
Replace terminal UI with a more readable logger when src actions exec is run in verbose mode

### LINKED ISSUE #119: a8n: Output shows duplicate lines when verbose flag is given
See this recording:

verbose

mrnugget:
When `src action exec` previously ran in verbose-mode with the `-v` flag, it looked like this:

old_verbose

With the changes in this PR it looks like this:

new_logger

Please note: I didn't focus on the details (such as colors, exact words in message, etc). The goal was to untangle the logging from the `actionExecutor` and the terminal UI, while keeping the behavior in place:

* In verbose mode, repo-specific log messages are sent to standard error and to a repo-specific log file
* When not in verbose mode, repo-specific log messages are only sent to repo-specific log file
* Log files are cleaned up if not `-keep-logs` flag is given or the repo didn't produce an error

This fixes #119


cmd/src/actions_exec.go
```diff
@@ -182,15 +177,23 @@ Format of the action JSON files:
...
 		ctx := context.Background()
 
+		logger := newActionLogger(*verbose, *keepLogsFlag)
+
 		diffSupportsNoDereference, err = diffSupportsFlag(ctx, "--no-dereference")
 		if err != nil {
 			return err
 		}
+		if !diffSupportsNoDereference {
+			logger.Warnf("Local installation of 'diff' doesn't support '%s' flag. Consider upgrading to GNU diff >=3.7.\n", "--no-dereference")
```
eseliger: 3.3 I think. For some reason the `suggestion` button is disabled so I cannot make one :facepalm:

mrnugget: You think this is necessary? Newest version is `3.7`, I think we should recommend upgrading to that :)

eseliger: yup definitely makes sense to have people use always the most recent version, I was thinking it might confuse people when they are on some OS that doesn't have 3.7 in their package lists that they cannot use src-cli, but it might be unreasonable to account for that 😁

mrnugget: Let's stay with this for now, since I also want to explore some other options (to fix #137) and this might be invalidated soon.

eseliger: sure :)

PR #142, opened 2020-02-25
Add -create-plan flag to 'src action exec'

mrnugget:
When the -create-plan flag is given the `src action exec` command will
create a campaign plan from the produced patches, making it not
necessary to pipe the output to `src campaign plan create-from-patches`.

If the execution of the action failed in a subset of the repositories,
the user is prompted and asked whether they would create a campaign plan
from the successfully produced patches or not.

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
@@ -17,6 +18,7 @@ import (
 	"time"
 
 	"github.com/fatih/color"
+	"github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
 	"github.com/sourcegraph/go-diff/diff"
 )
@@ -61,19 +63,23 @@ Examples:
 
   Execute an action defined in ~/run-gofmt-in-dockerfile.json:
 
-    	$ src actions exec -f ~/run-gofmt-in-dockerfile.json
-
-  Verbosely execute an action and keep the logs available for debugging:
-
-		$ src -v actions exec -keep-logs -f ~/run-gofmt-in-dockerfile.json
+	$ src actions exec -f ~/run-gofmt-in-dockerfile.json
 
   Execute an action and create a campaign plan from the patches it produced:
 
-    	$ src actions exec -f ~/run-gofmt-in-dockerfile.json | src campaign plan create-from-patches
+	$ src actions exec -f ~/run-gofmt-in-dockerfile.json -create-plan
+
+  Verbosely execute an action and keep the logs available for debugging:
+
+	$ src -v actions exec -keep-logs -f ~/run-gofmt-in-dockerfile.json
+
+  Execute an action and pipe the patches it produced to 'src campaign plan create-from-patches':
+
+	$ src actions exec -f ~/run-gofmt-in-dockerfile.json | src campaign plan create-from-patches
 
   Read and execute an action definition from standard input:
 
-		$ cat ~/my-action.json | src actions exec -f -
+	$ cat ~/my-action.json | src actions exec -f -
 
 
 Format of the action JSON files:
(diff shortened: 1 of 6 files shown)


cmd/src/actions_exec.go
```diff
@@ -141,6 +147,10 @@ Format of the action JSON files:
 		cacheDirFlag    = flagSet.String("cache", displayUserCacheDir, "Directory for caching results.")
 		keepLogsFlag    = flagSet.Bool("keep-logs", false, "Do not remove execution log files when done.")
 		timeoutFlag     = flagSet.Duration("timeout", defaultTimeout, "The maximum duration a single action run can take (excluding the building of Docker images).")
+
+		createPlanFlag = flagSet.Bool("create-plan", false, "Create a campaign plan from the produced set of patches. When the execution of the action fails in a single repository a prompt will ask to confirm or reject the campaign plan creation.")
```
eseliger: to make this easier to use in a non tty environment, can we have a second flag like `--force-campaign-plan-creation`?

cmd/src/actions_exec.go
```diff
@@ -489,3 +531,28 @@ func diffStatDiagram(stat diff.Stat) string {
...
+		response, err := reader.ReadString('\n')
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		response = strings.ToLower(strings.TrimSpace(response))
+
+		if response == "y" || response == "yes" {
+			return true
+		} else if response == "n" || response == "no" {
```
eseliger: maybe everything else just defaults to no? That's what `terraform` does

mrnugget: Sure. I just copy/pasted this function from somewhere, haven't looked into it that much yet (branch is still a draft)

eseliger: oops, I could swear I read it was not 🤦‍♂

christinaforney:
image

Wording feedback:
- "Cached result without patch found" - does this mean that there will be a diff for it? Or just that it didn't need to re-create the diff because there is a cached result? If this is accurate, perhaps rewording: "Found cached result, using cached diff"
- How do I clear my cached results? (this can be solved in docs)
- Future feature: ignore caches flag
- The location of the log file appears to only be printed on Enqueue - this means it won't be shown for every repo. Perhaps show on completion (failure or success) instead?
- The failures that say "run docker container: exit status 1" is that the step that failed? Can we preface that with "Failed to apply changes: <error message>. Logfile: <path to logfile>"
- Style 'Action produced 2 patches' similar to 'Campaign plan saved', make sure there is a new line between the last repo status line and this result line.
- Style 'WARNING' similar to campaign saved but make it yellow (or red) and with an X instead of a check. 
- WARNING only shows one error, either show all of them in a list, or none of them and say to look above. Instead "WARNING: N Action(s) failed with errors."
- "To preview and run the campaign (and create branches and changesets):" -> "Preview and create this campaign on Sourcegraph using one of the following options:"

mrnugget:
>     * "Cached result without patch found" - does this mean that there will be a diff for it? Or just that it didn't need to re-create the diff because there is a cached result? If this is accurate, perhaps rewording: "Found cached result, using cached diff"

"does this mean that there will be a diff for it?" — No. "without patch" means the exact opposite: there is no diff. We ran the action in this repository before, which is why the result is cached, but it didn't produce a diff.

That's because an action can successfully (!) run in a repository without producing diff/patch. Example: `"scopeQuery": "repohasfile:README.md"` and the action replaces every `example\.com` in the `README.md` with `blog\.example\.com`. If there is no `example\.com` in the `README.md` then nothing is replaced and no diff is produced.

> How do I clear my cached results? (this can be solved in docs)

Does ["Clearing the campaign action cache"](https://docs.sourcegraph.com/user/campaigns#clearing-the-campaign-action-cache) in the docs explain that sufficiently? Short version: the cache is just a folder on disk. If you delete its contents, the cache is cleared. Default cache location is shown when running `src action exec -help` and a custom location/folder can be set with `src action exec -cache my-cache`

> Future feature: ignore caches flag

Yeah, I think a `-no-cache` flag might help.

> The failures that say "run docker container: exit status 1" is that the step that failed? Can we preface that with "Failed to apply changes: . Logfile: "

Exactly. In the line above you can see `Step 0: error: exit status 1` and the `run docker container` is another line. It's confusing that there are two lines. I'll try to see if I can fix that. 

---

Here's what I have so far:

#### 2 patches, 2 failures, `-create-plan`

Screen Shot 2020-02-28 at 15 33 40

#### 2 patches, 2 failures

Screen Shot 2020-02-28 at 15 34 30

#### 2 patches, 0 failures

Screen Shot 2020-02-28 at 15 37 28

#### 2 patches, 0 failures, `-create-plan`

Screen Shot 2020-02-28 at 15 38 01

mrnugget:
@christinaforney I'm going to merge this branch now and will address the double-logging of errors and future feedback you have on wording etc. in separate PRs.

christinaforney:
> No. "without patch" means the exact opposite

Ahhhh gotcha. I missed the subtlety of with patch and without patch.

Then instead, how about this wording for those two states:
- Found cached result: using cached diff
- Found cached result: no diff produced  for this repository

The confusing part for me was that there was a cache with no changes and that it was a "good"/successful result to apply. Otherwise, it sounds like there was a cached result, but the diff wasn't found and I'm worried I missed a change.

Improvements look great!

PR #143, opened 2020-02-27
Log cache hits when running 'src action exec'

mrnugget:
I forgot to add this when introducing the new logger. This now logs when a result for a repository was found in the cache and depending on whether the result has a patch it's either bold or not bold.

Screen Shot 2020-02-27 at 14 54 57


mrnugget:
> I almost could't tell the difference if I haven't read your description. Maybe we should try some emoji magic? :trollface:

I know it's super subtle and it's more of a joke on my part, but... I was _this_ close to using emojis... or at least some ASCII characters but I couldn't come up with a consistent scheme. Let me know if you have any ideas!

PR #145, opened 2020-02-27
Ensure we infer the LSIF upload root when no value is supplied.

efritz:


cmd/src/lsif_upload.go
```diff
@@ -13,13 +13,31 @@ import (
...
 	"os/exec"
+
 	"path/filepath"
 	"strings"
 
 	"github.com/kballard/go-shellquote"
 	"github.com/mattn/go-isatty"
 )
 
+// stringFlag is used to wrap the value of the -root argument so that we can
```
creachadair: This works, but if we only need to tell whether a flag was explicitly specified or not, this may be easier:

```go
func isFlagSet(fs *flag.FlagSet, name string) bool {
   var found bool
   fs.Visit(func(f *flag.Flag) {
      if f.Name == name {
         found = true
      }
   })
   return found
}
```

efritz: Updated.

PR #146, opened 2020-02-28
Change wording of "cached result found" log message

mrnugget:
@christinaforney I hit "merge" too early in the other PR, so here's a follow-up with your suggestion.

### DIFF
--- a/cmd/src/actions_exec_logger.go
+++ b/cmd/src/actions_exec_logger.go
@@ -102,7 +102,7 @@ func (a *actionLogger) Infof(format string, args ...interface{}) {
 func (a *actionLogger) RepoCacheHit(repo ActionRepo, patchProduced bool) {
 	if a.verbose {
 		if patchProduced {
-			fmt.Fprintf(os.Stderr, "%s -> Found cached result: using cached patch.\n", boldGreen.Sprint(repo.Name))
+			fmt.Fprintf(os.Stderr, "%s -> Cached result found: using cached diff.\n", boldGreen.Sprint(repo.Name))
 			return
 		}
 
--- a/cmd/src/actions_exec_logger.go
+++ b/cmd/src/actions_exec_logger.go
@@ -106,7 +106,7 @@ func (a *actionLogger) RepoCacheHit(repo ActionRepo, patchProduced bool) {
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


cmd/src/actions_exec_logger.go
```diff
@@ -102,11 +102,11 @@ func (a *actionLogger) Infof(format string, args ...interface{}) {
 func (a *actionLogger) RepoCacheHit(repo ActionRepo, patchProduced bool) {
 	if a.verbose {
 		if patchProduced {
-			fmt.Fprintf(os.Stderr, "%s -> Cached result with patch found.\n", boldGreen.Sprint(repo.Name))
+			fmt.Fprintf(os.Stderr, "%s -> Found cached result: using cached patch.\n", boldGreen.Sprint(repo.Name))
```
christinaforney:
```suggestion
			fmt.Fprintf(os.Stderr, "%s -> Cached result found: using cached diff.\n", boldGreen.Sprint(repo.Name))
```

cmd/src/actions_exec_logger.go
```diff
@@ -102,11 +102,11 @@ func (a *actionLogger) Infof(format string, args ...interface{}) {
...
 			return
 		}
 
-		fmt.Fprintf(os.Stderr, "%s -> Cached result without patch found.\n", green.Sprint(repo.Name))
+		fmt.Fprintf(os.Stderr, "%s -> Found cached result: no patch produced for this repository.\n", green.Sprint(repo.Name))
```
christinaforney:
```suggestion
		fmt.Fprintf(os.Stderr, "%s -> Cached result found: no diff produced for this repository.\n", green.Sprint(repo.Name))
```

mrnugget:
> Thoughts on diff vs. patch? I like diff because it reminds me of `git diff`, but I know we've also been using patches in a lot of our other language.

I honestly prefer `diff` but I always assumed that `patch` sounds more correct?

christinaforney:
I just found this: https://devmanual.gentoo.org/tools-reference/diff-and-patch/index.html and https://stackoverflow.com/questions/4215722/difference-between-patch-and-diff-files

I think we can choose what we'd like. One person on stack overflow said "patch is a collection of diffs", but that doesn't seem to be widely adopted. The gentoo devmanual is probably a better standard to goo off of:
> "The diff tool is used to create patches (sometimes called diffs)"

So we can choose whatever we'd like and right now we're 2/2 for `diff`.

PR #147, opened 2020-03-01
Fix new git based diff generation file paths

eseliger:
Before, `/a` -> `/b` was the prefix for all files, so new files have the `/b` prefix and existing ones are moved.
This flag omits them.

### DIFF
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -239,7 +239,10 @@ func runAction(ctx context.Context, prefix, repoID, repoName, rev string, steps 
 		return nil, errors.Wrap(err, "git add failed")
 	}
 
-	// --no-prefix omits the a/ and b/ folder prefixes, otherwise the diff would be interpreted as renaming the file
+	// As of Sourcegraph 3.14 we only support unified diff format.
+	// That means we need to strip away the `a/` and `/b` prefixes with `--no-prefix`.
+	// See: https://github.com/sourcegraph/sourcegraph/blob/82d5e7e1562fef6be5c0b17f18631040fd330835/enterprise/internal/campaigns/service.go#L324-L329
+	// 
 	diffOut, err := runGitCmd("diff", "--cached", "--no-prefix")
 	if err != nil {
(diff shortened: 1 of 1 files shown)


mrnugget:
I don't understand where this caused a problem, can you explain?

The standard `git diff` has these `a/` and `b/` prefixes:
```shell
$ echo "\n\n\n foobar" >> README.md
$ git diff
diff --git a/README.md b/README.md
index a94dcc0..00bccbe 100644
--- a/README.md
+++ b/README.md
@@ -134,3 +134,7 @@ For example, suppose we have the the recommended versions.
 If a new feature is added to a new `3.91.6` release of src-cli and this change requires only features available in Sourcegraph `3.99`, then this feature should also be present in a new `3.85.8` release of src-cli. Because a Sourcegraph instance will automatically select the highest patch version, all non-breaking changes should increment only the patch version.

 Note that if instead the recommended src-cli version for Sourcegraph `3.99` was `3.90.4` in the example above, there is no additional step required, and the new patch version of src-cli will be available to both Sourcegraph versions.
+
+
+
+ foobar

File: cmd/src/actions_exec_backend_runner.go [hunk 1]
```diff
@@ -239,7 +239,8 @@ func runAction(ctx context.Context, prefix, repoID, repoName, rev string, steps
 		return nil, errors.Wrap(err, "git add failed")
 	}
 
-	diffOut, err := runGitCmd("diff", "--cached")
+	// --no-prefix omits the a/ and b/ folder prefixes, otherwise the diff would be interpreted as renaming the file
```
mrnugget:
```suggestion
	// As of Sourcegraph 3.14 we only support unified diff format.
	// That means we need to strip away the `a/` and `/b` prefixes with `--no-prefix`.
	// See: https://github.com/sourcegraph/sourcegraph/blob/82d5e7e1562fef6be5c0b17f18631040fd330835/enterprise/internal/campaigns/service.go#L324-L329
	// 
```

eseliger:
yup, true. we can probably detect the `--git` flag and remove them

PR #149, opened 2020-03-02
Add json schema for actions definitions

### LINKED ISSUE #151: Remove inline dockerfile option
I like the intention behind making it super simple to run a container without creating a Dockerfile, but what's the likelihood of this being used in a content change scenario?

The [example used in the docs](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@61c81dd96206b3ad59ff175cbbc054ff4e984631/-/blob/doc/user/campaigns.md#L89) is valid for sure, but couldn't this be achieved by specifying `alpine:3` as the image, then supplying the `CMD` in the example as `args` instead?

Removing the option to me would simplify things, and allow code to be removed, all without losing functionality (just needs to be done in a different way).

eseliger:
As discussed, there were issues every now and then because the schema is not fully validated. This should cover off all cases we currently support and can probably simplify the `validateAction` method. 

Closes #151

### DIFF
--- a/actions.schema.json
+++ b/actions.schema.json
@@ -1,0 +1,78 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "$id": "actions.schema.json#",
+  "title": "Action Definition",
+  "description": "Describes an action with its scope and steps to perform.",
+  "allowComments": true,
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "scopeQuery",
+    "steps"
+  ],
+  "properties": {
+    "scopeQuery": {
+      "description": "A Sourcegraph search query to generate a list of repositories over which to run the action. Use 'src actions scope-query' to see which repositories are matched by the query.",
+      "type": "string",
+      "minLength": 1
+    },
+    "steps": {
+      "description": "A list of action steps to execute in each repository.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["type"],
+        "additionalProperties": false,
+        "properties": {
+          "type": {
+            "description": "Can either be of type \"command\", which means the step is executed on the machine on which 'src actions exec' is executed, or it can be of type \"docker\" which then (optionally builds) and runs a container in which the repository is mounted.",
+            "type": "string",
+            "enum": ["command", "docker"]
+          },
+          "args": {
+            "description": "The command to execute.",
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "type": "string"
+            }
+          },
+          "image": {
+            "description": "Docker image handle to pull for executing this step. If no `args` are provided, the default CMD will be executed.",
+            "type": "string",
+            "minLength": 1
+          },
+          "dockerfile": {
+            "description": "An inline defined dockerfile to build the image from",
+            "type": "string",
+            "minLength": 1
+          }
+        },
+        "oneOf": [
+          {
+            "required": ["args"],
+            "properties": {
+              "type": { "const": "command" },
+              "image": { "type": "null" },
+              "dockerfile": { "type": "null" }
+            }
+          },
+          {
+            "required": ["image"],
+            "properties": {
+              "type": { "const": "docker" },
+              "dockerfile": { "type": "null" }
+            }
+          },
+          {
+            "required": ["dockerfile"],
+            "properties": {
+              "type": { "const": "docker" },
+              "image": { "type": "null" }
+            }
+          }
+        ]
+      }
+    }
+  }
+}
--- a/actions.schema.json
+++ b/actions.schema.json
@@ -38,12 +38,7 @@
             }
           },
           "image": {
-            "description": "Docker image handle to pull for executing this step. If no `args` are provided, the default CMD will be executed.",
-            "type": "string",
-            "minLength": 1
-          },
-          "dockerfile": {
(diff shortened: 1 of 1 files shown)


actions.schema.json
```diff
@@ -0,0 +1,78 @@
...
+    "steps": {
+      "description": "A list of action steps to execute in each repository.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["type"],
+        "additionalProperties": false,
+        "properties": {
+          "type": {
+            "description": "Can either be of type \"command\", which means the step is executed on the machine on which 'src actions exec' is executed, or it can be of type \"docker\" which then (optionally builds) and runs a container in which the repository is mounted.",
```
ryan-blunden:
```suggestion
            "description": "Can be either \"command\", which executes the step in the native environment (OS) of the machine where 'src actions exec' is executed, or \"docker\" which runs a container with the repository contents mounted in at `/work`.",
```

actions.schema.json
```diff
@@ -0,0 +1,78 @@
...
+          "args": {
+            "description": "The command to execute.",
+            "type": "array",
+            "minItems": 1,
+            "items": {
+              "type": "string"
+            }
+          },
+          "image": {
+            "description": "Docker image handle to pull for executing this step. If no `args` are provided, the default CMD will be executed.",
```
ryan-blunden:
```suggestion
            "description": "The Docker image handle for running the container executing this step. Just like when running `docker container run`,  `args` here override the default `CMD` to be executed. ",
```

actions.schema.json
```diff
@@ -0,0 +1,78 @@
...
+            "type": "string",
+            "minLength": 1
+          },
+          "dockerfile": {
+            "description": "An inline defined dockerfile to build the image from",
```
ryan-blunden:
```suggestion
            "description": "An inline defined Dockerfile to build the image from",
```

ryan-blunden: I created https://github.com/sourcegraph/src-cli/issues/151 as I think this should be removed. I would suggest removing it from the schema at this point unless we have customers that are using it.

eseliger:
I think therefore it would make sense to have the schema hosted somewhere. What would be a good place to deploy that to? 

I just tried it with a github gist and it seems vscode accepts arbitrary URLs

```json
{
    "$schema": "https://gist.githubusercontent.com/eseliger/2d30d1b730e5b71503fd98acf91dfbdb/raw/72e0fc9d4049cc48c3b0ea5c5210fc5cd281c642/actions.schema.json",
    "scopeQuery": "repo:go-* -repohasfile:INSTALL.md",
    "steps": [
        {
        "type": "command",
        "args": ["sh", "-c", "echo '# Installation' > INSTALL.md"]
        },
        {
        "type": "docker",
        "dockerfile": "FROM alpine:3 \n CMD find /work -iname '*.md' -type f | xargs -n 1 sed -i s/this/that/g"
        },
        {
        "type": "docker",
        "image": "golang:1.13-alpine",
        "args": ["go", "fix", "/work/..."]
        }
    ]
}
```

image

ryan-blunden:
Any reason why we wouldn't just link to the raw file contents of the schema in the master branch?

eseliger:
Yea that might be alright. I was just thinking of the case where we update the schema on master but an older version is the one compatible with what is currently recommended for users of sourcegraph@v.XXX so they might encounter new fields that aren't yet supported. But it is probably good-enough for a long time 👍

actions.schema.json
```diff
@@ -0,0 +1,64 @@
...
+    "steps": {
+      "description": "A list of action steps to execute in each repository.",
+      "type": "array",
+      "items": {
+        "type": "object",
+        "required": ["type"],
+        "additionalProperties": false,
+        "properties": {
+          "type": {
+            "description": "Can be either \"command\", which executes the step in the native environment (OS) of the machine where 'src actions exec' is executed, or \"docker\" which runs a container with the repository contents mounted in at `/work`. Note that local images (not from a Docker registry) must be built manually prior to executing.",
```
mrnugget:
```suggestion
            "description": "Can be either \"command\", which executes the step in the native environment (OS) of the machine where 'src actions exec' is executed, or \"docker\" which runs a container with the repository contents mounted in at `/work`. Note that images from the Docker registry must be pulled manually prior to executing.",
```

I think it's clear that local _images_ must be built before (where would you get the image name from, if not from `docker build`?) See my other comment for the `dockerfile` attribute.

actions.schema.json
```diff
@@ -0,0 +1,64 @@
...
+            "items": {
+              "type": "string"
+            }
+          },
+          "image": {
+            "description": "The Docker image handle for running the container executing this step. Just like when running `docker run`, `args` here override the default `CMD` to be executed.",
+            "type": "string",
+            "minLength": 1
+          }
+        },
```
mrnugget: The `"dockerfile"` property is missing.

This is a valid definition:

```json
{
  "scopeQuery": "repo:github",
  "steps": [
    {
      "type": "docker",
      "dockerfile": "FROM alpine:3 \n CMD find /work -iname '*.txt' -type f | xargs -n 1 sed -i s/this/that/g"
    }
  ]
}
```

eseliger: Had that but followed up on
https://github.com/sourcegraph/src-cli/pull/149#discussion_r386700542

PR #158, opened 2020-03-11
Add BaseRef field to CampaignPlanPatch and use commitID in BaseRevision

mrnugget:
This implements the src CLI tasks of https://github.com/sourcegraph/sourcegraph/issues/8846

With this change src CLI uses the commit ID (instead of refspec) as the
`CampaignPlanPatch.BaseRevision` when creating a `CampaignPlan`.

It also adds the additional `baseRef` field to `CampaignPlanPatch` and
sends that to the GraphQL.

Since Sourcegraph <3.14 expects the `baseRevision` of a
`CampaignPlanPatch` to contain a ref (e.g. "refs/heads/master") we check
the Sourcegraph version and fall back to that behaviour so that
customers can use a newer src CLI version with their <3.14
instance.

With the addition of these fields to ActionRepo and CampaignPlanPatch
the cache of src CLI is now also busted when the revision of the default
branch of a repo changes.

### DIFF
--- a/cmd/src/campaign_plans_create_from_patches.go
+++ b/cmd/src/campaign_plans_create_from_patches.go
@@ -97,6 +97,33 @@ func createCampaignPlanFromPatches(
 	var result struct {
 		CreateCampaignPlanFromPatches CampaignPlan
 	}
+
+	version, err := getSourcegraphVersion()
+	if err != nil {
+		return err
+	}
+	supportsBaseRef, err := sourcegraphVersionCheck(version, ">= 3.14-0", "2020-03-11")
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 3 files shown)


cmd/src/campaign_plans_create_from_patches.go
```diff
@@ -97,6 +97,33 @@ func createCampaignPlanFromPatches(
 	var result struct {
 		CreateCampaignPlanFromPatches CampaignPlan
 	}
+
+	version, err := getSourcegraphVersion()
+	if err != nil {
+		return err
+	}
+	supportsBaseRef, err := sourcegraphVersionCheck(version, ">= 3.14-0", "2020-03-11")
```
eseliger:
```suggestion
	supportsBaseRef, err := sourcegraphVersionCheck(version, ">= 3.14.0", "2020-03-11")
```

PR #162, opened 2020-03-27
Update LSIF upload parameter docs

efritz:
- Fix ambiguous documentation for the repo (should it be a cloneable URI?, etc).
- Fix incorrect default for root.

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -75,7 +75,7 @@ Examples:
 		if repoFlag == nil || *repoFlag == "" {
 			remoteURL, err := exec.Command("git", "remote", "get-url", "origin").Output()
 			if err != nil {
-				fmt.Printf("Failed to invoke git: %q\n", err)
+				fmt.Printf("Failed to invoke git: %v\n", err)
 				fmt.Println("Unable to detect repository from environment.")
 				fmt.Println("Either cd into a git repository or set -repo explicitly.")
 				os.Exit(1)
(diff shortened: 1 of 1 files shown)


cmd/src/lsif_upload.go
```diff
@@ -59,11 +59,11 @@ Examples:
 		fmt.Println(usage)
 	}
 	var (
-		repoFlag        = flagSet.String("repo", "", `The name of the repository. By default, derived from the origin remote.`)
+		repoFlag        = flagSet.String("repo", "", `The name of the repository (e.g. github.com/gorilla/mux). By default, derived from the origin remote.`)
```
creachadair: Does a URL work? If so maybe: `The name or fetch URL…`? (I _think_ anything that can be passed to `git fetch` works, but I'm not sure)

cmd/src/lsif_upload.go
```diff
@@ -75,7 +75,7 @@ Examples:
 		if repoFlag == nil || *repoFlag == "" {
 			remoteURL, err := exec.Command("git", "remote", "get-url", "origin").Output()
 			if err != nil {
-				fmt.Println(err)
+				fmt.Printf("Failed to invoke git: %q\n", err)
```
creachadair: For values of type error, prefer `%v` or `%s` (they are equivalent).

cmd/src/lsif_upload.go
efritz: We explicitly don't pass git://foo/bar.git (we parse it out below).

Issue #165, opened 2020-03-30
Add `src action clear-cache` subcommand


mrnugget:
The feedback in https://gist.github.com/unknwon/b7c80d23e6b2a624e50e4073454959e0 is now another mention of how difficult it is to clear the cache for `src action exec`.

I discussed with @unknwon that it would be best to add a `src action clear-cache` command that deletes the cache (as opposed to adding a `-no-cache` flag to `src action exec`, because it's complicated to explain that it ignores cache on _read_ but probably not on _write_)

mrnugget:
On second thought: a `-clear-cache` flag to `src action exec` might be better, since it would work with the specified `-cache` directory.

PR #166, opened 2020-03-30
Add -clear-cache flag to src action exec

mrnugget:
This fixes https://github.com/sourcegraph/src-cli/issues/165 and works like this:

Screen Shot 2020-03-30 at 16 25 38

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -164,9 +164,6 @@ Format of the action JSON files:
 			return errors.New("cache is not a valid path")
 		}
 
-		if *clearCacheFlag {
-		}
-
 		var (
 			actionFile []byte
 			err        error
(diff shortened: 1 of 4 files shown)


cmd/src/actions_exec.go
```diff
@@ -161,6 +164,9 @@ Format of the action JSON files:
 			return errors.New("cache is not a valid path")
 		}
 
+		if *clearCacheFlag {
```
eseliger: Hm?

mrnugget: Compiler optimization: if I check for the flag, the compiler will more likely put the boolean into a register so the check will be faster down the line.

mrnugget: (Just kidding of course, I forgot to remove this 😛 )

eseliger: The world of Go is endless 🙃

Issue #111, opened 2020-02-13
Confused by 405 error caused by src-config.json endpoint having a trailing slash


ryan-blunden:
If the` endpoint` in the src-config.json has a trailing slash, a `405 method not allowed` error is returned. I expected this to be handled gracefully, or complain about incorrect configuration.

PR #167, opened 2020-03-30
Remove trailing slash from SRC_ENDPOINT

mrnugget:
We already had this in place, just not for `SRC_ENDPOINT`. This should fix https://github.com/sourcegraph/src-cli/issues/111 even though I cannot reproduce it with my local setup. I assume it only happens when nginx is in front of sourcegraph.


mrnugget:
Can confirm that this fixes #111. Just tested it with k8s.sgdev.org

PR #169, opened 2020-03-30
Cancel context on os.Interrupt

mrnugget:
This fixes https://github.com/sourcegraph/src-cli/issues/117 by
cancelling the global context when Ctrl-C is hit.

The cancelation propagates down to `runAction` where containers that are
still running are killed.

Screen Shot 2020-03-30 at 18 00 22

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -10,6 +10,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"regexp"
 	"runtime"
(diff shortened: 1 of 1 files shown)


cmd/src/actions_exec.go
```diff
@@ -183,7 +184,20 @@ Format of the action JSON files:
 			return errors.Wrap(err, "invalid JSON action file")
 		}
 
-		ctx := context.Background()
+		ctx, cancel := context.WithCancel(context.Background())
+		c := make(chan os.Signal, 1)
+		signal.Notify(c, os.Interrupt)
+		defer func() {
+			signal.Stop(c)
+			cancel()
```
eseliger: is there any effect of calling cancel twice?

mrnugget: Why should it be called twice?

eseliger: I would have guessed once when the signal handler hits in the go function, and once when the outer function is over running the defer function? 🤔 not sure if that's the case though ..

mrnugget: Ah! I see.

https://golang.org/pkg/context/#CancelFunc

> After the first call, subsequent calls to a CancelFunc do nothing.

PR #172, opened 2020-03-31
Add 'src action create' cmd to create empty action definition

### LINKED ISSUE #153: Add `src actions create` command to create actions.json boilerplate
To help surface the actions schema and show different step examples, create a `create-schema` command which would create an `action.json` file in the current directory. I'd advise skipping the dumping it to stdout so the user can successfully do it the first time around without re-executing it with redirection to a file.

In the boilerplate file, a [`$schema`](https://github.com/sourcegraph/src-cli/pull/149#issuecomment-593668192) field would be added to surface the schema. 

This would make tutorial videos and instructions for creating an initial `action.json` file super easy and preconfiguring support for code completion and code doc hovers in VS Code would be really cool.

mrnugget:
This fixes #153 but depends on #171.

### DIFF
--- a/cmd/src/actions_create.go
+++ b/cmd/src/actions_create.go
@@ -1,0 +1,60 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"os"
+)
+
+const actionDefinitionTemplate = `{
+  "$schema": "https://github.com/sourcegraph/src-cli/tree/master/schema/actions.schema.json",
+  "scopeQuery": "",
+  "steps": [
+  ]
+}
+`
+
+func init() {
+	usage := `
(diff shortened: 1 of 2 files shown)


cmd/src/actions_create.go
```diff
@@ -0,0 +1,60 @@
...
+  "$schema": "https://github.com/sourcegraph/src-cli/tree/master/schema/actions.schema.json",
+  "scopeQuery": "",
+  "steps": [
+  ]
+}
+`
+
+func init() {
+	usage := `
+Create an empty action definition in action.json (if not -o flag is given). This command is meant to help with creating action definitions to be used with 'src actions exec'.
```
ryanslade:
```suggestion
Create an empty action definition in action.json (if no -o flag is given). This command is meant to help with creating action definitions to be used with 'src actions exec'.
```

PR #173, opened 2020-04-03
Skip repositories being cloned in action exec instead of returning error

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/9149

When a repository is being cloned while we execute the `scopeQuery` of
an action, the GraphQL query returns something like this:

    {
      "errors": [
        {
          "message": "repository does not exist: github.com/sourcegraphtest/alwayscloningtest",
          "path": [ "search", "results", "results", 2, "defaultBranch" ]
        }
      ],
      "data": {
        "search": {
          "results": {
            "results": [
              {
                "__typename": "Repository",
                "id": "UmVwb3NpdG9yeTox",
                "name": "github.com/sourcegraph/automation-testing",
                "defaultBranch": { "name": "refs/heads/master", "target": { "oid": "3a0d12026c1349c4aefb712145433cc26c330a06" }
                }
              },
              {
                "__typename": "Repository",
                "id": "UmVwb3NpdG9yeTozOTU=",
                "name": "github.com/sourcegraphtest/AlwaysCloningTest",
                "defaultBranch": null
              }
            ]
          }
        }
      }
    }

There is an error for `AlwaysCloningTest` but `AlwaysCloningTest` is
also being returned in the data, except that its `defaultBranch` is
`null`.

So instead of returning an error when one of the repositories is being
cloned and doesn't have a default branch, we simply check for the
existence of the `defaultBranch` attribute and if it's not there, we
skip the repository.

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -465,25 +465,32 @@ query ActionRepos($query: String!) {
 `
 	type Repository struct {
 		ID, Name      string
-		DefaultBranch struct {
+		DefaultBranch *struct {
 			Name   string
 			Target struct{ OID string }
 		}
 	}
 	var result struct {
-		Search struct {
-			Results struct {
-				Results []struct {
-					Typename      string `json:"__typename"`
-					ID, Name      string
-					DefaultBranch struct {
-						Name   string
-						Target struct{ OID string }
+		Data struct {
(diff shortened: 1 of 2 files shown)


eseliger:
nice 🎊 Reminder to myself: port this to `actions-runner`

PR #176, opened 2020-04-15
Fix -clear-cache flag when cache is cleared and improve error handling

mrnugget:


cmd/src/actions_exec_backend_runner.go
```diff
@@ -233,7 +233,7 @@ func runAction(ctx context.Context, prefix, repoID, repoName, rev string, steps
 			elapsed := time.Since(t0).Round(time.Millisecond)
 			if err != nil {
 				logger.DockerStepErrored(repoName, i, err, elapsed)
-				return nil, errors.Wrap(err, "run docker container")
+				return nil, errors.Wrapf(err, "Running Docker container for image '%s' failed", step.Image)
```
ryanslade, line 236: Maybe use %q?

PR #178, opened 2020-04-21
Add cacheDirs to action definition JSON schema

mrnugget:
Looks like we missed this.

### DIFF
--- a/schema/actions.schema.json
+++ b/schema/actions.schema.json
@@ -49,9 +49,8 @@
             "minLength": 1
           },
           "cacheDirs": {
-            "description": "Names of directories to create for in a temporary location and mount into each \"docker\" step container under the specified name.",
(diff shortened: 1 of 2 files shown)


schema/actions.schema.json
```diff
@@ -47,6 +47,14 @@
             "description": "The Docker image handle for running the container executing this step. Just like when running `docker run`, `args` here override the default `CMD` to be executed.",
             "type": "string",
             "minLength": 1
+          },
+          "cacheDirs": {
+            "description": "Names of directories to create for in a temporary location and mount into each \"docker\" step container under the specified name.",
```
ryanslade, line 52:
```suggestion
            "description": "Names of directories to create in a temporary location and mount into each \"docker\" step container under the specified name.",
```

PR #191, opened 2020-05-05
Pull missing docker images automatically

### LINKED ISSUE #183: Automatically pull missing docker images
I think it would be expected behavior to pull any missing docker images off the registry, instead of failing with an error that they aren't found.

eseliger:
Closes #183

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -216,7 +216,7 @@ Format of the action JSON files:
 			return errors.Wrap(err, "Validation of action failed")
 		}
 
-		// Build Docker images etc.
+		// Build Docker images, pull Docker images from registries, etc.
 		err = prepareAction(ctx, action, logger)
 		if err != nil {
 			return errors.Wrap(err, "Failed to prepare action")
@@ -408,29 +408,17 @@ func getDockerImageContentDigest(ctx context.Context, image string, logger *acti
 	// https://github.com/moby/moby/issues/32016.
 	out, err := exec.CommandContext(ctx, "docker", "image", "inspect", "--format", "{{.Id}}", "--", image).CombinedOutput()
 	if err != nil {
-		// An error could mean the image just didn't exist, so we need to pull it first.
-		logger.Infof("Pulling image %q...\n", image)
+		if strings.Index(err.Error(), "No such image") != -1 {
+			return "", fmt.Errorf("error inspecting docker image %q: %s", image, bytes.TrimSpace(out))
+		}
+		logger.Infof("Pulling Docker image %q...\n", image)
 		pullCmd := exec.CommandContext(ctx, "docker", "image", "pull", image)
-		stdout, err := pullCmd.StdoutPipe()
-		if err != nil {
-			return "", fmt.Errorf("error pulling docker image %q: %s", image, err)
+		if logger.verbose {
+			stdout := logger.InfoPipe(fmt.Sprintf("%q", image))
+			pullCmd.Stdout = stdout
 		}
-		stderr, err := pullCmd.StderrPipe()
-		if err != nil {
-			return "", fmt.Errorf("error pulling docker image %q: %s", image, err)	
-		}
-		outScanner := bufio.NewScanner(stdout)
-		go func() {
-			for outScanner.Scan() {
-				logger.Infof("%s\n", outScanner.Text())
-			}
-		}()
-		errScanner := bufio.NewScanner(stderr)
-		go func() {
-			for errScanner.Scan() {
-				logger.Warnf("%s\n", errScanner.Text())
-			}
-		}()
+		stderr := logger.ErrorPipe(fmt.Sprintf("%q", image))
+		pullCmd.Stderr = stderr
 		err = pullCmd.Start()
 		if err != nil {
 			return "", fmt.Errorf("error pulling docker image %q: %s", image, err)
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -241,7 +241,7 @@ Format of the action JSON files:
 
 		logger := newActionLogger(*verbose, *keepLogsFlag)
 
-		// Build Docker images etc.
+		// Fetch Docker images etc.
 		err = prepareAction(ctx, action, logger)
 		if err != nil {
 			return errors.Wrap(err, "Failed to prepare action")
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -413,10 +413,8 @@ func getDockerImageContentDigest(ctx context.Context, image string, logger *acti
 		}
 		logger.Infof("Pulling Docker image %q...\n", image)
 		pullCmd := exec.CommandContext(ctx, "docker", "image", "pull", image)
-		if logger.verbose {
-			stdout := logger.InfoPipe(fmt.Sprintf("docker image pull %s", image))
-			pullCmd.Stdout = stdout
-		}
+		stdout := logger.InfoPipe(fmt.Sprintf("docker image pull %s", image))
+		pullCmd.Stdout = stdout
 		stderr := logger.ErrorPipe(fmt.Sprintf("docker image pull %s", image))
 		pullCmd.Stderr = stderr
 		err = pullCmd.Start()
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -424,7 +424,7 @@ func getDockerImageContentDigest(ctx context.Context, image string, logger *acti
 	// https://github.com/moby/moby/issues/32016.
 	out, err := exec.CommandContext(ctx, "docker", "image", "inspect", "--format", "{{.Id}}", "--", image).CombinedOutput()
 	if err != nil {
-		if !strings.Contains(err.Error(), "No such image") {
+		if !strings.Contains(string(out), "No such image") {
 			return "", fmt.Errorf("error inspecting docker image %q: %s", image, bytes.TrimSpace(out))
 		}
 		logger.Infof("Pulling Docker image %q...\n", image)
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -217,7 +217,7 @@ Format of the action JSON files:
 		}
 
 		// Build Docker images etc.
-		err = prepareAction(ctx, action)
+		err = prepareAction(ctx, action, logger)
 		if err != nil {
 			return errors.Wrap(err, "Failed to prepare action")
 		}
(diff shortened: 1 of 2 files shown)


cmd/src/actions_exec.go
```diff
@@ -401,14 +401,49 @@ func prepareAction(ctx context.Context, action Action) error {
...
+			for outScanner.Scan() {
+				logger.Infof("%s\n", outScanner.Text())
+			}
+		}()
+		errScanner := bufio.NewScanner(stderr)
+		go func() {
+			for errScanner.Scan() {
+				logger.Warnf("%s\n", errScanner.Text())
+			}
+		}()
```
mrnugget, line 433: How are these goroutines terminated?

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,49 @@ func prepareAction(ctx context.Context, action Action) error {
...
+func getDockerImageContentDigest(ctx context.Context, image string, logger *actionLogger) (string, error) {
 	// TODO!(sqs): is image id the right thing to use here? it is NOT the
 	// digest. but the digest is not calculated for all images (unless they are
 	// pulled/pushed from/to a registry), see
 	// https://github.com/moby/moby/issues/32016.
 	out, err := exec.CommandContext(ctx, "docker", "image", "inspect", "--format", "{{.Id}}", "--", image).CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("error inspecting docker image (try `docker pull %q` to fix this): %s", image, bytes.TrimSpace(out))
+		// An error could mean the image just didn't exist, so we need to pull it first.
+		logger.Infof("Pulling image %q...\n", image)
```
mrnugget, line 412:
```suggestion
		logger.Infof("Pulling Docker image %q...\n", image)
```

cmd/src/actions_exec.go
```diff
@@ -217,7 +217,7 @@ Format of the action JSON files:
 		}
 
 		// Build Docker images etc.
```
mrnugget, line 219: Can you update that comment?

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,49 @@ func prepareAction(ctx context.Context, action Action) error {
...
-func getDockerImageContentDigest(ctx context.Context, image string) (string, error) {
+func getDockerImageContentDigest(ctx context.Context, image string, logger *actionLogger) (string, error) {
 	// TODO!(sqs): is image id the right thing to use here? it is NOT the
 	// digest. but the digest is not calculated for all images (unless they are
 	// pulled/pushed from/to a registry), see
 	// https://github.com/moby/moby/issues/32016.
 	out, err := exec.CommandContext(ctx, "docker", "image", "inspect", "--format", "{{.Id}}", "--", image).CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("error inspecting docker image (try `docker pull %q` to fix this): %s", image, bytes.TrimSpace(out))
+		// An error could mean the image just didn't exist, so we need to pull it first.
```
mrnugget, line 411: How do we determine that? What if the error is "docker: command does not exist"? We then log "Pulling image..." which would fail.

Is there are a way to determine whether we should pull the image or return the error?

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,49 @@ func prepareAction(ctx context.Context, action Action) error {
...
+			return "", fmt.Errorf("error pulling docker image %q: %s", image, err)	
+		}
+		outScanner := bufio.NewScanner(stdout)
+		go func() {
+			for outScanner.Scan() {
+				logger.Infof("%s\n", outScanner.Text())
+			}
+		}()
+		errScanner := bufio.NewScanner(stderr)
+		go func() {
```
mrnugget, line 429: Instead of doing this whole dance with `StdoutPipe`/`StderrPipe` and the scanners, I think we can simplify this by moving some of that into the logger:

```go
stdout := logger.InfoPipe()
pullCmd.Stdout = stdout
```

The `InfoPipe` and `WarnPipe` methods could return a prefixed `io.Writer` that **only logs if the verbose mode is turned on**.

Take a look at this method to get some ideas:

https://github.com/sourcegraph/src-cli/blob/1266471f6ceb8c23d527bc068799ba7913971aee/cmd/src/actions_exec_logger.go#L144-L160

cmd/src/actions_exec.go
eseliger: As far as I understood it it should terminate after EOF. But not a 100% sure, the docs aren't super clear about it https://golang.org/pkg/text/scanner/#Scanner.Scan

cmd/src/actions_exec.go
eseliger: The logic 

- tries to get the digest
- on fail, tries to pull
- tries to get the digest again
- on fail, it's not related to the pull and we return the error

mrnugget: Yeah, I know that, but the "on fail" could be caused by anything, right?

If "tries to get the digest" fails because docker is not installed, then we run "docker pull" and print more stuff to stderr.

In other words: can't we check the output returned by "docker image inspect" to see if it contains "No such image" (or something) and only then try to pull?

eseliger:
I reworked the log streaming, PTAL :)

cmd/src/actions_exec.go
```diff
@@ -216,8 +216,8 @@ Format of the action JSON files:
 			return errors.Wrap(err, "Validation of action failed")
 		}
 
-		// Build Docker images etc.
-		err = prepareAction(ctx, action)
+		// Build Docker images, pull Docker images from registries, etc.
```
mrnugget, line 219: We don't build images anymore :)

```suggestion
		// Pull Docker images from registries, get image digests, etc.
```

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,37 @@ func prepareAction(ctx context.Context, action Action) error {
...
-func getDockerImageContentDigest(ctx context.Context, image string) (string, error) {
+func getDockerImageContentDigest(ctx context.Context, image string, logger *actionLogger) (string, error) {
 	// TODO!(sqs): is image id the right thing to use here? it is NOT the
 	// digest. but the digest is not calculated for all images (unless they are
 	// pulled/pushed from/to a registry), see
 	// https://github.com/moby/moby/issues/32016.
 	out, err := exec.CommandContext(ctx, "docker", "image", "inspect", "--format", "{{.Id}}", "--", image).CombinedOutput()
 	if err != nil {
-		return "", fmt.Errorf("error inspecting docker image (try `docker pull %q` to fix this): %s", image, bytes.TrimSpace(out))
+		if strings.Index(err.Error(), "No such image") != -1 {
```
mrnugget, line 411:
```suggestion
		if !strings.Contains(err.Error(), "No such image") {
```

cmd/src/actions_exec_logger.go
```diff
@@ -141,6 +141,21 @@ func (a *actionLogger) RepoWriter(repoName string) (io.Writer, bool) {
 	return w, ok
 }
 
+func (a *actionLogger) InfoPipe(prefix string) io.Writer {
+	if !a.verbose {
+		return ioutil.Discard
+	}
+	stdoutPrefix := fmt.Sprintf("%s -> [STDOUT]: ", yellow.Sprint(prefix))
+	stdout := textio.NewPrefixWriter(os.Stderr, stdoutPrefix)
+	return io.Writer(stdout)
```
mrnugget, line 150: It's called `stdout` but writes to `os.Stderr`. Why? Shouldn't this be independent of "STDOUT" or "STDERR"? If not, then I'd rename both methods to `StdoutPipe` and `StderrPipe`.

cmd/src/actions_exec_logger.go
```diff
@@ -141,6 +141,21 @@ func (a *actionLogger) RepoWriter(repoName string) (io.Writer, bool) {
...
+}
+
+func (a *actionLogger) ErrorPipe(prefix string) io.Writer {
```
mrnugget, line 153: Why doesn't this have a `if !a.verbose` check?

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,37 @@ func prepareAction(ctx context.Context, action Action) error {
...
+			return "", fmt.Errorf("error inspecting docker image %q: %s", image, bytes.TrimSpace(out))
+		}
+		logger.Infof("Pulling Docker image %q...\n", image)
+		pullCmd := exec.CommandContext(ctx, "docker", "image", "pull", image)
+		if logger.verbose {
```
mrnugget, line 416: That's redundant, because the check is also in `InfoPipe`.

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,37 @@ func prepareAction(ctx context.Context, action Action) error {
...
+			stdout := logger.InfoPipe(fmt.Sprintf("%q", image))
```
mrnugget, line 417:
```suggestion
			stdout := logger.InfoPipe(fmt.Sprintf("docker image pull %s", image))
```

cmd/src/actions_exec.go
```diff
@@ -401,14 +401,37 @@ func prepareAction(ctx context.Context, action Action) error {
...
+			pullCmd.Stdout = stdout
+		}
+		stderr := logger.ErrorPipe(fmt.Sprintf("%q", image))
```
mrnugget, line 420:
```suggestion
		stderr := logger.ErrorPipe(fmt.Sprintf("docker image pull %s", image))
```

eseliger:
I finished this one up and seems to work well for me, can I get another review after I merged master into this branch? :)

cmd/src/actions_exec.go
```diff
@@ -417,14 +417,35 @@ func prepareAction(ctx context.Context, action Action) error {
...
-		return "", fmt.Errorf("error inspecting docker image (try `docker pull %q` to fix this): %s", image, bytes.TrimSpace(out))
+		if !strings.Contains(string(out), "No such image") {
+			return "", fmt.Errorf("error inspecting docker image %q: %s", image, bytes.TrimSpace(out))
+		}
+		logger.Infof("Pulling Docker image %q...\n", image)
+		pullCmd := exec.CommandContext(ctx, "docker", "image", "pull", image)
+		stdout := logger.InfoPipe(fmt.Sprintf("docker image pull %s", image))
+		pullCmd.Stdout = stdout
+		stderr := logger.ErrorPipe(fmt.Sprintf("docker image pull %s", image))
+		pullCmd.Stderr = stderr
```
mrnugget, line 435:
```suggestion
		prefix := fmt.Sprintf("docker image pull %s", image)
		pullCmd.Stdout = logger.InfoPipe(prefix)
		pullCmd.Stderr = logger.ErrorPipe(prefix)

```

PR #192, opened 2020-05-05
Fix action create schema url

eseliger:
The other URL was wrong because it pointed to the HTML rendered page on GitHub, which doesn't work.


mrnugget:
Thank you! I remember noticing this but apparently I messed up when trying to "fix" it 🤦

PR #193, opened 2020-05-05
Only yield repositories from supported codehosts

### LINKED ISSUE #185: Number of patches in TUI is different from number in UI
After running an action, the UI said `Action produced 41 patches.`. When I went in the UI, I only got 30 patches displayed. Maybe the other 11 were empty after parsing in the frontend? 🤔

### LINKED ISSUE #186: Filter out repos from unsupported codehosts when evaluating scope query
When running an action, I produced patches for gitolite and gitlab as well, but they didn't end up shown in the UI. We should not even spend the compute and filter them out early

eseliger:
Closes #186 

I also retried the action that I reported https://github.com/sourcegraph/src-cli/issues/185 on, it is also fixed by this change: Closes #185

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -441,30 +441,37 @@ query ActionRepos($query: String!) {
 			results {
 				__typename
 				... on Repository {
-					id
-					name
-					defaultBranch {
-						name
-						target { oid }
-					}
+					...repositoryFields
 				}
 				... on FileMatch {
 					repository {
-						id
-						name
-						defaultBranch {
-							name
-							target { oid }
-						}
+						...repositoryFields
 					}
 				}
 			}
 		}
 	}
 }
+
+fragment repositoryFields on Repository {
+	id
+	name
+	externalRepository {
+		serviceType
+	}
+	defaultBranch {
+		name
+		target {
+			oid
+		}
+	}
+}
 `
 	type Repository struct {
-		ID, Name      string
+		ID, Name           string
+		ExternalRepository struct {
+			ServiceType string
+		}
 		DefaultBranch *struct {
 			Name   string
 			Target struct{ OID string }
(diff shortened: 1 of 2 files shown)


eseliger:
I see what you mean. I was worried that it would be unintuitive, as it took me almost 2 months to realize my actions had run over way more repos than needed, making them feel unnecessarily slow :/ 
If we go with the "we allow to generate and track all patches until a codehost integration is done", I think it should be under a flag or something `--all-codehosts` for example, making the default `false`, so people aren't missing it and are feeling the execution of an action is slower than it needs to be. WDYT?

mrnugget:
But then we'd still be stuck with this problem:

> If we were to add this restriction to src-cli we'd effectively lock ourselves out of this feature until we release a new src-cli and users upgrade.

Right? If we filter out non-GitHub/BBS by default and then, in 2 months, allow creation of drafts for GitLab, or even have full GitLab support, we would need to upgrade the clients.

If you think that's acceptable, then at the very least I think we should

1. add the skipped repositories to the `skipped` list (or another list)
2. print those repositories in `-v` mode (e.g. "Skipping 18 repositories on unsupported code hosts")
3. the check for the servicetype should lower-case everything, so that we're safe.
 
---

Another option:

1. add a field to our GraphQL APIs' `repository`, something like `supportedByCampaigns`?
2. query that field (if the sourcegraph version is new enough)
3. if it's false, add the repo to the skipped list etc.

eseliger:
I think with the flag, people can even use this version of src cli to generate patches for unsupported code hosts, so we wouln't give up much. Some way to disable building for code hosts that we don't even support would be helpful in the future for that anyways, so I don't think it's a waste here. Plus, in the future we could set the default to `true`

mrnugget:
> Plus, in the future we could set the default to `true`

Sure, but not for old clients 😃

---

But let's go with this!

Issue #189, opened 2020-04-30
Notify user if scope query didn't match anything


ryanslade:
This should happen even in "non verbose" mode.

Currently the user has no way to know that the query hasn't matched.

PR #194, opened 2020-05-05
Warn user when scopeQuery doesn't match anything

ryanslade:
Closes: https://github.com/sourcegraph/src-cli/issues/189

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -559,6 +559,11 @@ query ActionRepos($query: String!) {
 	for _, repo := range reposByID {
 		repos = append(repos, repo)
 	}
+
+	if len(repos) == 0 && !*verbose {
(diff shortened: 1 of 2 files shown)


cmd/src/actions_exec.go
```diff
@@ -559,6 +559,11 @@ query ActionRepos($query: String!) {
 	for _, repo := range reposByID {
 		repos = append(repos, repo)
 	}
+
+	if len(repos) == 0 && !*verbose {
+		yellow.Fprintf(os.Stderr, "WARNING: No repositories matched by scopeQuery\n")
```
eseliger, line 564: should that affect the exit code as well?

ryanslade: I don't think so, but happy to be persuaded otherwise. It's not really a hard error

eseliger: No blocker, we can revisit that decision when we feel it would be beneficial

PR #196, opened 2020-05-05
Allow yaml action files and make it the default format

eseliger:
This is the general implementation that makes YAML as an input file work. But there's food for discussion here:


felixfbecker:
>  action.yml is associated with GitHub actions by default in vscode - should we change the name of these files alltogether?

Maybe `src-action.yml`?

eseliger:
Everything with action in the name didn't work 🙄 :(

ryan-blunden:
What about `campaign.yml`?

mrnugget:
> What about `campaign.yml`?

I think that would be confusing, since we what's in the file is an _action definition_ to produce patches which can then be used to create a campaign.



> Everything with action in the name didn't work 🙄 :(

Even `src.action.yml`, `action-definition.yml`, `campaign-action.yml`?

Since you include a schema too, how is that now a bug in the editor?

felixfbecker:
Can we do a PR to VS Code (or the responsible extension) to fix that? I imagine it could be a quick fix

cmd/src/actions_create.go
```diff
@@ -7,27 +7,29 @@ import (
...
-  ]
-}
+const actionDefinitionTemplate = `$schema: https://raw.githubusercontent.com/sourcegraph/src-cli/master/schema/actions.schema.json
+
+scopeQuery: ""
+
+steps:
+- type: command
+  args:
+  - /bin/sh
```
mrnugget, line 17: That's not a good example since it's not immediately clear whether the args to `bin/sh` end up on this line or not.

cmd/src/actions_create.go
```diff
@@ -7,27 +7,29 @@ import (
...
 `
 
 func init() {
 	usage := `
-Create an empty action definition in action.json (if not -o flag is given). This command is meant to help with creating action definitions to be used with 'src actions exec'.
+Create an empty action definition in action.yml (if not -o flag is given). This command is meant to help with creating action definitions to be used with 'src actions exec'.
```
mrnugget, line 22:
```suggestion
Create an empty action definition in action.yml (if no -o flag is given). This command is meant to help with creating action definitions to be used with 'src actions exec'.
```

cmd/src/actions_exec.go
```diff
@@ -184,13 +185,19 @@ Format of the action JSON files:
 			return err
 		}
 
-		err = validateActionDefinition(actionFile)
+		// Convert action file to JSON.
+		jsonActionFile, err := yaml.YAMLToJSONStrict(actionFile)
```
mrnugget, line 189: Does this break with JSON files? Because that would be a blocker. We have to be backwards compatible.

cmd/src/actions_scope_query.go
```diff
@@ -52,8 +53,19 @@ Examples:
 			return err
 		}
 
+		// Convert action file to JSON, if it was yaml.
+		jsonActionFile, err := yaml.YAMLToJSONStrict(actionFile)
```
mrnugget, line 57: Same here: we need to be backwards compatible and still allow reading JSON files.

cmd/src/actions_exec.go
felixfbecker: I think every JSON file should also be a valid YAML file. But if we support C-style `//` line comments in JSON, that would break.

eseliger: Correct, I figured that would be fine, but am open to discuss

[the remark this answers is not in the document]
eseliger: The idea is to be as safe as possible, and the one more roundtrip is quite cheap, so I figured let's err on the safe side

cmd/src/actions_create.go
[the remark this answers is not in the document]
eseliger: sure

ryanslade:
Sorry if this comment is a little late, but what are the arguments for supporting YAML? I know it's been requested by a couple of people but is that enough of a reason to introduce it?

I think @mrnugget said it well in the past with regards to the complexity of YAML itself and on top of that we'll now need to support both YAML and JSON.

I'm happy to be persuaded otherwise, but I currently feel that the added complexity might not be worth it.

ryanslade:
> Quite a few people asked for YAML, especially as a burden on getting a first action up quickly (need to figure out escaping in scopeQuery, no multi line strings, so harder to write a whole script, ..)
> 
> In my understanding, the usage of YAML should not make it much more difficult to maintain, as json is compatible to yaml, and as seen in this PR, I think the complexity added is quite low.
> 
> Do you think it is unreasonable to do so?

I guess having dealt with walls of Kubernetes YAML has left me with a bit of a negative view of it :)

Agreed that the code complexity introduced here is low which I'm happy about. I guess as long as we're careful to keep the scope of our action definitions fairly narrow we should be ok.

sqs:
I personally agree with YAML being the best for widespread adoption. (And of course we can experiment with DSLs at any time, that need not block this.)

I was the one who chose JSON originally for simplicity/quick prototyping. My being the original author doesn't give me any special weight in this decision now; I'm just stating that I didn't have any deeper reasons behind choosing JSON over YAML.

I do also quite like how VS Code (and now other editors) makes editing JSON documents with a JSON Schema quite easy, with completion and validation. And with JSONC (comments + trailing commas), it's not bad. But campaign actions will have tons of multi-line strings with quotes and backslashes, and doing those in JSON(C) is completely terrible. That downside of JSON(C) outweighs anything else, IMO.

felixfbecker:
@sqs just as an FYI: Since YAML is very similar to JSON from a semantic perspective, JSON schemas work to describe YAML too. If you install https://github.com/redhat-developer/vscode-yaml you actually get the same autocompletion, validation etc for YAML (powered by JSON schemas). After I found that extension writing YAML became a real pleasure for me.

sqs:
@felixfbecker Great. I tried that a year ago and it just didn't work for some weird reason (I think because it failed to fetch the JSON Schema doc properly given a URL). Haven't tried it since.

felixfbecker:
It doesn't have `$schema` support implemented afaik, but it "just works" if the schema is registered in the [JSON schema store](http://schemastore.org/json/). Alternatively you can define custom schema URLs with glob patterns in VS Code settings.

PR #203, opened 2020-05-19
Do not attempt multipart LSIF upload pre-3.16

efritz:
This probably should have made it into the 3.12 src-cli release so that we maintain large-upload backwards compat (but this is trading one error for another in most cases).

I'll release 3.12.1 with this fix so that this version can still work with pre-3.16 Sourcegraph instances.

### DIFF
--- a/cmd/src/version.go
+++ b/cmd/src/version.go
@@ -92,3 +92,33 @@ func getRecommendedVersion() (string, error) {
 
 	return payload.Version, nil
 }
+
+//
+//
+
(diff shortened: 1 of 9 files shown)


cmd/src/version.go
```diff
@@ -92,3 +92,33 @@ func getRecommendedVersion() (string, error) {
 
 	return payload.Version, nil
 }
+
+//
+//
+
```
mrnugget, line 98:
```suggestion
```

efritz:
Closing as we shouldn't need to maintain backcompat for this.

Issue #156, opened 2020-03-07
Deprecate config file


eseliger:
~To be aligned with what other tools do, I want to propose that we rename the src-cli config file to a hidden file.~

See discussion below, we've decided to deprecate the config file.

creachadair:
I like this idea in general, but rather than add yet-another hidden file to `$HOME` I suggest we use a more principled location, e.g. https://godoc.org/os#UserConfigDir.

ryan-blunden:
As a macOS User, I always expect config files for CLI tools to be in the root as a hidden (dot) file. Configure for macOS apps on the other hand I do expect to be in the ‘Library/Application Support’ folder.

Regardless, does the src CLI default help output contain the path to the configuration file for the platform?

slimsag:
In retrospect, I definitely regret placing the config file in `$HOME/src-config.json` -- that was silly of me. The `src` CLI was a quick 4hr hack project initially, I briefly considered a hidden file but thought that would make it harder to discover for some naive reason. 

Alternative proposal: Deprecate the config file entirely. I no longer believe it is the right method for configuring the `src` CLI because:

1. In scripting use cases, it is almost always the case that you will want to configure it with env variables. I have never once observed anyone prefer doing this via a config file.
2. In dev environment setups, configuring env variables is far easier and more familiar than writing config files in JSON.
3. The original motivation for using a config file was that "you might want to change the endpoint depending on the context you are working in (closed source vs. open source code)" but this is easier done with dotenv or similar.

This could be done by simply:

1. Undocumenting the `-config` flag
2. Ensuring the env variables work as expected and are documented in the README and anywhere else (check docs.sourcegraph.com).

@ryan-blunden Yes, `src -h` does report it:

```
[...]
The options are:
	-config=$HOME/src-config.json    specifies a file containing {"accessToken": "<secret>", "endpoint": "https://sourcegraph.com"}
[...]
```

mrnugget:
> Alternative proposal: Deprecate the config file entirely. I no longer believe it is the right method for configuring the `src` CLI because:

I think this is the better approach. I configure the `src` CLI with `.env` files in different folders and it works really well, especially since all the other tokens/URLs for other tools/external-services are also configured right next to them.

ryan-blunden:
Sounds good to me!

ryanslade:
I'd like to take this for 3.17

sqs:
[...]

PR #205, opened 2020-05-26
Deprecate config file in favour of env variables

ryanslade:
We still read from the config file but it has been undocumented so
that we can remove it at a later point.

Users should instead use the `SRC_ACCESS_TOKEN` and 
`SRC_ENDPOINT` environment variables.

Part of: https://github.com/sourcegraph/src-cli/issues/156


ryanslade:
I removed the docs, would it have been better to add a DEPRECATED warning instead?

Issue #190, opened 2020-04-30
Include progress bar in verbose mode


ryanslade:
The verbose mode of src action exec has no indicator of current progress. You don't know what's currently happening and how far along you are. 

Idea: progress bar at the bottom, like cargo has

ryanslade:
We want something that looks like this:

```text
Running [==================>                                      ] 24/72: jobs(5)

24/72 is the number of total steps completed
jobs(5) indicates how many jobs are currently running

```

PR #209, opened 2020-05-28
Display a progress bar in verbose mode

ryanslade:
progress

Closes: https://github.com/sourcegraph/src-cli/issues/190

### DIFF
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -47,6 +47,7 @@ func (x *actionExecutor) do(ctx context.Context, repo ActionRepo) (err error) {
 			status := ActionRepoStatus{Cached: true, Patch: result}
 			x.updateRepoStatus(repo, status)
 			x.logger.RepoCacheHit(repo, status.Patch != PatchInput{})
+			x.logger.progress.IncStepsComplete(int64(len(x.action.Steps)))
 			return nil
 		}
 	}
--- a/cmd/src/actions_exec_backend_runner.go
+++ b/cmd/src/actions_exec_backend_runner.go
@@ -46,8 +46,7 @@ func (x *actionExecutor) do(ctx context.Context, repo ActionRepo) (err error) {
 		} else if ok {
 			status := ActionRepoStatus{Cached: true, Patch: result}
 			x.updateRepoStatus(repo, status)
-			x.logger.RepoCacheHit(repo, status.Patch != PatchInput{})
-			x.logger.progress.IncStepsComplete(int64(len(x.action.Steps)))
+			x.logger.RepoCacheHit(repo, len(x.action.Steps), status.Patch != PatchInput{})
 			return nil
 		}
 	}
--- a/cmd/src/actions_exec_logger.go
+++ b/cmd/src/actions_exec_logger.go
@@ -240,7 +240,7 @@ func (a *actionLogger) DockerStepDone(repoName string, step int, elapsed time.Du
 	a.write(repoName, yellow, "%s Done. (%s)\n", boldBlack.Sprintf("[Step %d]", step), elapsed)
 }
 
-// write writes to RepoWriter and if in verbose mode, to stderr
+// write writes to the RepoWriter associated with the given repoName and logs the message using the log method.
 func (a *actionLogger) write(repoName string, c *color.Color, format string, args ...interface{}) {
 	if w, ok := a.RepoWriter(repoName); ok {
 		fmt.Fprintf(w, format, args...)
--- a/cmd/src/actions_exec_logger.go
+++ b/cmd/src/actions_exec_logger.go
@@ -33,8 +33,8 @@ type actionLogger struct {
 
 	highlight func(a ...interface{}) string
 
-	progress       *progress
-	progressWriter *progressWriter
+	progress *progress
+	out      io.Writer
 
 	mu         sync.Mutex
 	logFiles   map[string]*os.File
@@ -46,20 +46,20 @@ func newActionLogger(verbose, keepLogs bool) *actionLogger {
 	if useColor {
 		color.NoColor = false
 	}
+
 	progress := new(progress)
-	progressWriter := &progressWriter{
-		p: progress,
-		w: os.Stderr,
-	}
 
 	return &actionLogger{
-		verbose:        verbose,
-		keepLogs:       keepLogs,
-		highlight:      color.New(color.Bold, color.FgGreen).SprintFunc(),
-		progress:       progress,
-		progressWriter: progressWriter,
-		logFiles:       map[string]*os.File{},
-		logWriters:     map[string]io.Writer{},
+		verbose:   verbose,
+		keepLogs:  keepLogs,
+		highlight: color.New(color.Bold, color.FgGreen).SprintFunc(),
+		progress:  progress,
+		out: &progressWriter{
+			p: progress,
+			w: os.Stderr,
+		},
+		logFiles:   map[string]*os.File{},
+		logWriters: map[string]io.Writer{},
 	}
 }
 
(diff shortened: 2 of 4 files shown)


cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +225,106 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+func (p *progress) Write(data []byte) (int, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	length := 50
+	if p.shouldClear {
+		// Clear current progress
+		fmt.Fprintf(p.w, "\r")
+		// TODO: We could be more precise here by taking into account the number of steps / jobs
+		fmt.Fprintf(p.w, strings.Repeat(" ", 100))
```
mrnugget, line 269: Can't you remember the length of the last line that was written and then after writing the new line check whether the new line is shorter than the old one and, if so, emit as many spaces as possible?

cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +225,106 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+
+	if a.verbose {
+		if len(repoName) > 0 {
+			format = fmt.Sprintf("%s -> %s", c.Sprint(repoName), format)
+		}
+		fmt.Fprintf(a.progress, format, args...)
+	}
+}
+
+type progress struct {
```
mrnugget, line 250: Hmmm, I think the API of this could be better.

`progress` is a really generic term and doesn't tell me what it does. But it seems to do lots of things: act as an IO pipe, track progress, format progress, clean previously written output.

How about splitting it up? For example, `progress` could be turned into a `progressBar` which only tracks the progress and has a `String() string` method that returns the correctly formatted bar for current progress.

The logger can then take care of when to print that progress bar and when not.

cmd/src/actions_exec_backend_runner.go
```diff
@@ -47,6 +47,7 @@ func (x *actionExecutor) do(ctx context.Context, repo ActionRepo) (err error) {
 			status := ActionRepoStatus{Cached: true, Patch: result}
 			x.updateRepoStatus(repo, status)
 			x.logger.RepoCacheHit(repo, status.Patch != PatchInput{})
+			x.logger.progress.IncStepsComplete(int64(len(x.action.Steps)))
```
mrnugget, line 50: Why not do that inside `x.logger.RepoCacheHit`? It seems a bit unnecessary for `actionExecutor` needing to know about the `x.logger.progress`.

mrnugget: That also applies to the `x.logger.progress.SetTotalSteps`/`x.logger.progress.IncJobs()`/... calls above. I think _how_ that's reported and tracked should be concern of the logger and hidden from `actionExecutor`

cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +225,106 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+	mu          sync.Mutex
+	w           io.Writer
+	shouldClear bool
+}
+
+func (p *progress) Write(data []byte) (int, error) {
+	p.mu.Lock()
+	defer p.mu.Unlock()
+
+	length := 50
```
mrnugget, line 264: I'd suggest moving this definition down to where it's first used and calling it `maxLength` or something like that, because `length` without any context could be anything 😄

cmd/src/actions_exec_logger.go
```diff
@@ -101,18 +109,17 @@ func (a *actionLogger) ActionSuccess(patches []PatchInput, newLines bool) {
...
-		grey.Fprintf(os.Stderr, format, args...)
+		a.write("", grey, format, args...)
 	}
 }
 
 func (a *actionLogger) RepoCacheHit(repo ActionRepo, patchProduced bool) {
 	if a.verbose {
 		if patchProduced {
-			fmt.Fprintf(os.Stderr, "%s -> Cached result found: using cached diff.\n", boldGreen.Sprint(repo.Name))
+			a.write(repo.Name, boldGreen, "Cached result found: using cached diff.\n")
```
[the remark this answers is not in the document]
ryanslade: That wasn't intentional, I've split it out into two functions now `log` that logs and `write` that writes and logs

cmd/src/actions_exec_logger.go
ryanslade: I split it up into `progress` that tracks the progress state and `progressWriter` that uses that state to print a progress bar after performing it's write.

We need to pass in two writers for stderr and stdout here:
https://github.com/sourcegraph/src-cli/blob/1cd4ce40d15bfce1369fd508640e742d606acf2a/cmd/src/actions_exec_logger.go#L163-L167

And those need access to the same overall progress state.

ryanslade:
@mrnugget PTAL

ryanslade:
New look:

Screenshot 2020-05-29 at 10 46 10

cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +230,126 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+	}
+	bar += strings.Repeat(" ", maxLength-len(bar))
+	progessText := fmt.Sprintf("[%s] steps(%d/%d) %s %s", bar, w.p.StepsComplete(), w.p.TotalSteps(), boldRed.Sprintf("failed(%d)", w.p.TotalStepsFailed()), hiGreen.Sprintf("patches(%d)", w.p.PatchCount()))
+	fmt.Fprintf(w.w, progessText)
+	w.shouldClear = true
+	w.progressLogLength = len(progessText)
+	return n, err
+}
+
+func (p *progress) SetTotalSteps(n int64) {
```
mrnugget, line 325: Nitpicky nitpick: I'd move the method definitions on `*progress` up to the definition of `*progress`.

cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +230,126 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+		pctDone = float64(done) / float64(total)
+	}
+
+	maxLength := 50
+	bar := strings.Repeat("=", int(float64(maxLength)*pctDone))
+	if len(bar) < maxLength {
+		bar += ">"
+	}
+	bar += strings.Repeat(" ", maxLength-len(bar))
+	progessText := fmt.Sprintf("[%s] steps(%d/%d) %s %s", bar, w.p.StepsComplete(), w.p.TotalSteps(), boldRed.Sprintf("failed(%d)", w.p.TotalStepsFailed()), hiGreen.Sprintf("patches(%d)", w.p.PatchCount()))
```
mrnugget, line 318:
```suggestion
	progessText := fmt.Sprintf("[%s] Steps: %d/%d (%s, %s)", bar, w.p.StepsComplete(), w.p.TotalSteps(), boldRed.Sprintf("%d failed", w.p.TotalStepsFailed()), hiGreen.Sprintf("%d patches", w.p.PatchCount()))
```

cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +230,126 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+
+// write writes to RepoWriter and if in verbose mode, to stderr
+func (a *actionLogger) write(repoName string, c *color.Color, format string, args ...interface{}) {
+	if w, ok := a.RepoWriter(repoName); ok {
+		fmt.Fprintf(w, format, args...)
+	}
+	a.log(repoName, c, format, args...)
+}
+
+// log logs only to stderr, it does not log to our repoWriters
```
mrnugget, line 251:
```suggestion
// log logs only to stderr, it does not log to our repoWriters. When not in verbose mode, it's a noop.
```

cmd/src/actions_exec_logger.go
```diff
@@ -227,9 +230,126 @@ func (a *actionLogger) DockerStepStarted(repoName string, step int, image string
...
+	a.progress.IncStepsFailed()
 	a.write(repoName, boldRed, "%s %s. (%s)\n", boldBlack.Sprintf("[Step %d]", step), err, elapsed)
 }
 
 func (a *actionLogger) DockerStepDone(repoName string, step int, elapsed time.Duration) {
+	a.progress.IncStepsComplete(1)
 	a.write(repoName, yellow, "%s Done. (%s)\n", boldBlack.Sprintf("[Step %d]", step), elapsed)
 }
+
+// write writes to RepoWriter and if in verbose mode, to stderr
```
mrnugget, line 243:
```suggestion
// write writes to the RepoWriter associated with the given repoName and logs the message using the log method.
```

cmd/src/actions_exec_logger.go
```diff
@@ -31,35 +33,49 @@ type actionLogger struct {
 
 	highlight func(a ...interface{}) string
 
+	progress       *progress
+	progressWriter *progressWriter
```
mrnugget, line 37: I don't think there's a reason for actionLogger to know about `*progressWriter`, right?

```suggestion
	out io.Writer
```

cmd/src/actions_exec_logger.go
ryanslade: Ah, yeah, missed that after splitting it up

PR #211, opened 2020-05-29
migrate Configuration usage to Settings

bobheadxi:
closes https://github.com/sourcegraph/sourcegraph/issues/10419

looks like `ConfigurationSubject` was removed a while back, and `SettingsSubject` has replaced it

> SettingsSubject is something that can have settings: a site ("global settings", which is different from "site configuration"), an organization, or a user.

```
go build ./cmd/src
./src config list
./src config get
```

a variety of other `Configuration`-related fields have been listed as deprecated, so I've gone ahead and migrated them to use `Settings`

### DIFF
--- a/cmd/src/config.go
+++ b/cmd/src/config.go
@@ -50,7 +50,7 @@ Use "src config [command] -h" for more information about a command.
 }
 
 const configurationSubjectFragment = `
-fragment ConfigurationSubjectFields on ConfigurationSubject {
+fragment ConfigurationSubjectFields on SettingsSubject {
     id
     latestSettings {
         id
--- a/cmd/src/config.go
+++ b/cmd/src/config.go
@@ -54,7 +54,7 @@ fragment SettingsSubjectFields on SettingsSubject {
     id
     latestSettings {
         id
-		contents
+        contents
         author {
             ...UserFields
         }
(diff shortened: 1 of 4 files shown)


cmd/src/config.go
```diff
@@ -50,7 +50,7 @@ Use "src config [command] -h" for more information about a command.
 }
 
 const configurationSubjectFragment = `
-fragment ConfigurationSubjectFields on ConfigurationSubject {
+fragment ConfigurationSubjectFields on SettingsSubject {
```
mrnugget, line 53: To avoid others being confused by this in the future, let's make the naming consistent 🙂

```suggestion
fragment SettingsSubjectFields on SettingsSubject {
```

bobheadxi: I dug a bit deeper and the changes turned out to be pretty-far reaching - all the `Configuration` fields are listed as deprecated, and the shape of some things have changed, so I went ahead and made everything `Settings`: 9a39f3e59ec73c7c282d3cdbdf73a501c65e4203

slimsag:
Once you merge, please create a release following these instructions: https://github.com/sourcegraph/src-cli/blob/master/DEVELOPMENT.md#releasing

mrnugget:
Good stuff, thanks @bobheadxi!

Issue #202, opened 2020-05-15
Include instructions after src action exec


ryanslade:
Add some instructions to src action exec when everything went well and we print a patch to stdout. 

Print instructions to STDERR after the json has been dumped.

mrnugget:
A thought I had: maybe we should rethink dumping the patch to stdout and instead use a default location, like `./patches.json`, that can be overwritten with `-o`. The message is then something like "Patches saved to `patches.json`. Use 'src action patchset create-from-patches -f patches.json` ..."

ryanslade:
If we make creating a file the default do we still want to handle the case where they pipe the output to create patches? eg:

```
src actions exec -f ~/run-gofmt.json | src campaign patchset create-from-patches
```

We could check if stdout is pointing at a pipe 'tee' output there too in that case?

mrnugget:
> We could check if stdout is pointing at a pipe 'tee' output there too in that case?

I think that's a good default, yes.

- If `-o filename` is given, write to `filename`.
- If no `-o filename` is given and stdout is not a pipe, write to `patches.json`.
- If no `-o filename` is given and stdout _is_ a pipe, write to pipe.

ryanslade:
Yep, working on the instructions now

PR #214, opened 2020-06-03
By default, write patches to an output file

ryanslade:
If output of `src actions exec` is piped we'll write the patches to stdout. Otherwise we'll use the file specified by the `-o` flag, defaulting to `patches.json'.

Closes: https://github.com/sourcegraph/src-cli/issues/202

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


cmd/src/actions_exec.go
```diff
@@ -80,6 +81,10 @@ Examples:
 
 	$ src actions exec -f ~/run-gofmt.json | src campaign patchset create-from-patches
 
+  Execute an action and save the patches it produced to 'patches.json'
+
+	$ src actions exec -f ~/run-gofmt.json -o patches.json 
+
```
mrnugget, line 87: Can you add a comment to the first example in this description, without `-create-patchset` & without a pipe, that the patches are saved by default into `patches.json`?

ryanslade:
> I wonder whether we shouldn't introduce this rule: `If -o filename is given but stdout is a pipe, write to pipe.`
> 
> That would make the code easier, since you could use `patches.json` as the default value for the flag and then your logic would be: "if pipe; use pipe; else; use outputFlag"

Agreed. If someone is piping the output they're going to be surprised if instead of writes to a file

Issue #222, opened 2020-06-09
Env var SRC_ENDPOINT is ignored when config file exists


sqs:
If `~/src-config.json` exists, the env var `SRC_ENDPOINT` is ignored (see https://sourcegraph.com/github.com/sourcegraph/src-cli@6f6674f8e67e380efe64cd53662f342cd4604a74/-/blob/cmd/src/main.go#L106). This is confusing. Also, `SRC_ACCESS_TOKEN` behaves in the opposite way; it *overrides* the value set in the config file.

I believe the ideal behavior is:

- Both env vars override the config file.
- If a config file exists, either zero or both env vars must be specified. It is an error if only one is specified. This is to avoid confusion; we never merge config from the file and env vars.

Reported by @nicksnyder in https://sourcegraph.slack.com/archives/CMMTWQQ49/p1591657309056300.

mrnugget:
@ryanslade Wasn't this fixed with your #205?

ryanslade:
No, that change only removed the documentation regarding the config file as a first step before removing support for it. 

We also have the `-endpoint` flag. I think maybe we should remove that and then follow the rules set out by @sqs above.

I'm happy to take this as I worked on the code recently, assign it to me if you're happy with that.

ryanslade:
@sqs Are you happy with the above suggestion, specifically removing the `-endpoint` flag?

sqs:
Yep :+1:

ryanslade:
We need to keep the endpoint flag around for now as some of our LSIF actions depend on it. 

I'll submit another PR that removes it from docs but still respects it.

I think the simplest way to do this is to follow the same logic as before but if the flag has been set it'll override both the config file and environment variables.

burmudar:
Closing this issue as part of repository maintenance. We are no longer using GitHub issues to track src-cli requests or reports. For future questions or support requests, please contact Sourcegraph Support (mailto:support@sourcegraph.com).

PR #225, opened 2020-06-10
Remove -endpoint flag

ryanslade:
This change removes the -endpoint flag

It also ensures the following:

* Both env vars override the config file.
* If a config file exists, either zero or both env vars must be
  specified. It is an error if only one is specified. This is
  to avoid confusion; we never merge config from the file and env vars.

Closes: https://github.com/sourcegraph/src-cli/issues/222

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,22 @@
+<!--
+###################################### READ ME ###########################################
+### This changelog should always be read on `master` branch. Its contents on version   ###
+### branches do not necessarily reflect the changes that have gone into that branch.   ###
+##########################################################################################
+-->
+
+# Changelog
+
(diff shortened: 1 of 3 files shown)


ryanslade:
Some of our LSIF docs mention `-endpoint` so we'll need to update that after this is released

slimsag:
Adding more color to my comment: I think just documenting this in a new CHANGELOG.md file would be good enough.

If you want to get more fancy, you could detect if `-endpoint` is passed in code and error out / say it is deprecated - but I don't think that's needed.

ryanslade:
I've added a skeleton changelog, I assume we didn't want any historical changes in there?

mrnugget:
Yep, looks good!

PR #226, opened 2020-06-17
Reduce memory pressure on frontend

ryanslade:
This change updates the actionRepos method so that it doesn't request
default branch details for every search result.

Instead, after fetching search results it does a further round of
requests in batches of 100 requesting default branch details from the
deduplicated list of repos matched.

Part of: https://github.com/sourcegraph/sourcegraph/issues/10540

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -471,7 +471,9 @@ func actionRepos(ctx context.Context, scopeQuery string, includeUnsupported bool
 		scopeQuery = scopeQuery + " count:999999"
 	}
 
-	query := `
+	// NOTE: This query does not fetch default branch details.
+	// Once we have the list of matching repos we'll fetch the branches in another query
+	searchQuery := `
 query ActionRepos($query: String!) {
 	search(query: $query, version: V2) {
 		results {
@@ -496,14 +498,9 @@ fragment repositoryFields on Repository {
 	externalRepository {
 		serviceType
 	}
-	defaultBranch {
-		name
-		target {
-			oid
-		}
-	}
 }
 `
+
 	type Repository struct {
 		ID, Name           string
 		ExternalRepository struct {
@@ -514,7 +511,8 @@ fragment repositoryFields on Repository {
 			Target struct{ OID string }
 		}
 	}
-	var result struct {
+
+	var searchResult struct {
 		Data struct {
 			Search struct {
 				Results struct {
@@ -540,38 +538,44 @@ fragment repositoryFields on Repository {
 		} `json:"errors,omitempty"`
 	}
 
-	if err := (&apiRequest{
-		query: query,
-		vars: map[string]interface{}{
-			"query": scopeQuery,
-		},
-		// Do not unpack errors and return error. Instead we want to go through
-		// the results and check whether they're complete.
-		// If we don't do this and the query returns an error for _one_
-		// repository because that is still cloning, we don't get any repositories.
-		// Instead we simply want to skip those repositories that are still
-		// being cloned.
-		dontUnpackErrors: true,
-		result:           &result,
-	}).do(); err != nil {
-
-		// Ignore exitCodeError with error == nil, because we explicitly set
-		// dontUnpackErrors, which can lead to an empty exitCodeErr being
-		// returned.
-		exitCodeErr, ok := err.(*exitCodeError)
-		if !ok {
-			return nil, err
+	// Helper function to perform a graphql request
+	doRequest := func(query string, vars map[string]interface{}, dest interface{}) error {
+		if err := (&apiRequest{
+			query: query,
+			vars: map[string]interface{}{
+				"query": scopeQuery,
+			},
+			// Do not unpack errors and return error. Instead we want to go through
+			// the results and check whether they're complete.
+			// If we don't do this and the query returns an error for _one_
+			// repository because that is still cloning, we don't get any repositories.
+			// Instead we simply want to skip those repositories that are still
+			// being cloned.
+			dontUnpackErrors: true,
+			result:           dest,
+		}).do(); err != nil {
+			// Ignore exitCodeError with error == nil, because we explicitly set
+			// dontUnpackErrors, which can lead to an empty exitCodeErr being
+			// returned.
+			exitCodeErr, ok := err.(*exitCodeError)
+			if !ok {
+				return err
+			}
+			if exitCodeErr.error != nil {
+				return exitCodeErr
+			}
 		}
-		if exitCodeErr.error != nil {
-			return nil, exitCodeErr
-		}
+		return nil
 	}
 
-	skipped := []string{}
-	unsupported := []string{}
-	reposByID := map[string]ActionRepo{}
-	for _, searchResult := range result.Data.Search.Results.Results {
+	err = doRequest(searchQuery, map[string]interface{}{"query": scopeQuery}, &searchResult)
+	if err != nil {
+		return nil, err
+	}
 
+	unsupported := make(map[string]struct{})
+	reposByID := map[string]*ActionRepo{}
+	for _, searchResult := range searchResult.Data.Search.Results.Results {
 		var repo Repository
 		if searchResult.Repository.ID != "" {
 			repo = searchResult.Repository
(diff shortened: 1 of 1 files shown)


eseliger:
Did that actually help? 
In my investigations in https://github.com/sourcegraph/sourcegraph/issues/10540, the most memory consumption came from massive zoekt results stored in memory, not from the repository resolvers (which are also deduped on the backend now) 🤔 Do you have any sample queries that work better than before?

ryanslade:
When running the old code, even with your dedupe change the frontend crashes and returns a 502 error. I can then see in Grafana that the frontend was killed and restarted and sometimes if metrics publish occurred before the crash we see a spike in memory usage.

When I run with this branch we get our results without a crash and memory usage on the frontend only spikes a little.

I've been testing with this action, saved as `oom.json` against our k8s dogfood instance:

```
{
  "scopeQuery": "lang:go repo:sourcegraph",
  "steps": [
    {
      "type": "docker",
      "image": "comby/comby",
      "args": ["-in-place", "fmt.Sprintf(\"%d\", :[v])", "strconv.Itoa(:[v])", "-matcher", ".go", "-d", "/work"]
    },
    {
      "type": "docker",
      "image": "cytopia/goimports",
      "args": ["-w", "/work"]
    }
  ]
}
```
 and this command:

```
src -v actions scope-query -f oom.json
```

mrnugget:
What are the performance implications for `src actions scope-query`? Is it noticeable slower since we make more requests?

cmd/src/actions_exec.go
```diff
@@ -471,7 +471,9 @@ func actionRepos(ctx context.Context, scopeQuery string, includeUnsupported bool
 		scopeQuery = scopeQuery + " count:999999"
 	}
 
-	query := `
+	// NOTE: This query does not fetch default branch details.
+	// Once we have the list of matching repos we'll fetch the branches in another query
```
mrnugget, line 475: I don't think this comment is necessary :)

cmd/src/actions_exec.go
```diff
@@ -586,38 +590,116 @@ fragment repositoryFields on Repository {
...
+								OID string `json:"oid"`
+							} `json:"target"`
+						} `json:"defaultBranch"`
+					} `json:"nodes"`
+				} `json:"repositories"`
+			} `json:"data,omitempty"`
+
+			Errors []struct {
+				Message string
+				Path    []interface{}
```
mrnugget, line 647: I don't think we need them, but: you added `json:""` tags to every other field but not here.

cmd/src/actions_exec.go
```diff
@@ -540,38 +538,44 @@ fragment repositoryFields on Repository {
...
-			return nil, err
-		}
-		if exitCodeErr.error != nil {
-			return nil, exitCodeErr
+	// Helper function to perform a graphql request
+	doRequest := func(query string, vars map[string]interface{}, dest interface{}) error {
+		if err := (&apiRequest{
+			query: query,
+			vars: map[string]interface{}{
+				"query": scopeQuery,
```
mrnugget, line 546: The `vars` you pass in to `doRequest` are not used. Instead you always pass in this set of `vars`.

cmd/src/actions_exec.go
```diff
@@ -586,38 +590,116 @@ fragment repositoryFields on Repository {
...
+}
+
+`
+
+	skipped := make(map[string]struct{})
+
+	// This helper function fetches a subset of repos by name and is used lower down
+	// so that we query in batches.
+	fetchRepos := func(names []string) error {
+		reposQuery := fmt.Sprintf(reposQueryTemplate, strings.Join(names, ","))
```
mrnugget, line 627: Any reason why you use template+`fmt.Sprintf` instead of using a GraphQL var and passing the `names` to `doRequest`?

cmd/src/actions_exec.go
```diff
@@ -586,38 +590,116 @@ fragment repositoryFields on Repository {
...
+			}
 		}
+	}
 
-		if repo.DefaultBranch.Target.OID == "" {
-			skipped = append(skipped, repo.Name)
-			continue
+	reposQueryTemplate := `{
+  repositories(names: [%s]) {
+    nodes {
```
mrnugget, line 607: Since this is a connection — does this always return all default branches, or do we need to check for `pageInfo.HasNextPage`?

cmd/src/actions_exec.go
```diff
@@ -586,38 +590,116 @@ fragment repositoryFields on Repository {
...
+				skipped[repo.Name] = struct{}{}
+				continue
+			}
+			existing, ok := reposByID[repo.ID]
+			if !ok {
+				continue
 			}
+			existing.Rev = repo.DefaultBranch.Target.OID
+			existing.BaseRef = repo.DefaultBranch.Name
+		}
```
mrnugget, line 671: I find it hard to follow the control flow in this function. I think moving this loop out of `fetchRepos` and have `fetchRepos` just return `repoResult.Data.Repositories.Nodes` and then doing the looping in the other loop further down makes it easier to follow what's going on.

cmd/src/actions_exec.go
```diff
@@ -586,38 +590,116 @@ fragment repositoryFields on Repository {
...
+
+		return nil
+	}
+
+	repoNames := make([]string, 0, len(reposByID))
+	for _, r := range reposByID {
+		repoNames = append(repoNames, `"`+r.Name+`"`)
```
mrnugget, line 678: Nitpicky, but I personally to keep the number of quotes per line to a minimum 😄 

```suggestion
		repoNames = append(repoNames, fmt.Sprintf("%q", r.Name))
```

ryanslade:
**WIP NOTES**

Some notes on trying to pin down why we're causing an OOM trigger

Using the following query and altering the count param:

```graphql
{
  search(query: "lang:go repo:sourcegraph count:140000", version: V2) {
    results {
      matchCount
      results {
        __typename
        ... on Repository {
          ...repositoryFields
        }
        ... on FileMatch {
          repository {
            ...repositoryFields
          }
        }
      }
    }
  }
}

fragment repositoryFields on Repository {
  id
  name
  externalRepository {
    serviceType
  }
  defaultBranch {
    name
    target {
      oid
    }
  }
}

```

Initial query caused a memory usage to grow, but not in one large spike. It rose in increments and then started to trail off. Subsequent queries didn't seem to cause it to spike a second time.

Screenshot 2020-06-18 at 14 27 59

I performed multiple queries. increasing the count by 5000 from 120000 to 140000

cmd/src/actions_exec.go
ryanslade: We need double quotes here because it's used as a parameter for a GraphQL query

mrnugget: Yeah, I know and `%q` gives you those? Or am I missing something?

cmd/src/actions_exec.go
ryanslade: Missed that during a refactor, I'll update it

ryanslade:
I can reproduce this locally but the allocations don't appear using either `pprof` or MacOS instruments.

When I search including `defaultBranch` memory usage jump from under 100MB to over 700MB and then slowly drops back down. It consistently spikes after every query,  but peaks at around 700MB.

When commenting out `defaultBranch` from the query it stays under 100MB.

mrnugget:
That sounds like it has an effect. I think we should merge it, although it's weird that you cannot see the allocations. Have you tried playing around with [runtime variables](https://golang.org/pkg/runtime/) to peek inside the GC, like `gctrace=1`?

ryanslade:
I'll try that this morning. I think I'm pretty close to finding the root cause of this so I'm going to give it another half a day so that we can hopefully fix it in the resolver layer instead.

ryanslade:
Here are some lines from gctrace=1 immediatly after running the query:

```text
gc 19 @71.823s 0%: 0.025+4.8+0.034 ms clock, 0.30+0.33/13/33+0.41 ms cpu, 64->65->43 MB, 65 MB goal, 12 P
gc 20 @71.969s 0%: 0.15+37+0.34 ms clock, 1.8+229/86/0+4.1 ms cpu, 84->86->61 MB, 86 MB goal, 12 P
gc 21 @72.154s 0%: 0.14+102+0.007 ms clock, 1.6+483/307/342+0.089 ms cpu, 119->120->107 MB, 123 MB goal, 12 P
```

The main thing to note here is that the live heap number, the last one after the arrow peaks at 107MB whereas the OS reports the frontend process using almost 700MB.

mrnugget:
> The main thing to note here is that the live heap number, the last one after the arrow peaks at 107MB whereas the OS reports the frontend process using almost 700MB.

I would suspect that that's the difference between "live heap" and "overall memory allocated (but possibly not in use) by the process".

<skip if you're familiar with Go's GC or GCs in general>
With growing memory needs, Go allocates more memory. The GC cleans up that memory of unused data, but it doesn't necessarily give that memory back to the OS. (I think the newest versions of Go do give it back after _some_ time.)

Simplified example: you read in 100mb of data from `todos.txt`. Go's runtime allocates 100mb of memory so it can keep the contents of `todos.txt` in memory. You run `strings.Contains(todoContent, "eat icecream")` and after that, your function returns, the open file handle is closed and the `todoContent` string is unused. Go's GC cleans up the space in memory taken up `todoContent` as available. If you then read in the `todos.txt` file again, Go's runtime doesn't need to allocate another 100mb, it can reuse the now available space. From the outside, from the OS perspective, it looks like the Go processes uses 100mb all the time, even though the space was available.
</skip if you're familiar with Go's GC or GCs in general>

So, if your OS says the process uses 600mb, but Go says that its live heap is only 107mb, then ~500mb might be available to Go, but have not been given back to the OS. That indicates that at some point Go _did_ need these additional 500mb. The question is: why don't they show up in the `gctrace`?

ryanslade:
I've checked both live heap and total allocations and both stay well below the amount reported by the OS. For example, even after taking an allocation profile after running the search a few times the **total** memory allocated since program start is still well below what the OS reports. It's not OS specific since we're seeing the same issue on my laptop and in our k8s deployment. 

I also thought it could be related to the amount of data produced by our tracing infrastructure but I disabled tracing and that didn't help.

It's very strange, this is the first time I've experienced something like this where the OS is reporting significantly more usage than the Go profiler.

ryanslade:
The plot thickens.. 

When I running a query locally that returns ~18k results the frontend launches over 100k goroutines. When adding the defaultBranch option it adds a further ~18k. This doesn't explain why usage jumps so much unless the runtime is allocating extra stack space over some goroutine threshold.

However, launching over 100k goroutines seems crazy and could easily explain our spike is memory usage. It would also explain why the profiler doesn't show the memory as the goroutines are using stack space, not heap.

After the query completes we drop back down to about 100 goroutines.

ryanslade:
There's already an issue raised about this: https://github.com/graph-gophers/graphql-go/issues/375

ryanslade:
I manually disabled concurrency by editing `graphql-go` and now memory usage stays under 100MB and the query time is the same.

`graphql-go` already has a `MaxParallelism` setting but it doesn't limit the number of goroutines, just the number that are actually doing work at any time. 

I'll see if it's possible to make it limit the number of routines spawned instead.

eseliger:
Great find! So glad we dug deeper into this topic 😬

keegancsmith:
> `graphql-go` already has a `MaxParallelism` setting but it doesn't
> limit the number of goroutines, just the number that are actually
> doing work at any time.

How easy do you think it would be to implement this? I can imagine it
can be quite tricky due to the way resolvers nest. But I think if you
just limit the number of goroutines spawned to MaxParallelism on any
particular execList, you will quite effectively bound the number of
goroutines without leading into any sort of deadlocks.

I am saying this without checking the code though, so even the above may
be too much work. Do you mind taking a look?

ryanslade:
> `graphql-go` already has a `MaxParallelism` setting but it doesn't limit the number of goroutines, just the number that are actually doing work at any time.
> How easy do you think it would be to implement this? I can imagine it can be quite tricky due to the way resolvers nest. But I think if you just limit the number of goroutines spawned to MaxParallelism on any particular execList, you will quite effectively bound the number of goroutines without leading into any sort of deadlocks. I am saying this without checking the code though, so even the above may be too much work. Do you mind taking a look?

I've already taken a look at limiting the goroutine spawning but your intuition is correct. Due to the graph structure we end up in a deadlock as we walk through the graph. 

One quick, but potentially slower for some queries option, would be to add an override that causes us to always do everything synchronously. At least for the query I've been testing there's no slowdown.

Their code already has branches depending on whether `async` is enabled so we could override those spots. That's what I did manually in the previous comment and seems to work fairly well.

keegancsmith:
@ryanslade I took a quick look at the code, I think this hack may work. Can you try it out? https://github.com/keegancsmith/graphql-go/commit/3567186be2f9d0dd5b67ff68b09edd8652b07f7a

ryanslade:
> @ryanslade I took a quick look at the code, I think this hack may work. Can you try it out? [keegancsmith/graphql-go@3567186](https://github.com/keegancsmith/graphql-go/commit/3567186be2f9d0dd5b67ff68b09edd8652b07f7a)

Yep, that works, nice! I was trying to reuse their limiter but as it went deeper we'd eventually hit a deadlock.

The number of goroutines is slightly higher than just forcing it to be synchronous but not significantly (about 180 vs 120)

The next question is whether we want to initially make this change on our fork? If we do, we'll need to get it in sync with upstream which is currently about 130 commits ahead. Our main repo isn't using our fork so I think it's safe to fast forward and apply this change or something similar on top.

mrnugget:
> it's safe to fast forward and apply this change or something similar on top.

That plus opening a PR with @keegancsmith's change, that maybe uses the `MaxParallelism` value instead of a hardcoded 10, sounds good to me.

keegancsmith:
Thanks for trying out! I agree one should thread in MaxParallelism now. Additionally please also send a PR to upstream :)

Also note you can avoid the WaitGroup. Instead of `.Wait` at the end, just do 
```go
for i := 0; i < MaxParallelism; i++ {
  sem <- struct{}{}
}
```

IE you can only claim all the semaphore spots once everything else is done.

ryanslade:
@keegancsmith I've opened a PR on our repo here: https://github.com/sourcegraph/graphql-go/pull/1

(And also on upstream)

ryanslade:
Closing this in favour of: https://github.com/sourcegraph/sourcegraph/pull/11736

PR #232, opened 2020-06-24
Output more detailed JSON errors

LawnGnome:
This requires sourcegraph/jsonx#8, which we'd have to decide we wanted and approve before this could be merged.

Right now, if you load an invalid JSON file into `src-cli`, the error message is pretty rough:

image

Instead, wouldn't it be nice if we could point out the actual error(s):

image

That's what this PR (and sourcegraph/jsonx#8) enables.

### DIFF
--- a/cmd/src/extsvc_edit.go
+++ b/cmd/src/extsvc_edit.go
@@ -1,7 +1,6 @@
 package main
 
 import (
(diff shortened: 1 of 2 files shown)


mrnugget:

PR #233, opened 2020-06-25
Update jsonx dependency

LawnGnome:
This PR is the last step that fixes sourcegraph/sourcegraph#10541 in full.

However, if we decide we'd like #232, we should wait and get sourcegraph/jsonx#8 and #232 merged first, then I can update this again to that version of jsonx.


LawnGnome:
Rebased after landing sourcegraph/jsonx#7 and sourcegraph/jsonx#8, which will allow us both to fix sourcegraph/sourcegraph#10541 and set us up to review and land #232.

PR #235, opened 2020-06-30
Deprecate -endpoint flag

ryanslade:
This removes the documentation of the endpoint flag but still allows it
to be used to override either the config or ennvironment variables.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Changed
 
 - The terminal UI has been replaced by the logger-based UI that was previously only visible in verbose-mode (`-v`). [#228](https://github.com/sourcegraph/src-cli/pull/228)
-- Deprecated the `-endpoint` flag. Instead, use the `SRC_ENDPOINT` environment variable.
+- Deprecated the `-endpoint` flag. Instead, use the `SRC_ENDPOINT` environment variable. [#235](https://github.com/sourcegraph/src-cli/pull/235)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,10 @@ All notable changes to `src-cli` are documented in this file.
 ### Changed
 
 - The terminal UI has been replaced by the logger-based UI that was previously only visible in verbose-mode (`-v`). [#228](https://github.com/sourcegraph/src-cli/pull/228)
+- Deprecated the `-endpoint` flag. Instead, use the `SRC_ENDPOINT` environment variable.
 
 ### Fixed
 
 ### Removed
+
+- Remove the `-endpoint` flag. [#225](https://github.com/sourcegraph/src-cli/pull/225)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,6 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 - Pull missing docker images automatically. [#191](https://github.com/sourcegraph/src-cli/pull/191)
-
 - Searches that result in errors will now display any alerts returned by Sourcegraph, including suggestions for how the search could be corrected. [#221](https://github.com/sourcegraph/src-cli/pull/221)
 
 ### Changed
@@ -26,4 +25,3 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Removed
 
-- Remove the `-endpoint` flag. [#225](https://github.com/sourcegraph/src-cli/pull/225)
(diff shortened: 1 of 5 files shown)


efritz:
What is the original purpose of removing this (especially if we're hardening it with tests in this PR)?

uwedeportivo:
i'm using it here https://github.com/sourcegraph/deploy-sourcegraph/pull/687

cmd/src/main.go
```diff
@@ -96,23 +99,39 @@ func readConfig() (*config, error) {
...
+	if endpoint != nil && *endpoint != "" {
+		cfg.Endpoint = *endpoint
+	}
+
 	cfg.Endpoint = strings.TrimSuffix(cfg.Endpoint, "/")
 
 	return &cfg, nil
 }
+
+var errConfigMerge = errors.New("config merging not supported, zero or both environment variables must be set")
```
mrnugget, line 137: I'd make this more descriptive. As a user it's hard to know what "config merging not supported" means when I just set an env var.

```suggestion
var errConfigMerge = errors.New("when using a configuration file, zero or all environment variables must be set")
```

Or something like that.

cmd/src/main_test.go
```diff
@@ -0,0 +1,179 @@
...
+			defer func() { os.Setenv("SRC_ACCESS_TOKEN", oldToken) }()
+			oldEndpoint := os.Getenv("SRC_ENDPOINT")
+			defer func() { os.Setenv("SRC_ENDPOINT", oldEndpoint) }()
+
+			if err := os.Setenv("SRC_ACCESS_TOKEN", test.envToken); err != nil {
+				t.Fatal(err)
+			}
+			if err := os.Setenv("SRC_ENDPOINT", test.envEndpoint); err != nil {
+				t.Fatal(err)
+			}
```
mrnugget, line 165: Maybe a bit easier to parse visually and read 😄 But, I was in the mood to write a tiny function just now, feel free to ignore:

```suggestion
			setEnvVar := func(name, val string) {
				old := os.Getenv(name)
				if err := os.Setenv(name, val); err != nil {
					t.Fatal(err)
				}
				t.Cleanup(func() { os.Setenv(name, old) })
			}
			setEnvVar("SRC_ACCESS_TOKEN", test.envToken)
			setEnvVar("SRC_ENDPOINT", test.envEndpoint)
```

cmd/src/main_test.go
```diff
@@ -0,0 +1,179 @@
...
+		tmpDir, err := ioutil.TempDir("", "")
+		if err != nil {
+			t.Fatal(err)
+		}
+		filePath := filepath.Join(tmpDir, "config.json")
+		err = ioutil.WriteFile(filePath, data, 0600)
+		if err != nil {
+			t.Fatal(err)
+		}
+		return filePath, func() { os.RemoveAll(tmpDir) }
```
mrnugget, line 28: Since you already have `t` at hand, you could do this:

```suggestion
		t.Cleanup(func() { os.RemoveAll(tmpDir) })
		return filePath
```

CHANGELOG.md
```diff
@@ -20,7 +20,10 @@ All notable changes to `src-cli` are documented in this file.
 ### Changed
 
 - The terminal UI has been replaced by the logger-based UI that was previously only visible in verbose-mode (`-v`). [#228](https://github.com/sourcegraph/src-cli/pull/228)
+- Deprecated the `-endpoint` flag. Instead, use the `SRC_ENDPOINT` environment variable. [#235](https://github.com/sourcegraph/src-cli/pull/235)
 
 ### Fixed
 
 ### Removed
+
+- Remove the `-endpoint` flag. [#225](https://github.com/sourcegraph/src-cli/pull/225)
```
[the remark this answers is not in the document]
ryanslade: Ah, left over from the previous change. I've left it in "changed"

ryanslade:
@efritz The original issue was this one: https://github.com/sourcegraph/src-cli/issues/222

We're also deprecating the config file: https://github.com/sourcegraph/src-cli/issues/156

After looking at both of these it seemed to make sense to move towards only using environment variables.

cmd/src/main_test.go
ryanslade: I used `t.Cleanup` originally but `src-cli` is using an older version of Go in the build pipeline.

Let me check how easy it is to update that

ryanslade:
Nothing I've tried seems to get Appveyor to use Go 1.14 so I'm giving up for now and I'll create an issue

cmd/src/main_test.go
ryanslade: See comments and commits, I couldn't get Appveyor to use Go 1.14

PR #236, opened 2020-06-30
Visualize unsupported repos better

eseliger:
In order to make more transparent why some repos are excluded, we now visualize better which repos are excluded because they aren't supported, and also display the hint on how to use them in non-verbose mode. In addition, up to 10 repo names will appear in the UI to get a better sense which are the unsupported ones.

image

### DIFF
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -476,10 +476,7 @@ fragment repositoryFields on Repository {
 	for _, repo := range reposByID {
 		repos = append(repos, repo)
 	}
-	for _, r := range skipped {
-		logger.Infof("Skipping repository %s because we couldn't determine default branch.\n", r)
-	}
-	logger.RepoMatches(len(repos), unsupported)
+	logger.RepoMatches(len(repos), skipped, unsupported)
 
 	if content, err := result.Data.Search.Results.Alert.Render(); err != nil {
 		yellow.Fprint(os.Stderr, err)
--- a/cmd/src/actions_exec.go
+++ b/cmd/src/actions_exec.go
@@ -479,25 +479,7 @@ fragment repositoryFields on Repository {
 	for _, r := range skipped {
 		logger.Infof("Skipping repository %s because we couldn't determine default branch.\n", r)
 	}
-	for _, r := range unsupported {
-		logger.Infof("# Skipping repository %s because it's on a not supported code host.\n", r)
-	}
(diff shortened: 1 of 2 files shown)


internal/campaigns/logger.go
```diff
@@ -242,6 +242,38 @@ func (a *ActionLogger) DockerStepDone(repoName string, step int, elapsed time.Du
...
+		var warnStr string
+		if repoCount == 0 {
+			warnStr = "WARNING: "
+		}
+		matchesStr = fmt.Sprintf("%s%d repositories match the scopeQuery.", warnStr, repoCount)
+	}
+	if unsupportedCount > 0 {
+		matchesStr += fmt.Sprintf(" (Including %d on unsupported code hosts.)", unsupportedCount)
+	}
+	if unsupportedCount > 0 {
```
LawnGnome, line 260: These `if` statements can be combined.

internal/campaigns/logger.go
```diff
@@ -242,6 +242,38 @@ func (a *ActionLogger) DockerStepDone(repoName string, step int, elapsed time.Du
...
+		matchesStr += "\n\nThe following repositories were filtered out because they are on a codehost not supported by campaigns (Use -include-unsupported to generate patches for them anyways.):\n"
```
LawnGnome, line 261: Slight grammar nit: `anyways` should be singular in formal contexts. I also find the full sentence a little distracting in the parentheses, so I'd be inclined to make it a sentence fragment like the below:

```suggestion
		matchesStr += "\n\nThe following repositories were filtered out because they are on a codehost not supported by campaigns (use -include-unsupported to generate patches for them anyway):\n"
```

cmd/src/actions_exec.go
```diff
@@ -479,25 +479,7 @@ fragment repositoryFields on Repository {
 	for _, r := range skipped {
 		logger.Infof("Skipping repository %s because we couldn't determine default branch.\n", r)
 	}
```
mrnugget, line 481: I'd pull that into the `RepoMatches` method too.

internal/campaigns/logger.go
mrnugget, line 261: I'd get rid of the additional `(Including %d on unsupported code hosts.)` further up, which looks awkward with the parens, and just use this line with the count added:

```suggestion
		matchesStr += fmt.Sprintf("\n\n%d repositories were filtered out because they are on a codehost not supported by campaigns. Use -include-unsupported to generate patches for them anyway:\n", unsupportedCount)
```

Issue #238, opened 2020-07-01
Allow user to supply http headers for requests src-cli sends


gbrik:
Needed for customer whose sg instance is behind an auth proxy.

PR #239, opened 2020-07-01
Allow users to supply arbitrary HTTP headers for requests

efritz:
Closes https://github.com/sourcegraph/src-cli/issues/238.

This threads header values into the HTTP calls to the Sourcegraph instance via envvars such as `SRC_HEADER_FOO=BAR` (where `Foo: bar` would be the additional header value - note that header names are case **in**sensitive).

This is applied to:
- GraphQL requests
- LSIF uploads
- get-curl commands
- version command
- campagins actions

This is notably to fix a particular customer's issue, which has their instance behind an authenticating proxy that accepts special bearer tokens.


efritz:
cc @ryanslade as a sanity check to ensure that this isn't undoing any work in deprecating non-envvar settings.

cmd/src/headers.go
```diff
@@ -0,0 +1,30 @@
+package main
+
+import (
+	"os"
+	"strings"
+)
+
+// getAdditionalHeaders reads the environment for values like SRC_HEADER_NAME=VALUE
```
gbrik, line 8: typo:
getAdditionalHeaders -> parseAdditionalHeaders

README.md
```diff
@@ -73,6 +73,14 @@ Point `src` to your instance and access token using environment variables:
 SRC_ENDPOINT=https://sourcegraph.example.com SRC_ACCESS_TOKEN="secret" src search 'foobar'
 ```
 
+If your instance is behind an auth proxy that requires additional headers, these can be supplied via environment variables:
+
+```sh
+SRC_HEADER_NAME=value src search 'foobar'
+```
+
+In this example, the header name-value pair `Name: value` will be threaded to all HTTP requests to your instance. Multiple such headers can be supplied.
```
slimsag, line 82: This should not go in the README, it adds too much bloat. Instead please create a new `AUTH_PROXY.md` document and add just this to the README:

> Sourcegraph behind a custom auth proxy? See [auth proxy configuration](AUTH_PROXY.md) docs.

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Added
 
+- Add support for `SRC_HEADER_` environment variables.
```
slimsag, line 16:
```suggestion
- `SRC_HEADER_NAME=value` is now supported for authenticating `src` with custom auth proxies. See [auth proxy configuration docs](AUTH_PROXY.md) for more information.
```

AUTH_PROXY.md
```diff
@@ -0,0 +1,9 @@
+# Authenticating requests behind a proxy
+
+If your instance is behind an authenticating proxy that requires additional headers, they can be supplied via environment variables as follows:
+
+```sh
+SRC_HEADER_NAME=value src search 'foobar'
```
mrnugget, line 6: I suggest to use real values in examples and use multiple of them.

Why? `SRC_HEADER_NAME=value` can also be read as "the name of the header is set by using `SRC_HEADER_NAME` and the name itself is then the `value`"

If you use `SRC_HEADER_AUTHORIZATION="Bearer my-token"` and explains that this sets `authorization: Bearer my-token` then it's clear what's what.

mrnugget: Question: how would you set a header that contains a hyphen in the name, e.g. `Proxy-Authorization`? `SRC_HEADER_PROXY-AUTHORIZATION`? Or is the `-` turned into a `_`?

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Added
 
+- `SRC_HEADER_NAME=value` is now supported for authenticating `src` with custom auth proxies. See [auth proxy configuration docs](AUTH_PROXY.md) for more information.
```
mrnugget, line 16: Same here, I'd use a concrete header instead of `_NAME`. That makes more sense in the context of "auth proxy configuration"

cmd/src/headers.go
```diff
@@ -0,0 +1,30 @@
...
+	return parseAdditionalHeadersFromMap(os.Environ())
+}
+
+const additionalHeaderPrefix = "SRC_HEADER_"
+
+func parseAdditionalHeadersFromMap(environ []string) map[string]string {
+	additionalHeaders := map[string]string{}
+	for _, value := range environ {
+		parts := strings.SplitN(value, "=", 2)
+		if len(parts) != 2 || len(parts[0]) <= len(additionalHeaderPrefix) || len(parts[1]) == 0 || !strings.HasPrefix(parts[0], additionalHeaderPrefix) {
```
ryanslade, line 22: You don't need to do ```len(parts[0]) <= len(additionalHeaderPrefix)```, strings.HasPrefix does the same check

efritz: This checks that it’s a proper prefix and it’s not the exact string so that when we trim below there’s still something left.

efritz: Adding additional docs here.

AUTH_PROXY.md
efritz: They're copied verbatim, but you'd need to use env to supply them, e.g. `env "SRC_HEADER_FOO-BAR=whyamilikethis" src doathing`

PR #249, opened 2020-07-21
Add GitLab to the list of supported hosts

LawnGnome:


cmd/src/actions_exec.go
```diff
@@ -448,7 +448,7 @@ fragment repositoryFields on Repository {
 		}
 
 		// Skip repos from unsupported code hosts but don't report them explicitly.
-		if !includeUnsupported && strings.ToLower(repo.ExternalRepository.ServiceType) != "github" && strings.ToLower(repo.ExternalRepository.ServiceType) != "bitbucketserver" {
+		if !includeUnsupported && strings.ToLower(repo.ExternalRepository.ServiceType) != "github" && strings.ToLower(repo.ExternalRepository.ServiceType) != "bitbucketserver" && strings.ToLower(repo.ExternalRepository.ServiceType) != "gitlab" {
```
mrnugget, line 451: Should we use a version check here? We do have this:

https://github.com/sourcegraph/src-cli/blob/d343155a96c664de1c789b743a2c84e54bd4dbff/cmd/src/campaigns_create.go#L92-L100

LawnGnome: Yep, implemented.

PR #250, opened 2020-07-21
serve-git: port src-expose serve to hidden sub command

keegancsmith:
This initial commit copy pastes the serve code from the Sourcegraph repo
at commit d0d94c90. It's mostly copy paste, but we remove some
side-effects we no longer needed to rely on.

I'd like to continue developing this against master and when it is more
ready I'll document it in the CLI and other docs.

Part of https://github.com/sourcegraph/sourcegraph/issues/9243
Part of https://github.com/sourcegraph/sourcegraph/issues/12363


cmd/src/gitserve.go
```diff
@@ -0,0 +1,370 @@
...
+	"sync/atomic"
+	"text/template"
+
+	"github.com/pkg/errors"
+)
+
+func init() {
+	flagSet := flag.NewFlagSet("git-serve", flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), `'src git-serve' is a tool which serves your local git repositories over HTTP for Sourcegraph to pull.
```
mrnugget, line 26:
```suggestion
		fmt.Fprintf(flag.CommandLine.Output(), `'src git-serve' serves your local git repositories over HTTP for Sourcegraph to pull.
```

cmd/src/gitserve.go
```diff
@@ -0,0 +1,370 @@
...
+</html>`))
+
+type Repo struct {
+	Name string
+	URI  string
+}
+
+func (s *Serve) handler() (http.Handler, error) {
+	s.Info.Printf("serving git repositories from %s", s.Root)
+	s.configureRepos()
```
mrnugget, line 135: I'd pull these two out of `handler()` and put them in `Start()`, where the rest of the imperative code lives. That would make `handler()` a "pure" method that only returns the handler.

btw. the `s.configureRepos()` is also in `mux.HandleFunc("/v1/list-repos", func(w http.ResponseWriter, r *http.Request) { ... }` — the call here is only done for side-effects I assume?

keegancsmith: will do. Yes it is done for side-effect. It's a side-effect we can remove in another commit I am getting done as part of making src-expose better.

keegancsmith:
Thanks for the quick review.

> One thing to keep in mind: if it's in any way possible, I think it would be great to move the code to, say, internal/gitserve. We started moving code in there, because ./cmd/src becomes.. unwieldy.

Will do. Wasn't sure if I should do that given the code is kinda compact. But I agree :)

cmd/src/gitserve.go
keegancsmith: we don't actually need the side-effect anymore, so nice spotting.

PR #251, opened 2020-07-21
servegit: use git protocol v2

keegancsmith:
This commit copy pastes the upload pack service we use in gitserver from
the Sourcegraph repo at commit d0d94c90. It is mostly copy paste, but we
remove prometheus metrics and "--strict" since we allow non-bare repos.

Previously we supported only git protocol v0. This is regarded as the
"dumb" protocol. We ran into issues of it being slow to update (requires
running git update-server-info). It also requires write access to the
files.

Git protocol v2 allows more efficient fetching and extra cool features
we may start to rely on.

Part of https://github.com/sourcegraph/sourcegraph/issues/12363
Part of https://github.com/sourcegraph/sourcegraph/issues/11707

### DIFF
--- a/internal/servegit/service.go
+++ b/internal/servegit/service.go
@@ -1,0 +1,109 @@
+package servegit
+
+import (
+	"log"
+	"net/http"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+	"time"
+)
+
+var uploadPackArgs = []string{
+	// Partial clones/fetches
+	"-c", "uploadpack.allowFilter=true",
+
+	// Can fetch any object. Used in case of race between a resolve ref and a
+	// fetch of a commit. Safe to do, since this is only used internally.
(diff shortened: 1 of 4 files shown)


internal/servegit/service.go
```diff
@@ -0,0 +1,109 @@
...
+
+	var repo, svc string
+	for _, suffix := range []string{"/info/refs", "/git-upload-pack"} {
+		if strings.HasSuffix(r.URL.Path, suffix) {
+			svc = suffix
+			repo = strings.TrimSuffix(r.URL.Path, suffix)
+			repo = strings.TrimPrefix(repo, "/")
+			break
+		}
+	}
```
mrnugget, line 55: Do we need a sanity check that `repo != "" && svc != ""`?

keegancsmith: empty repo is allowed (the root is a git repo), so don't wanna guard against it. We check for svc == "" later.

PR #264, opened 2020-08-13
api: add Client.NewHTTPRequest

keegancsmith:
I want to make authenticated requests against the Sourcegraph API which
are not via GraphQL. In particular I am experimenting with Server Sent
Events endpoint. This commit adds NewHTTPRequest to the Client
interface, which allows access to an authenticated http.Request. This
will then be used to communicate with the non-GraphQL endpoint.

I have no idea if this fits your vision for the api package. After a little thought this seemed like the cleanest way to achieve this access, but I'm not a big fan of expanding interfaces.

Side-note: We could avoid adding to both interface and impl if this package returned a Client struct (but still kept the Request interface). The call sites could then use an interface for Client to make them testable.

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -1,4 +1,4 @@
-// Package api provides a basic client library for the Sourcegraph GraphQL API.
+// Package api provides a basic client library for the Sourcegraph API.
 package api
 
 import (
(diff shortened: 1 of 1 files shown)


keegancsmith:
> [Well, this is awkward.](https://github.com/sourcegraph/src-cli/pull/260/files#diff-1c93640aec056ac956d6bd96ee83a3c8)
> 
> I just hadn't had a chance to split my version of this out into a new PR yet! Let's land this, and I can then use it for #260. Thanks!

Hah. Cool yeah they looks like very similiar approaches. Only real difference is this method expects a path not a uri. Not sure which you prefer, but feel free to change in your PR :)

PR #268, opened 2020-08-20
output: add a basic output package

LawnGnome:
This gives us some basic primitives to output structured, human readable information from commands. This is currently intended only for the campaigns work taking place in #260, but also gives us options for additional functionality elsewhere, such as internal/api.

This is somewhat duplicative with things that we currently have in cmd/src and internal/campaigns; I intend to gradually clean those up as time permits.

This will be included in #260 as well, but I've split it out for easier review. Here's a demo:

out1


LawnGnome:
@mrnugget reviewed this in #260, so I'm going to close this and just handle it over there. (But this will remain so you know what the baby `internal/output` package looked like!)

PR #278, opened 2020-08-24
Update sourcegraph/codeintelutils version

efritz:
This stops retrying lsif upload requests on errors that aren't retryable (will now only retry on 500-level and network errors).


Strum355 (approved):
Would be cool to do tagging going forward :+1:

PR #282, opened 2020-08-26
Do not re-execute campaign spec steps when the diff is still valid

chrispine:
By changing the JSON marshalling of `Task` to ignore the `Template`, this makes it so that changes to the template (assuming no changes to `steps:` or `on:`) will serialize to the same JSON, so the previous run is found in the cache.

We then reuse the diff from the cache, and otherwise use the new spec.

**Testing:** I tested this manually with the hello world campaign. I _think_ everything worked correctly, and that my errors were unrelated to this feature, but I'd really appreciate someone else running a campaign and verifying that this works for them, too.

Closes [#13172](https://github.com/sourcegraph/sourcegraph/issues/13172)

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -145,7 +145,10 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 			err = errors.Wrapf(err, "checking cache for %q", task.Repository.Name)
 			return
 		} else if result != nil && len(result.Commits) == 1 {
-			// Build the changeset spec.
+			// Build a new changeset spec. We don't want to use `result` as is,
+			// because the changesetTemplate may have changed. In that case
+			// the diff would still be valid, so we take it from the cache,
+			// but we still build a new ChangesetSpec from the task.
 			diff := result.Commits[0].Diff
 			spec := createChangesetSpec(task, diff)
 
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -23,7 +23,7 @@ type Executor interface {
 type Task struct {
 	Repository *graphql.Repository
 	Steps      []Step
-	Template   *ChangesetTemplate
+	Template   *ChangesetTemplate `json:"-"`
 }
 
 func (t *Task) cacheKey() ExecutionCacheKey {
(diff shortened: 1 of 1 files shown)


internal/campaigns/executor.go
```diff
@@ -144,15 +144,19 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		if result, err = x.cache.Get(ctx, cacheKey); err != nil {
 			err = errors.Wrapf(err, "checking cache for %q", task.Repository.Name)
 			return
-		} else if result != nil {
+		} else if result != nil && len(result.Commits) == 1 {
+			// Build the changeset spec.
```
LawnGnome, line 148: Can you expand this out a little to note _why_ we're rebuilding the spec from the cached diff, please? (It took me a couple of reads to realise that it's because the changeset template may have changed, which might affect the "commit" that gets embedded in the changeset spec.)

internal/campaigns/executor.go
```diff
@@ -144,15 +144,19 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		if result, err = x.cache.Get(ctx, cacheKey); err != nil {
 			err = errors.Wrapf(err, "checking cache for %q", task.Repository.Name)
 			return
-		} else if result != nil {
+		} else if result != nil && len(result.Commits) == 1 {
```
mrnugget, line 147: Why this check?

I think it's here because you want to avoid that the `result.Commits[0].Diff` further down crashes, right? If that is the case, I'd move it there:

```go
var diff string
if len(result.Commits) >= 1 {
  diff = result.Commits[0].Diff
}

spec := createChangesetSpec(task, diff)
```

Otherwise, I fear, the "is something in the cache?"-check might be dependent on some side-effects (what if we _want_ to cache a result without any commits? etc.

mrnugget (approved):
Wow, this was ... less hairy than I thought it would be. Well done! 👏 

Agree with Adam that a comments would be cool

chrispine:
Credit for elegance of the solution goes entirely to Erik. 😄

internal/campaigns/executor.go
chrispine: I was actually more concerned about the case where there are more than one commit. Is it fine to ignore the diffs of all subsequent commits?

I know that there is only one commit at the moment, but for future-proofing, I'd prefer a solution that either gracefully degrades or instantly fails. By calling it a cache miss, I was opting for "gracefully degrades". (Maybe I'm worrying about something I don't need to be worrying about?)

chrispine: (To be clear, I'm happy to change this for the zero-commits case as you suggest. I'm just not sure what to do when there are 2+ commits.)

internal/campaigns/executor.go
chrispine: Hopefully my new comment is more clear. Thanks!

LawnGnome: It is. Thanks!

internal/campaigns/executor.go
mrnugget: > I was actually more concerned about the case where there are more than one commit. Is it fine to ignore the diffs of all subsequent commits?

Right now, yeah. If we have more than one commit, _right now_, we should either panic because something's wrong or we can ignore it.

In the future, I think what we'd need to do here is to create a commit for each diff that we produce in a repository.

mrnugget:
Left my thoughts here https://github.com/sourcegraph/src-cli/pull/282#discussion_r478898078 (and linking to them because it's easy to miss in GitHub's PR view).

I'm going to merge this now so it's in `main` and I can use it along with the other fixes 😄

internal/campaigns/executor.go
chrispine: Alright, I'll open another PR that accepts 0 or 1 commits, and panics on more than 1.

PR #283, opened 2020-08-26
campaigns: improve error output

LawnGnome:
This enforces a newline separation above and below the error block, and ensures that the preview command doesn't erroneously print a preview URL after an error.


eseliger (approved):
LGTM

PR #284, opened 2020-08-26
campaigns: show the repos that are unsupported

LawnGnome:
Previously, only the code host types were displayed when a campaign was previewed or applied with unsupported repos. This commit extends the error message to include the actual repos as well:

image

### DIFF
--- a/internal/campaigns/errors.go
+++ b/internal/campaigns/errors.go
@@ -1,0 +1,41 @@
+package campaigns
+
+import (
+	"fmt"
+	"strings"
+
+	"github.com/sourcegraph/src-cli/internal/campaigns/graphql"
+)
+
+// unsupportedRepoSet provides a set to manage repositories that are on
+// unsupported code hosts. This type implements error to allow it to be
+// returned directly as an error value if needed.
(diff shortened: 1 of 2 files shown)


internal/campaigns/errors.go
```diff
@@ -0,0 +1,41 @@
...
+
+	types := []string{}
+	for t := range typeSet {
+		types = append(types, t)
+	}
+
+	return fmt.Sprintf(
+		"found repositories on unsupported code hosts: %s\nrepositories:\n\t%s",
+		strings.Join(types, ", "),
+		strings.Join(repos, ", "),
```
eseliger, line 31: as this list could grow a bit larger, maybe `"\n- "` would be a better separator?

eseliger (approved):
nice, even easier to grasp

internal/campaigns/errors.go
LawnGnome: Blarg, that was `\n\t` before I started refactoring this out of an ad hoc structure in `Service`. Let me fix that real quick.

PR #286, opened 2020-08-27
campaigns: move campaign spec schema to ./schema

LawnGnome:
Originally, I intended the string constant in internal/campaigns to be a stopgap until a route was added to sourcegraph/sourcegraph to allow for downloading the schema, thereby allowing sourcegraph/sourcegraph to be the single source of truth for spec schemas.

However, after working more on prototyping schema changes, I've come to realise that I was mistaken, and that the tight coupling of the `CampaignSpec` type in internal/campaigns to the schema proper makes it — at best — extremely difficult to have the schema come from an external source.

Therefore, we need a more robust solution to handle the schema than copy/pasting the raw JSON into a string constant, getting a compile error because I forget there are backticks in the schema, and fixing those by hand. We already have the bits needed in schema to handle stringdata generation, so let's apply that here.


eseliger (approved):
Kill all dem TODOs 🎊

PR #287, opened 2020-08-27
Do not create ChangesetSpecs on server with empty diff

mrnugget:
I ran into this yesterday with Chris: we ran a command across multiple
repositories and in one repository it didn't produce a diff.

That's totally to be expected, since you don't know which command will
produce which diff in which repository.

The problem is that src-cli then sends up the changeset spec, even
though the diff is empty and we then try to apply the diff in gitserver,
which is where it fails (bonus: our error message looked really good!)

This commit here fixes it by not adding a spec with an empty diff to the
"list of completed specs".

I've looked at a few places where we could add this check (for example:
right before we make the GraphQL request we could do an early exit, or
before we add it to the cache, ...) and I'm still not 100% whether
that's the best place.

But I think the requirements are these:
- We should not send up empty diffs (because it looks weird in the UI if
  we say "we will create this changeset", even though it has no diff).
- We should still cache the result, even though the diff is empty.
- We should be able to report progress when executing it (we did do _something_, but we
  just didn't produce a diff)
- We shouldn't say "sending up 1 changeset spec" if we don't do that

That's why I added it where it is, which also required to fix the
progress bar for the case where "bar.Max == 0" so we don't run into a
panic.

I think this is the least invasive option of all the ones I considered
(I could also, for example, blast a few `if len(specs) > 0` around the
codebase...).

Let me know what you think, especially you @LawnGnome!

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -200,10 +200,8 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 	status.ChangesetSpec = spec
 	x.updateTaskStatus(task, status)
 
-	// Add the spec to the executor's list of completed specs.
-	x.specsMu.Lock()
-	x.specs = append(x.specs, spec)
-	x.specsMu.Unlock()
+	// Add it to the status
+	status.ChangesetSpec = spec
 
 	// Add to the cache. We don't use runCtx here because we want to write to
 	// the cache even if we've now reached the timeout.
@@ -211,6 +209,16 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		err = errors.Wrapf(err, "caching result for %q", task.Repository.Name)
 	}
 
+	// If the steps didn't result in any diff, we don't need to add it to the
+	// list of specs that are displayed to the user and send to the server.
+	if len(diff) == 0 {
+		return
+	}
+
+	// Add the spec to the executor's list of completed specs.
+	x.specsMu.Lock()
+	x.specs = append(x.specs, spec)
+	x.specsMu.Unlock()
 	return
 }
 


eseliger (approved):
Approving for the idea, but the actual code review, especially regarding the logging, I want to defer to @LawnGnome

internal/output/progress_tty.go
```diff
@@ -206,7 +206,13 @@ func (p *progressTTY) writeLine(bar *ProgressBar) {
...
 	// Unicode box drawing gives us eight possible bar widths, so we need to
 	// calculate both the bar width and then the final character, if any.
-	segments := int(math.Round((float64(8*barWidth) * value) / bar.Max))
+	var segments int
+	if bar.Max == 0 {
+		// If we didn't have any progress, we use the max width
+		segments = 8 * barWidth
+	} else {
+		segments = int(math.Round((float64(8*barWidth) * value) / bar.Max))
+	}
```
LawnGnome, line 215: This conflicts with #289. (Flagging as much for @chrispine as you!)

internal/campaigns/executor.go
```diff
@@ -147,8 +147,6 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		} else if result != nil {
 			status.Cached = true
 			status.ChangesetSpec = result
-			status.FinishedAt = time.Now()
-			x.updateTaskStatus(task, status)
```
LawnGnome, line 151: Was there a reason for removing these (and the `updateTaskStatus` call below)?

internal/campaigns/executor.go
```diff
@@ -206,20 +204,26 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
...
 	// the cache even if we've now reached the timeout.
 	if err = x.cache.Set(ctx, cacheKey, spec); err != nil {
 		err = errors.Wrapf(err, "caching result for %q", task.Repository.Name)
 	}
 
+	// If the steps didn't result in any diff, we don't need to add it to the
+	// list of specs that are displayed to the user and send to the server.
+	if len(diff) == 0 {
+		return
+	}
```
LawnGnome, line 221: :+1:

internal/output/progress_tty.go
chrispine: Yes, but I committed first. 😛 

Adam and I chatted and thought that if there are zero items, we should leave it blank instead of full width. (Well, Adam thought so. I'm happy either way.)

internal/campaigns/executor.go
```diff
@@ -206,20 +204,26 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
...
 	}
-	status.ChangesetSpec = spec
-	x.updateTaskStatus(task, status)
 
-	// Add the spec to the executor's list of completed specs.
-	x.specsMu.Lock()
-	x.specs = append(x.specs, spec)
-	x.specsMu.Unlock()
+	// Add it to the status
+	status.ChangesetSpec = spec
```
chrispine, line 209: This will likely conflict with #282.

internal/output/progress_tty.go
mrnugget: No way! I committed first!

But doesn't matter: happy you found a fix that you like, I wasn't sure what the desired behaviour is.

internal/campaigns/executor.go
mrnugget: They're all handled in the `defer` up and I thought it was confusing that at some points in the function we duplicate the `defer` and at others we don't.

mrnugget:
I'm going to rebase and make the most minimal change out of this (`if len(diff) == 0 { ... }`) and merge it, because @LawnGnome gave me a 👍 on that.

PR #290, opened 2020-08-27
do not crash if termbox fails to init

chrispine:
I'd like to debug in VS Code, but termbox fails to init in that environment for some reason. Since all we want is the size, how about if we fall back to a reasonable default (80x25) in that case?


eseliger (approved):
I wonder what the cause is, but seems like a fair fix

mrnugget:
How did you find out it's termbox? Yesterday, when running into the `bar.Max == 0` bug, I thought to myself: "hey, instead of adding `fmt.Println` number 18, why not try to use a debugger for once?" But then when I booted up `dlv` nothing happened :(

chrispine:
The call stack made it pretty clear that it was failing on init.

PR #291, opened 2020-08-28
Add -n as an alias for -namespace

mrnugget:
-namespace is quite long and I've wished for a shorter version multiple
times now.

Apparently, though, introducing an alias can be controversial if all you
have is single-dash flags like `-namespace` (vs. `--namespace`), since
`-namespace` could then be interpreted as `-n -a -m -e ...` etc.

Not sure whether that's an _actual_ concern though, since we do have
single-letter flags already (`-v` and `-j`, for example).

### DIFF
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -68,6 +68,8 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir string) *campaignsAp
 		&caf.namespace, "namespace", "",
 		"The user of organization namespace to place the campaign within.",
 	)
+	flagSet.StringVar(&caf.namespace, "n", "", "Alias for -namespace.")
+
 	flagSet.IntVar(
 		&caf.parallelism, "j", 0,
 		"The maximum number of parallel jobs. (Default: GOMAXPROCS.)",


cmd/src/campaigns_common.go
```diff
@@ -68,6 +68,8 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir string) *campaignsAp
 		&caf.namespace, "namespace", "",
 		"The user of organization namespace to place the campaign within.",
 	)
+	flagSet.StringVar(&caf.namespace, "n", "", "Alias for -namespace.")
```
eseliger, line 71: does that mean `-n` takes precendence over `-namespace`? Just curious, not a blocker

eseliger (approved):

cmd/src/campaigns_common.go
mrnugget: I don't think so (or I'm misunderstanding what you mean). They're independent flags and if you give them different pointers, they'll work independently.

eseliger: I mean we're copying into `&caf.namespace` for both flags 

So if someone would provide `-n eseliger -namespace=mrnugget`, would namespace be `eseliger` or `mrnugget` is what I meant. I suppose it is `eseliger`?

mrnugget: Last one wins:

screenshot_2020-08-28_14 38 07@2x

eseliger: Interesting 👍 thanks

PR #292, opened 2020-08-28
Show progress bar when downloading Docker images

mrnugget:
Not sure whether that fits your vision for how you want to show progress, @LawnGnome, but since the image pulling should be done once for all steps and not for each task/repository, I think it makes sense to pull it out.

Also tried to follow your convention of not mixing `output` into the `Service` but providing a progress callback.

### Demo GIF

This uses some large images, so don't be surprised that it takes so long.

screenshot_2020-08-28_13 45 10


eseliger (approved):
I would be awesome if there was a way to get some more feedback on how fast it goes, but I think this will do for now.
(ex when on flaky network, is this a large image or is this a bad connection?)

PR #294, opened 2020-08-30
campaigns: allow partial publishing of changesets

LawnGnome:
This is a prototype of the option being presented in [RFC 228](https://docs.google.com/document/d/1A-5cbYGz1p1UB1eAFsIgpK5XDkvS7ZNAVKMdtBm_WY0/edit?usp=sharing). It needs to be used in conjunction with sourcegraph/sourcegraph#13476.

Note that this PR adds a dependency on https://github.com/LawnGnome/campaign-schema for schema-related code shared between this PR and sourcegraph/sourcegraph#13476; my intention is to move this to https://github.com/sourcegraph/campaignutils if accepted.

An example of the type of campaign you could run with this is:

```yaml
name: hello-world
description: Add Hello World to READMEs
​
on:
  - repositoriesMatchingQuery: file:README.md r:github.com/*
  - repositoriesMatchingQuery: file:README.md r:gitlab.sgdev.org
​
# In each repository, run this command. Each repository's resulting diff is captured.
steps:
  - run: echo Hello World | tee -a $(find -name README.md)
    container: alpine:3
​
changesetTemplate:
  title:
    default: Hello World
    except:
      - github.com/sourcegraph/*: Hello Sourcegraph!
  body: My first campaign!
  commit:
  published:
    - github.com/LawnGnome/*: true
```

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -294,6 +294,8 @@ func reachedTimeout(cmdCtx context.Context, err error) bool {
 }
 
 func createChangesetSpec(task *Task, diff string) *ChangesetSpec {
+	repo := task.Repository.Name
+
 	var authorName string
 	var authorEmail string
 
@@ -323,7 +325,7 @@ func createChangesetSpec(task *Task, diff string) *ChangesetSpec {
 					Diff:        string(diff),
 				},
 			},
-			Published: task.Template.Published,
+			Published: task.Template.Published.Value(repo),
 		},
 	}
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Added
 
+- The `published` flag in campaign specs may now be an array, which allows only specific changesets within a campaign to be published based on the repository name. [#294](https://github.com/sourcegraph/src-cli/pull/294)
+
 ### Changed
 
 - Error reporting by `src campaign [preview|apply]` has been improved and now includes more information about which step failed in which repository. [#325](https://github.com/sourcegraph/src-cli/pull/325)
--- a/go.mod
+++ b/go.mod
@@ -16,7 +16,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.4 // indirect
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/pkg/errors v0.9.1
-	github.com/sourcegraph/campaignutils v0.0.0-20200930010549-508f59a6ab8f
+	github.com/sourcegraph/campaignutils v0.0.0-20200930165749-a8777d33a817
 	github.com/sourcegraph/codeintelutils v0.0.0-20200824140252-1db3aed5cf58
 	github.com/sourcegraph/go-diff v0.6.0
 	github.com/sourcegraph/jsonx v0.0.0-20200629203448-1a936bd500cf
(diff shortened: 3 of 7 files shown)


mrnugget:
NICE! I like this the most so far.

Two thoughts:
1. The `only:` in the `title:` acts more like an `overwrites:` oder `except:` instead of an `only:`: "default value is 'Hello World', _except_ when we match this, then it's "Hello Sourcegraph".
2. The match in `published.except:` is not prefixed with `match:`. I wonder whether we could make that more uniform.

I feel like we're now 95% there!

LawnGnome:
This is roughly updated in line with the RFC and its discussion.

One thing we should talk about further is that I've separated out the schema and `OverridableString` into [a new repo](https://github.com/LawnGnome/campaign-schema), which I've temporarily namespaced under my own user until we decide what we want to do. The reason for this is because we'll need a lot of this in sourcegraph/sourcegraph as well, so it makes sense to me to migrate the schema and common types into that repo. I'd still need to move `OverridableBool` into that repo, but it's late, and I think this is enough to get a general sense of where this would be going.

So, concerns?

mrnugget:
Few thoughts on the schema and `OverridableString`:

1. Do we really need the whole functionality in `sourcegraph/sourcegraph`? I know that we need to share the raw JSON schema so we can do validation in the API layer, but while we currently [do unmarshal the CampaignSpec](https://github.com/sourcegraph/sourcegraph/blob/main/internal/campaigns/types.go#L1714-L1741) into a Go struct, we don't really _need to_, right? We're currently only using the top-level fields, such as `name`, `description`. Considering this, I think only in `src-cli` would we need to expand and evaluate the globs in the campaign spec. (Of course, with server-side execution we would need all of that in `sourcegraph/sourcegraph`, so maybe you meant that)
2. code intel related code in `src-cli` imports `github.com/sourcegraph/codeintelutils` — that would be our equivalent here, right?

LawnGnome:
> * Do we really need the whole functionality in `sourcegraph/sourcegraph`? I know that we need to share the raw JSON schema so we can do validation in the API layer, but while we currently [do unmarshal the CampaignSpec](https://github.com/sourcegraph/sourcegraph/blob/main/internal/campaigns/types.go#L1714-L1741) into a Go struct, we don't really _need to_, right? We're currently only using the top-level fields, such as `name`, `description`. Considering this, I think only in `src-cli` would we need to expand and evaluate the globs in the campaign spec. (Of course, with server-side execution we would need all of that in `sourcegraph/sourcegraph`, so maybe you meant that)

We don't, but we will, as you said. I'm happy enough to cut this back and embed the logic in `src-cli` for now, but with server side execution the shared bits will need to be shared.

> * code intel related code in `src-cli` imports `github.com/sourcegraph/codeintelutils` — that would be our equivalent here, right?

Exactly.

internal/campaigns/executor.go
```diff
@@ -312,18 +313,18 @@ func createChangesetSpec(task *Task, diff string) *ChangesetSpec {
...
 				{
-					Message:     task.Template.Commit.Message,
+					Message:     task.Template.Commit.Message.Value(repo),
 					AuthorName:  authorName,
 					AuthorEmail: authorEmail,
 					Diff:        string(diff),
 				},
 			},
-			Published: task.Template.Published,
+			Published: task.Template.Published.Is(repo),
```
mrnugget, line 327: Tiniest nitpicky thought: after reading `Value` and then `Is` I'm wondering whether a generic method name that says "give me the value for this repo" makes sense to have for every field that's overridable, like `ValueFor(repo)` for example.

mrnugget (approved):

eseliger (approved):
Very cool, so simple and straightforward!

LawnGnome:
Flipped to review-ready. I'm going to dismiss the existing approvals, but I don't think there's anything controversial here.

The original branch that supported strings is pushed to [the `aharvey/override-strings` branch](https://github.com/sourcegraph/src-cli/tree/aharvey/override-strings).

mrnugget (approved):

eseliger (approved):

PR #297, opened 2020-09-01
campaigns: allow user to set git commit author

chrispine:
This is the src-cli side of the fix for [#12769](https://github.com/sourcegraph/sourcegraph/issues/12769). ([back-end PR](https://github.com/sourcegraph/sourcegraph/pull/13574))

It implements the first part of [RFC 229](https://docs.google.com/document/d/1kDkIFQK9hiezEJrG5TfoGvsBKFKF6B6eUcKejYoX2XI/edit#), so campaign specs can have changeset templates that now look like:

```
changesetTemplate:
  title: Hello World
  body: My first campaign
  commit:
    author:
      name: Chris Pine
      email: chrispine@sourcegraph.com
  published: true
```

Issues

- I don't know how to say "this version of src-cli goes with that version of the back-end".

Future work

- Consider: instead of just going with the default author, check for local git config and use that. (But does that behave in ways we don't like: if I create a campaign without specifying an author, then you modify the campaign spec on your machine and apply it... would it change the commit author? Would we want it to?)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ## Unreleased changes
 
+- Campaigns specs now include an optional `author` property. (If not included, `src campaigns` generates default values for author name and email.) `src campaigns` now includes the name and email in all changeset specs that it generates.
+
 ### Changed
 
 - The default branch for the `src-cli` project has been changed to `main`. [#262](https://github.com/sourcegraph/src-cli/pull/262)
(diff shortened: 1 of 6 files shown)


mrnugget (approved):

eseliger (approved):

LawnGnome (approved):
LGTM!

I think the only other thing we would need is a changelog entry noting the extra fields. Beyond that, :+1:.

> I don't know how to say "this version of src-cli goes with that version of the back-end".

We discussed this in our sync today, and the answer is [`srccli.MinimumVersion`](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/src-cli/consts.go#L9:25).

PR #301, opened 2020-09-03
Replace travis by Github Actions as Linux / goreleaser CI

eseliger:
Replaces the travis workflow by GitHub actions, as we've had a lot of flakiness and broken builds with travis in the past. The `go-test` step is also faster than on travis.

Need to merge this in order to test the goreleaser config, hope it works first-try 🤞


efritz (approved):
who me

PR #312, opened 2020-09-09
Skip repos without default branch

eseliger:
These are usually empty repos with 0 branches.

Not sure if we would want to log those?

Closes #293


LawnGnome (approved):
This looks good.

In terms of logging, I'm on the fence. We could extend `ResolveRepositories()` to return a third array of skipped repositories, but we probably wouldn't surface that except when the user is calling `src campaign validate`, which might be too hidden for the normal case.

Let's leave it like this for now and see if it's confusing in practice.

PR #313, opened 2020-09-09
Fix empty changesets being uploaded

### LINKED ISSUE #307: Does src-cli correctly skip repos that didn't yield a change?
https://k8s.sgdev.org/users/erik/campaigns/go-prefer-strconv has empty changesets

eseliger:
Before, the changesets would still be uploaded, even if the diff was empty. This fixes it.

@LawnGnome is this correct usage of the progress API, or am I working around something here?

Closes #307

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -203,25 +203,23 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


LawnGnome (approved):
LGTM. The earlier `defer` in `do()` will take care of setting `status.FinishedAt`, and that's what we need for the progress bar to work properly.

Issue #310, opened 2020-09-09
Add visual step for "resolving repositories"


eseliger:
Resolving the repositories can take several seconds, so we should add a visual step for that as well, otherwise it'll just dangle around for seconds indicating it's doing nothing

PR #314, opened 2020-09-11
Add visual step for resolving repositories

eseliger:
Also fixes unsupported repositories aborting the workflow. We now simply skip those and let the user know.

Since this step can take some time, I was always confused what src-cli does. It resolves repositories apparently.

Closes https://github.com/sourcegraph/src-cli/issues/310

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to `src-cli` are documented in this file.
 
 - Repositories without a default branch are skipped when applying/previewing a campaign spec. [#312](https://github.com/sourcegraph/src-cli/pull/312)
 - Log files produced when applying/previewing a campaign spec now have the `.log` file extension for easier opening. [#315](https://github.com/sourcegraph/src-cli/pull/315)
-- If the `repositoriesMatchingQuery:` query in a campaign spec yields repositories on unsupported code hosts the process is no longer aborted, but the repositories simply skipped. [#314](https://github.com/sourcegraph/src-cli/pull/314)
+- Campaign specs that apply to unsupported repositories will no longer generate an error. Instead, those repositories will be skipped by default and the campaign will be applied to the supported repositories only. [#314](https://github.com/sourcegraph/src-cli/pull/314)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ All notable changes to `src-cli` are documented in this file.
 
 - Repositories without a default branch are skipped when applying/previewing a campaign spec. [#312](https://github.com/sourcegraph/src-cli/pull/312)
 - Log files produced when applying/previewing a campaign spec now have the `.log` file extension for easier opening. [#315](https://github.com/sourcegraph/src-cli/pull/315)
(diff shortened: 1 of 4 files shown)


mrnugget:
I added a changelog entry and changed the code to use the helper functions we already had. The code looks good to me. If @LawnGnome and @chrispine don't have any objections, I'd merge it.

mrnugget (approved):

CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ All notable changes to `src-cli` are documented in this file.
 
 - Repositories without a default branch are skipped when applying/previewing a campaign spec. [#312](https://github.com/sourcegraph/src-cli/pull/312)
 - Log files produced when applying/previewing a campaign spec now have the `.log` file extension for easier opening. [#315](https://github.com/sourcegraph/src-cli/pull/315)
+- If the `repositoriesMatchingQuery:` query in a campaign spec yields repositories on unsupported code hosts the process is no longer aborted, but the repositories simply skipped. [#314](https://github.com/sourcegraph/src-cli/pull/314)
```
LawnGnome, line 22: This is actually more general: unsupported repositories can also be specified directly.

How about:

```suggestion
- Campaign specs that apply to unsupported repositories will no longer generate an error. Instead, those repositories will be skipped by default and the campaign will be applied to the supported repositories only. [#314](https://github.com/sourcegraph/src-cli/pull/314)
```

cmd/src/campaigns_common.go
```diff
@@ -176,8 +176,19 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 	}
 	imageProgress.Complete()
 
+	pending = campaignsCreatePending(out, "Resolving repositories")
+	repos, err := svc.ResolveRepositories(ctx, campaignSpec)
+	if err != nil {
+		if _, ok := err.(campaigns.UnsupportedRepoSet); ok {
+			campaignsCompletePending(pending, "Resolved repositories. Found unsupported codehosts.")
```
LawnGnome, line 183: Should we also display the list of unsupported repositories? This feels exceptional enough that users will want as much detail as possible, generally speaking.

internal/campaigns/service.go
```diff
@@ -164,12 +164,7 @@ func (svc *Service) SetDockerImages(ctx context.Context, spec *CampaignSpec, pro
 	return nil
 }
 
-func (svc *Service) ExecuteCampaignSpec(ctx context.Context, x Executor, spec *CampaignSpec, progress func([]*TaskStatus)) ([]*ChangesetSpec, error) {
-	repos, err := svc.ResolveRepositories(ctx, spec)
-	if err != nil {
-		return nil, errors.Wrap(err, "resolving repositories")
-	}
-
+func (svc *Service) ExecuteCampaignSpec(ctx context.Context, repos []*graphql.Repository, x Executor, spec *CampaignSpec, progress func([]*TaskStatus)) ([]*ChangesetSpec, error) {
```
LawnGnome, line 167: This is what I should have done in the first place. My bad.

LawnGnome (approved):

cmd/src/campaigns_common.go
mrnugget: Done! Happy to update the formatting of this in the future, but before I dive into the `output` lib right before branch cut, I think I'll go with the minimal but readable/copy-pastable version.

screenshot_2020-09-15_13 44 17

PR #315, opened 2020-09-11
Properly have .log file extension for temp log files

### LINKED ISSUE #308: Add .log suffix to logfiles
My terminal always asks me what program to use to open a .<timestamp> file, adding a .log file extension should help circumvent.

eseliger:
Closes #308 

before, the file ending wouldn't be `.log`, causing some terminals to not know how to open that file when clicking on it (hence, always have to select my fav editor, because the ending will always be different and cannot be remembered.)

Before and after:

image


chrispine (approved):
LGTM

PR #317, opened 2020-09-14
src login command to help users configure authentication w/access token

sqs:
You can run `src login https://sourcegraph.example.com` to verify that your `src` is authenticated to access that Sourcegraph instance. If not, it'll show you friendly steps to help you get it set up.

This will simplify all of our docs and in-product text that relies on users using the CLI. The steps to get them to the right state will be just:

1. Install the Sourcegraph CLI (`src`) at https://github.com/sourcegraph/src-cli [or link to releases]
1. Run `src login SOURCEGRAPH_URL` [and we can even fill in `SOURCEGRAPH_URL` in in-product text]

Here is a screenshot of what it looks like in various states:

image
image

README section:

image

### DIFF
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -75,6 +75,8 @@ type config struct {
 	Endpoint          string            `json:"endpoint"`
 	AccessToken       string            `json:"accessToken"`
 	AdditionalHeaders map[string]string `json:"additionalHeaders"`
+
+	ConfigFilePath string
 }
 
 // apiClient returns an api.Client built from the configuration.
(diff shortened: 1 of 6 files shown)


cmd/src/main.go
```diff
@@ -75,8 +75,21 @@ type config struct {
...
 	AdditionalHeaders map[string]string `json:"additionalHeaders"`
+
+	EndpointSource    configSource
+	AccessTokenSource configSource
+	ConfigFilePath    string
 }
 
+type configSource int
+
+const (
```
sqs, line 86: This stuff ended up being unnecessary, will revert.

cmd/src/main.go
```diff
@@ -75,8 +75,21 @@ type config struct {
 	Endpoint          string            `json:"endpoint"`
 	AccessToken       string            `json:"accessToken"`
 	AdditionalHeaders map[string]string `json:"additionalHeaders"`
+
+	EndpointSource    configSource
```
sqs, line 79: This stuff ended up being unnecessary, will revert.

cmd/src/main.go
```diff
@@ -108,10 +121,17 @@ func readConfig() (*config, error) {
 	}
 	var cfg config
 	if err == nil {
+		cfg.ConfigFilePath = cfgPath
```
sqs, line 124: This stuff ended up being unnecessary, will revert.

sqs:
@slimsag mentioned ([on Slack](https://sourcegraph.slack.com/archives/CMMTWQQ49/p1600058737132500?thread_ts=1599987169.130000&cid=CMMTWQQ49)):

> Nice, one idea would be to have it automatically populate your .profile or something

My reply:

> I thought about having it automatically update your .profile, but that is tricky to get right, esp. when many people would be using this with multiple instances (could be good in the future though).
Or it could tell you things like "Or you can copy/paste this into your .profile: `export SRC_ENDPOINT=... ...`", but that adds a lot of noise and most users right now would not need that hand-holding.

LawnGnome:
We could have this populate the config file instead, except I'd suggest that we should simultaneously migrate that file from [being `~/src-config.json`](https://sourcegraph.com/github.com/sourcegraph/src-cli/-/blob/cmd/src/main.go#L101:39) to something less intrusive and more in line with platform norms (`path.Join(os.UserConfigDir(), "sourcegraph", "src-cli.json")`, or something similar?).

sqs:
@LawnGnome We've deprecated the config file and are supporting only env var config. See https://github.com/sourcegraph/src-cli/issues/156#issuecomment-596326294 and https://github.com/sourcegraph/src-cli/pull/205.

LawnGnome:
> @LawnGnome We've deprecated the config file and are supporting only env var config. See [#156 (comment)](https://github.com/sourcegraph/src-cli/issues/156#issuecomment-596326294) and #205.

Fair enough. Thanks for the context!

mrnugget (approved):
I like this!

Issue #316, opened 2020-09-11
campaigns: make log and cache directories configurable


LawnGnome:
Right now, both the log and cache directories are hardcoded, but there's no real reason they should be. They should be configurable, ideally through a combination of environment variables and command line flags.

mrnugget:
Can't you set the cache directory with the `-cache` flag?

mrnugget:
Closing this because #318 allows setting the directory used for logs with `-tmp` and `-cache` already exists.

PR #318, opened 2020-09-14
Allow setting temp dir used by campaigns through flag/env var

mrnugget:
This addresses a customer report in which `src campaigns apply` couldn't
be run in CI because the docker-in-docker setup prohibits mounting
`/tmp` into containers.

It also fixes parts of
https://github.com/sourcegraph/src-cli/issues/316, since the temporary
directory is also where the logs are stored, but we might want to split
those locations up (in case a user wants to keep logs but not inside the
temporary directory).

### DIFF
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -112,11 +112,15 @@ func campaignsDefaultTempDirPrefix() string {
 	if p != "" {
 		return p
 	}
-	// We use an explicit prefix for our temp directories, because otherwise Go
-	// would use $TMPDIR, which is set to `/var/folders` per default on macOS. But
-	// Docker for Mac doesn't have `/var/folders` in its default set of shared
-	// folders, but it does have `/tmp` in there.
-	return "/tmp"
+	// On macOS, we use an explicit prefix for our temp directories, because
+	// otherwise Go would use $TMPDIR, which is set to `/var/folders` per
+	// default on macOS. But Docker for Mac doesn't have `/var/folders` in its
+	// default set of shared folders, but it does have `/tmp` in there.
+	if runtime.GOOS == "darwin" {
+		return "/tmp"
+
+	}
+	return os.TempDir()
 }
 
 func campaignsOpenFileFlag(flag *string) (io.ReadCloser, error) {
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -59,7 +59,7 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 	)
 	flagSet.StringVar(
 		&caf.tempDir, "tmp", tempDir,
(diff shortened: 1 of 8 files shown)


eseliger (approved):
Cool :)

cmd/src/campaigns_common.go
```diff
@@ -56,6 +57,10 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir string) *campaignsAp
 		&caf.clearCache, "clear-cache", false,
 		"If true, clears the cache and executes all steps anew.",
 	)
+	flagSet.StringVar(
+		&caf.tempDir, "tmp", tempDir,
+		"Directory for storing temporary data, such as repository archives when executing campaign specs or log files. Default is /tmp. Can also be set with environment variable SRC_CAMPAIGNS_TMP_DIR. .",
```
eseliger, line 62:
```suggestion
		"Directory for storing temporary data, such as repository archives when executing campaign specs or log files. Default is /tmp. Can also be set with environment variable SRC_CAMPAIGNS_TMP_DIR.",
```

cmd/src/campaigns_common.go
```diff
@@ -99,6 +104,21 @@ func campaignsDefaultCacheDir() string {
...
+func campaignsDefaultTempDirPrefix() string {
+	p := os.Getenv("SRC_CAMPAIGNS_TMP_DIR")
+	if p != "" {
+		return p
+	}
+	// We use an explicit prefix for our temp directories, because otherwise Go
+	// would use $TMPDIR, which is set to `/var/folders` per default on macOS. But
+	// Docker for Mac doesn't have `/var/folders` in its default set of shared
+	// folders, but it does have `/tmp` in there.
+	return "/tmp"
```
LawnGnome, line 119: I know this is replicating the existing behaviour, but I wonder if we should actually use [`os.TempDir()`](https://golang.org/pkg/os/#TempDir) on non-macOS systems. Probably something we can do as a follow up, though.

cmd/src/campaigns_common.go
```diff
@@ -56,6 +57,10 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir string) *campaignsAp
 		&caf.clearCache, "clear-cache", false,
 		"If true, clears the cache and executes all steps anew.",
 	)
+	flagSet.StringVar(
+		&caf.tempDir, "tmp", tempDir,
+		"Directory for storing temporary data, such as repository archives when executing campaign specs or log files. Default is /tmp. Can also be set with environment variable SRC_CAMPAIGNS_TMP_DIR.",
```
LawnGnome, line 62: The precedence behaviour is implicit right now in the code, but let's document it explicitly.

```suggestion
		"Directory for storing temporary data, such as repository archives when executing campaign specs or log files. Default is /tmp. Can also be set with environment variable SRC_CAMPAIGNS_TMP_DIR; if both are set, this flag will be used and not the environment variable.",
```

LawnGnome (approved):
LGTM.

cmd/src/campaigns_common.go
mrnugget: Done! Added the explicit "if macOS use /tmp" check and fallback to `os.TempDir()`.

PR #320, opened 2020-09-18
git author name & email from git config

chrispine:
Adds a function which uses `git config` to find default git author name and email.

This function is _untested and unused_. I was going to use this as a fallback for when the campaign spec does not include author details. However, we decided we didn't want to do that.

When we implement `src campaign create-spec` or whatever we call it, this would be a useful function to have.

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -1,6 +1,7 @@
 package campaigns
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


mrnugget (commented):
I'd rather not merge unused code. But we can merge this once we make use of it (feel free to create that create-spec command!)

FWIW: our linter in `sourcegraph/sourcegraph` would mark the "naked returns" (named return values + `return` without values) as problems.

And, super FWIW, here's how I would write it:

```go
func gitAuthorDetails() (string, string, bool, error) {
	gitConfig := func(property string) (string, bool, error) {
		cmd := exec.Command("git", "config", property)

		var out bytes.Buffer
		cmd.Stdout = &out

		if err := cmd.Run(); err != nil {
			return "", false, err
		}

		return out.String(), out.Len() > 0, nil
	}

	name, ok, err := gitConfig("user.name")
	if err != nil {
		return "", "", ok, err
	}
	if !ok {
		return "", "", false, nil
	}

	email, ok, err := gitConfig("user.email")
	if err != nil {
		return "", "", ok, err
	}
	if !ok {
		return "", "", false, nil
	}

	return name, email, true, nil
}
```

The boolean can be handy if you want to distinguish between "ran command but no value" and "failed to run command".

But: nitpicking here.

PR #321, opened 2020-09-21
Retain repository archives across executions

mrnugget:
Copying from the CHANGELOG:

- The default behaviour of `src campaigns [preview|apply]` has been changed to retain downloaded archives of repositories for better performance across re-runs of the command.
- To use the old behaviour and delete the archives use the `-clean-archives` flag.
- Repository archives are also not stored in the directory for temp data (see `-tmp` flag) anymore but in the cache directory, which can be configured with the `-cache` flag. To manually delete archives between runs, delete the `*.zip` files in the `-cache` directory (see `src campaigns -help` for its default location)

I'm still not sure whether this is the best way to achieve the goal of
"do not delete the same zip file over and over again", because it's easy
ot have the cache fill up: if you run campaigns multiple times and the
base ref of the repository changes you get a new zip every time.

While I keep thinking about this and hacking on it: opinions?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Changed
 
 - Error reporting by `src campaign [preview|apply]` has been improved and now includes more information about which step failed in which repository. [#325](https://github.com/sourcegraph/src-cli/pull/325)
+- The default behaviour of `src campaigns [preview|apply]` has been changed to retain downloaded archives of repositories for better performance across re-runs of the command. To use the old behaviour and delete the archives use the `-clean-archives` flag. Repository archives are also not stored in the directory for temp data (see `-tmp` flag) anymore but in the cache directory, which can be configured with the `-cache` flag. To manually delete archives between runs, delete the `*.zip` files in the `-cache` directory (see `src campaigns -help` for its default location).
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Changed
 
+- The default behaviour of `src campaigns [preview|apply]` has been changed to retain downloaded archives of repositories for better performance across re-runs of the command. To use the old behaviour and delete the archives use the `-clean-archives` flag. Repository archives are also not stored in the directory for temp data (see `-tmp` flag) anymore but in the cache directory, which can be configured with the `-cache` flag. To manually delete archives between runs, delete the `*.zip` files in the `-cache` directory (see `src campaigns -help` for its default location).
```
eseliger, line 18:
```suggestion
- The default behavior of `src campaigns [preview|apply]` has been changed to retain downloaded archives of repositories for better performance across re-runs of the command. To use the old behavior and delete the archives use the `-clean-archives` flag. Repository archives are also not stored in the directory for temp data (see `-tmp` flag) anymore but in the cache directory, which can be configured with the `-cache` flag. To manually delete archives between runs, delete the `*.zip` files in the `-cache` directory (see `src campaigns -help` for its default location).
```

eseliger (approved):
Cool

mrnugget:
@LawnGnome @chrispine I'd still love to get your input on this. Erik's comment makes me think that this is not the super bad idea I suspected it _could_ be, but I just want confirmation :)

eseliger:
I think tmp files are cleared on restart, then I don't feel like it's a terrible problem, but we might revisit once we run server side and encounter exhausted ephemeral storage.

chrispine (approved):
LGTM 👍

PR #324, opened 2020-09-21
Show total number at end of repositories command output

mrnugget:
Previously it was really hard to tell how many repositories matched without having to scroll through the terminal output.

Before:

screenshot_2020-09-21_13 27 30@2x

After:
screenshot_2020-09-21_13 27 47@2x


eseliger (approved):
cool, I didn't know we had templates for that 🤯

PR #325, opened 2020-09-22
Improve error output when executing step fails

mrnugget:
This is the first of possibly many stabs at improving the error
reporting.

### Before

before

### After

after

### DIFF
--- a/internal/campaigns/run_steps.go
+++ b/internal/campaigns/run_steps.go
@@ -137,7 +137,7 @@ func runSteps(ctx context.Context, client api.Client, repo *graphql.Repository, 
 		if err != nil {
 			logger.Logf("[Step %d] took %s; error running Docker container: %+v", i+1, elapsed, err)
 
-			return nil, StepFailedErr{
+			return nil, stepFailedErr{
 				Err:         err,
 				Args:        cmd.Args,
 				Run:         step.Run,
@@ -309,7 +309,7 @@ func probeImageForShell(ctx context.Context, image string) (shell, tempfile stri
 	return
 }
 
-type StepFailedErr struct {
+type stepFailedErr struct {
 	Run       string
 	Container string
 
@@ -322,7 +322,7 @@ type StepFailedErr struct {
 	Err error
 }
 
-func (e StepFailedErr) Error() string {
+func (e stepFailedErr) Error() string {
 	var out strings.Builder
 
 	fmtRun := func(run string) string {
--- a/internal/campaigns/run_steps.go
+++ b/internal/campaigns/run_steps.go
@@ -70,7 +70,7 @@ func runSteps(ctx context.Context, client api.Client, repo *graphql.Repository, 
 	}
 
 	for i, step := range steps {
-		logger.Logf("[Step %d] docker run %s", i+1, step.image)
+		logger.Logf("[Step %d] docker run %s %q", i+1, step.Container, step.Run)
 
 		cidFile, err := ioutil.TempFile(tempDir, prefix+"-container-id")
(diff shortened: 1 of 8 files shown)


internal/campaigns/run_steps.go
```diff
@@ -297,3 +308,57 @@ func probeImageForShell(ctx context.Context, image string) (shell, tempfile stri
...
+		printOutput(e.Stdout)
+	}
+
+	if len(e.Stderr) > 0 {
+		out.WriteString("\nstandard error:\n")
+		printOutput(e.Stderr)
+	}
+
+	if exitErr, ok := e.Err.(*exec.ExitError); ok {
+		out.WriteString(fmt.Sprintf("\nCommand failed with exit code %d.", exitErr.ExitCode()))
```
eseliger, line 358: Seems like we always set the exit code to 1? If so, I think it would be confusing

internal/campaigns/run_steps.go
```diff
@@ -297,3 +308,57 @@ func probeImageForShell(ctx context.Context, image string) (shell, tempfile stri
...
+
+	if len(e.Stdout) > 0 {
+		out.WriteString("\nstandard out:\n")
+		printOutput(e.Stdout)
+	}
+
+	if len(e.Stderr) > 0 {
+		out.WriteString("\nstandard error:\n")
+		printOutput(e.Stderr)
+	}
```
eseliger, line 355: Do you think they should be printed separately? I don't really know many places where that's the case (CI providers, GH actions, ..) usually print everything together (and sometimes it could be confusing I guess if:

```
stdout: Processing resource file X.yml
stderr: Failed to process.
stdout: Processing resource file Y.yml
stdout: Done, 1 errors.
```
, how would I know which of the files failed. Maybe it's just me though, so leaving it up to you and approve
)

eseliger (approved):
Nayse

internal/campaigns/run_steps.go
mrnugget: Well, we always set it to 1 in _our `src campaigns [apply|preview]`_, right? But take a look at the screenshot in the PR description: the step fails with exit code 127.

internal/campaigns/run_steps.go
mrnugget: Yeah, they print everything together because they stream the output, we don't. We buffer and then print. And since we buffer them separately we can't easily weave them back together to get to your example output. We'd need to record some timing information for that, but that's a hole I don't want to jump into. Or we could use a custom entity that buffers and prefixes both.

Honestly, though, I like having them separate. I hate having to dig through lines and lines of output to get to the error messages. It's something that annoys me about CI providers actually 😄

internal/campaigns/run_steps.go
eseliger: Ah, I see, cool

internal/campaigns/run_steps.go
eseliger: we could use the same buffer though, no? 🤔 I mean, which tool prints simultaneously on both out and err 🤔 

But if you like it better, go for it, I was just scared that some tools might use the channels in a weird way that only works when they're printed next to each other, like in my sample.
Probably an unreasonable fear though 😬

mrnugget: Yeah, sure, we could use the same buffer, but then we'd just have everything together in one without knowing at a quick glance what's error or output. And considering how noisy some tools are... If we run into issues, we can revisit this, I think.

eseliger: totally

PR #326, opened 2020-09-22
Check that Docker and git are available when executing campaign spec

mrnugget:
We had the check for git before, but lost it in the switch to the new
model.

The Docker check I added after running into a hard-to-decipher error
when trying to run `src campaigns apply` on a VM without Docker.

Here's what it looks like:

screenshot_2020-09-22_16 37 30@2x

### DIFF
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"os/exec"
 	"path"
 	"runtime"
 	"strings"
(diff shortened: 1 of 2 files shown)


cmd/src/campaigns_common.go
```diff
@@ -331,3 +341,16 @@ func formatTaskExecutionErr(err campaigns.TaskExecutionErr) string {
 		err.Logfile,
 	)
 }
+
+func checkExecutable(cmd string, args ...string) error {
+	if err := exec.Command(cmd, args...).Run(); err != nil {
+		return fmt.Errorf(
+			"failed to execute \"%s %s\":\n\t%s\n\n'src campaigns' require %q to be available.",
```
eseliger, line 348:
```suggestion
			"failed to execute \"%s %s\":\n\t%s\n\n'src campaigns' requires %q to be available.",
```

eseliger (approved):

PR #323, opened 2020-09-21
Print diff stats of completed changeset specs

mrnugget:
This PR adds the `-v` behaviour of printing diff stats for each repository in which changes were made.

Screenshot:

screenshot_2020-09-22_17 01 27@2x

Only shows up in `-v`.

Note: I also added the number to the `Executing steps` in `Executing steps in %d repositories` since that is _always_ the information I'm looking for.

PR #327, opened 2020-09-23
Refactor progress printer for spec execution into own function

mrnugget:
I hit merge too early in https://github.com/sourcegraph/src-cli/pull/323 and forgot to do this refactoring in it.

So, here we go. Upside: easier to review :)


eseliger (approved):
bueno :)

LawnGnome (approved):

PR #328, opened 2020-09-23
Handle interrupt signal in src campaigns apply|preview

mrnugget:
Before this change Docker containers kept on running after the interrupt signal was sent to the process and killing it. This was a slight regression, since it was previously fixed by #169.

What this commit does is bring back the behaviour of #169 and also makes sure that we print the correct error message when we canceled a running step.

It also makes sure that we don't enqueue more tasks/steps after being canceled: that's the code in `(*executor).Start()` that aborts the `Range` call if the context has been cancelled/timed-out and aborts execution of a task if the goroutine already blocked on `x.par.Acquire()`.

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -19,6 +19,10 @@ type TaskExecutionErr struct {
 	Repository string
 }
 
+func (e TaskExecutionErr) Cause() error {
+	return e.Err
+}
+
 func (e TaskExecutionErr) Error() string {
 	return fmt.Sprintf(
 		"execution in %s failed: %s (see %s for details)",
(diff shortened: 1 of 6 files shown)


internal/campaigns/executor.go
```diff
@@ -110,13 +114,25 @@ func (x *executor) LogFiles() []string {
 
 func (x *executor) Start(ctx context.Context) {
 	x.tasks.Range(func(k, v interface{}) bool {
+		select {
+		case <-ctx.Done():
+			return false
```
eseliger, line 119: good protective approach

eseliger (approved):

PR #331, opened 2020-09-30
Add first high-level integration test for Executor

mrnugget:
I present to you: the first high-level integration test to make sure that executing steps works.

Am I proud of this? Yes and no. Yes, because it gives me confidence that something works and it is a nice integration test. No, because, well, look, I created a little Docker "emulator" in Bash and I'm not sure how I feel about that now.

Edit: also, let me add: it creates a ZIP file in memory, mocks the Sourcegraph backend, actually _runs commands in a temp directory_ and allows us to assert that changes were made to the files!

### DIFF
--- a/internal/campaigns/executor_test.go
+++ b/internal/campaigns/executor_test.go
@@ -22,6 +22,10 @@ import (
 )
 
 func TestExecutor_Integration(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Test doesn't work on Windows because dummydocker is written in bash")
+	}
+
 	addToPath(t, "testdata/dummydocker")
 
 	repo := &graphql.Repository{
--- a/internal/campaigns/executor_test.go
+++ b/internal/campaigns/executor_test.go
@@ -38,7 +38,7 @@ func TestExecutor_Integration(t *testing.T) {
 	}
 
 	steps := []Step{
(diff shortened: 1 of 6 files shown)


eseliger (approved):
I like it, mocks the necessary parts but keeps us away from actually having to run an extensive campaign, which likely has many failure points.

However, it seems windows hates your test 😬

mrnugget:
> However, it seems windows hates your test 😬

Yeah :( I'll try to see what I can do, but ... writing PowerShell? [Probably not](https://www.youtube.com/watch?v=9X_ViIPA-Gc)

eseliger:
I guess writing it in go would work, although a little overkill 😬

LawnGnome (approved):
I assume the concern here is running the commands in CI, where we might not have a real Docker available?

I'm surprised there isn't a way to run Docker from within a GitHub Action: we can't be the first to have this issue. Nothing leaps off the page after some Googling, though.

I like this in general, though, and I don't want to let perfect be the enemy of good.

mrnugget:
> I assume the concern here is running the commands in CI, where we might not have a real Docker available?

Yeah, exactly. And it's also nice that the tests are super fast to run. I'll try to see if I can find out how to run Docker in Docker in GitHub actions, but I was and am afraid that it's not easy to do without changing how we run containers in some way.

PR #334, opened 2020-10-01
PROTOTYPE: Add support for templated campaign specs

mrnugget:
**NOTE FOR REVIEWERS**: This is a prototype and should be treated as such, which means: let's review idea/design/feasibility instead of concrete implementation details (but if you do have concrete feedback on the implementation, feel free to leave it as such).

---

This is a prototype to add **templating** to the `steps` definitions in a campaign spec YAML file.

It's similar to the templating allowed in [GitHub Actions](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions).

Here is an excerpt from a campaign spec that uses templating. It runs comby _only over the search results_ found in the repository by the `repositoriesMatchingQuery` (that's the `${{ .Repository.SearchResultPaths }}` in the spec) and then runs `goimports` _only over the files modified by comby_ (the `${{ .PreviousStep.ModifiedFiles }}`):

```yaml
name: templated-campaign
description: Use templates

on:
  - repositoriesMatchingQuery: lang:go fmt.Sprintf("%d", :[v]) patterntype:structural -file:vendor count:10

steps:
  - run: comby -in-place 'fmt.Sprintf("%d", :[v])' 'strconv.Itoa(:[v])' ${{ .Repository.SearchResultPaths }}
    container: comby/comby
  - run: goimports -w ${{ .PreviousStep.ModifiedFiles }}

changesetTemplate:
  # [... changesetTemplate is unchanged ...]
```

### Screenshot

image

### Details

This is a tested prototype that works. Executing campaigns like this is _much_ faster because comby only has to look at and touch the files it needs to modify, same goes for goimports.

What else is available in the templates? [This is the definition of `StepContext`](https://github.com/sourcegraph/src-cli/blob/4022b92bc73a13c0011c94cea9f630c023629d04/internal/campaigns/run_steps.go#L197-L214), which is passed into the `step.run` string before executing a step.

That means a user can access:

* `${{ .Repository.SearchResultPaths }}`
* `${{ .Repository.Name }}` and [the other fields and methods on `*graphql.Repository`](https://github.com/sourcegraph/src-cli/blob/4022b92bc73a13c0011c94cea9f630c023629d04/internal/campaigns/graphql/repository.go#L27-L58)
* `${{ .PreviousStep.ModifiedFiles }}`
* `${{ .PreviousStep.AddedFiles }}`
* `${{ .PreviousStep.DeletedFiles }}`

### Drawbacks

- there are no shortened aliases available as variables (`${{ search_result_paths }}` for example). Why? Because I couldn't come up with a consistent naming scheme that allows us to have a hierarchy too (like `previous_step.modified_files` etc.). That doesn't mean we shouldn't do it though, since I don't like the `.Foobar` syntax of Go's templating system.
  - **Edit**: I managed to get lower_case names + field access working (see [here](https://gist.github.com/mrnugget/774812fe534fb019a80ad9be2c364a57)), which means we could use that instead.
- it might not be clear to the user that they have access to _file paths_ since they're defining `*repositories*MatchingQuery` in their campaign spec. There might be a better way to define this.
- with a large number of files and search results the rendered commands might become _very long_ and thus hard to debug

### DIFF
--- a/internal/campaigns/run_steps.go
+++ b/internal/campaigns/run_steps.go
@@ -191,7 +191,10 @@ func parseStepRun(stepCtx StepContext, run string) (*template.Template, error) {
 	t := template.New("step-run").Delims("${{", "}}")
 
 	funcMap := template.FuncMap{
-		"repository_name": func() string { return stepCtx.Repository.Name },
+		"repository_name":    func() string { return stepCtx.Repository.Name },
+		"search_results":     stepCtx.Repository.FileMatchPathList,
+		"search_result_list": stepCtx.Repository.FileMatchPathList,
+		"join":               func(sl []string, sep string) string { return strings.Join(sl, sep) },
 	}
 
 	funcMap["modified_files"] = stepCtx.PreviousStep.ModifiedFiles
--- a/internal/campaigns/run_steps.go
+++ b/internal/campaigns/run_steps.go
@@ -65,7 +65,6 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
 
 	for i, step := range steps {
 		logger.Logf("[Step %d] docker run %s %q", i+1, step.Container, step.Run)
-		reportProgress(step.Run)
 
 		cidFile, err := ioutil.TempFile(tempDir, repo.Slug()+"-container-id")
 		if err != nil {
@@ -101,15 +100,19 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
 		if i > 0 {
 			stepContext.PreviousStep = results[i-1]
 		}
+
 		tmpl, err := parseStepRun(stepContext, step.Run)
 		if err != nil {
 			return nil, errors.Wrap(err, "parsing step run")
 		}
 
-		if err := tmpl.Execute(fp, stepContext); err != nil {
+		var buf bytes.Buffer
+		if err := tmpl.Execute(io.MultiWriter(&buf, fp), stepContext); err != nil {
 			return nil, errors.Wrap(err, "executing template")
 		}
 		fp.Close()
+
+		reportProgress(buf.String())
 
 		const workDir = "/work"
 		cmd := exec.CommandContext(ctx, "docker", "run",
--- a/internal/campaigns/run_steps.go
+++ b/internal/campaigns/run_steps.go
@@ -101,7 +101,7 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
 			stepContext.PreviousStep = results[i-1]
 		}
 
-		tmpl, err := parseStepRun(stepContext, step.Run)
+		tmpl, err := parseStepRun(step.Run)
 		if err != nil {
 			return nil, errors.Wrap(err, "parsing step run")
 		}
@@ -190,7 +190,7 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
 	return diffOut, err
 }
 
-func parseStepRun(stepCtx StepContext, run string) (*template.Template, error) {
+func parseStepRun(run string) (*template.Template, error) {
 	return template.New("step-run").Delims("${{", "}}").Parse(run)
 }
 
--- a/internal/campaigns/run_steps.go
+++ b/internal/campaigns/run_steps.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"text/template"
 	"time"
 
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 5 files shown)


internal/campaigns/run_steps.go
```diff
@@ -157,6 +182,70 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
 	return diffOut, err
 }
 
+func parseStepRun(stepCtx StepContext, run string) (*template.Template, error) {
+	t := template.New("step-run").Delims("${{", "}}")
+
+	funcMap := template.FuncMap{
+		"repository_name":    func() string { return stepCtx.Repository.Name },
```
chrispine, line 189: I assume this is a function (instead of a string) because the repository name can change? How/when does that happen?

internal/campaigns/run_steps.go
```diff
@@ -157,6 +182,70 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
...
+		"search_results":     stepCtx.Repository.FileMatchPathList,
+		"search_result_list": stepCtx.Repository.FileMatchPathList,
+		"join":               func(sl []string, sep string) string { return strings.Join(sl, sep) },
+	}
+
+	funcMap["modified_files"] = stepCtx.PreviousStep.ModifiedFiles
```
chrispine, line 195: Just curious as to why this line is separate from the rest of the items in the FuncMap.

chrispine:
This is really cool! I love it

internal/campaigns/run_steps.go
mrnugget: It's a function because (a) `FuncMap` only accepts functions and (b) we want to add it to the `FuncMap` so we can access the repository name through a `lower_case` var. I think it looks cleaner than accessing the repository name through the `StepContext`, which you can still do with `${{ .Repository.Name }}`.

It was a design idea we had yesterday and tried out, not set in stone :)

internal/campaigns/run_steps.go
mrnugget: Because the `PROTOTYPE` in the title implies "hacky shit, don't expect anything to make sense" 😄 It's a leftover.

chrispine:
I love how this is coming together! Thoughts on the drawbacks you mentioned:

- It's okay if our first release of this is only long-form names, as long as those don't break later. We don't have to figure out the perfect short-form naming right away.
- I feel like `Repository.SearchResultPaths` is pretty clear that it's paths, not repos, even though the paths come out of `repositoriesMatchingQuery`. (IMHO)
- This could be an issue. Plus, don't shells often have limits on how long a command can be? Can each item be run as a separate command?

LawnGnome:
> * there are no shortened aliases available as variables (`${{ search_result_paths }}` for example). Why? Because I couldn't come up with a consistent naming scheme that allows us to have a hierarchy too (like `previous_step.modified_files` etc.). That doesn't mean we shouldn't do it though, since I don't like the `.Foobar` syntax of Go's templating system.
>   
>   * **Edit**: I managed to get lower_case names + field access working (see [here](https://gist.github.com/mrnugget/774812fe534fb019a80ad9be2c364a57)), which means we could use that instead.

Are there any drawbacks to how that's implemented? If it supports the full range of functionality the default Go template syntax does, I feel like we should just make that the standard and be done with it, since I agree it's more readable.

> * with a large number of files and search results the rendered commands might become _very long_ and thus hard to debug

I think our limit here is going to be 131,072 bytes on Linux, regardless of configuration, because we'll hit [`MAX_ARG_STRLEN`](https://unix.stackexchange.com/questions/120642/what-defines-the-maximum-size-for-a-command-single-argument) when starting the container.

So, yeah, that's not a lot when you're dealing with a big vendored `node_modules`.

The ideal solution to me feels like it would be to do something sort of `xargs`-y: detect if we're going to go over `MAX_ARG_STRLEN` or `ARG_MAX` as configured, and then run the Docker container repeatedly with different subsets of the template expansion. Unfortunately, that probably opens its own can of worms in terms of what happens if the different runs of the same step somehow interact badly.

For now, we could probably just detect this situation independent of template expansion when running the campaign steps, and warn the user.

mrnugget:
> Are there any drawbacks to how that's implemented? If it supports the full range of functionality the default Go template syntax does, I feel like we should just make that the standard and be done with it, since I agree it's more readable.

The most obvious drawback is that we need to "manually" add aliases for fields: https://gist.github.com/mrnugget/774812fe534fb019a80ad9be2c364a57#file-lower_case_var_hierarchy-go-L36-L45

Reasonably sure, though, that we could somehow automate this by — hear me out — converting to JSON and then back to `map[string]interface{}`

mrnugget:
Closing in favor of #361.

PR #336, opened 2020-10-02
gzip http requests

chrispine:
This gzips http requests to the backend. It relies on [this](https://github.com/sourcegraph/sourcegraph/pull/14391) and partially fixes [this](https://github.com/sourcegraph/sourcegraph/issues/12840).

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -118,15 +118,19 @@ func (c *client) NewRequest(query string, vars map[string]interface{}) Request {
 
 func (c *client) NewHTTPRequest(ctx context.Context, method, p string, body io.Reader) (*http.Request, error) {
 	if c.supportsGzip == nil {
-		version, err := c.getSourcegraphVersion(ctx)
-		if err != nil {
-			return nil, err
-		}
-		supportsGzip, err := sourcegraphVersionCheck(version, ">= 3.21.0", "2020-10-12")
-		if err != nil {
-			return nil, err
-		}
+		// set to false, unless we have a new enough version
+		supportsGzip := false
 		c.supportsGzip = &supportsGzip
+
+		version, err := c.getSourcegraphVersion(ctx)
+
+		// ignore errors; we only care if the version is sufficently new
+		if err == nil {
+			supportsGzip, err = sourcegraphVersionCheck(version, ">= 3.21.0", "2020-10-12")
+			if err == nil {
+				c.supportsGzip = &supportsGzip
+			}
+		}
 	}
 
 	if *c.supportsGzip && body != nil {
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -10,10 +10,12 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
+	"regexp"
 	"strings"
 
+	"github.com/Masterminds/semver"
 	"github.com/hashicorp/go-multierror"
-	"github.com/jig/teereadcloser"
+	ioaux "github.com/jig/teereadcloser"
 	"github.com/kballard/go-shellquote"
 	"github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
@@ -54,7 +56,8 @@ type Request interface {
 
 // client is the internal concrete type implementing Client.
 type client struct {
-	opts ClientOpts
+	opts         ClientOpts
+	supportsGzip *bool
 }
 
 // request is the internal concrete type implementing Request.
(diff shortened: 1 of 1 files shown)


internal/api/api.go
```diff
@@ -13,10 +13,11 @@ import (
 	"strings"
 
 	"github.com/hashicorp/go-multierror"
-	"github.com/jig/teereadcloser"
+	ioaux "github.com/jig/teereadcloser"
```
mrnugget, line 16: Is that a leftover or is it required?

internal/api/api.go
```diff
@@ -164,7 +166,8 @@ func (r *request) do(ctx context.Context, result interface{}) (bool, error) {
 	}
 
 	// Create the HTTP request.
-	req, err := r.client.NewHTTPRequest(ctx, "POST", ".api/graphql", bytes.NewBuffer(reqBody))
+	zipped := codeintelutils.Gzip(bytes.NewBuffer(reqBody))
```
mrnugget, line 169: Ahh, that's handy :)

mrnugget (approved):

internal/api/api.go
chrispine: I have no idea why that happened... my editor must have done it on save. Weird. I will revert it.

chrispine: (Of course, being that my editor does it on save makes it slightly challenging to revert!)

LawnGnome (approved):

LawnGnome:
I approved this, but I do have a concern with the migration path: since there's no way for the Sourcegraph server to indicate its version before a request is made, we may attempt to speak `gzip` to a pre-3.21 version that isn't expecting it and doesn't know what to do with it.

One option that does occur to me would be to have the Sourcegraph server returns its version in a `Server` header: we could put a latch in the `src-cli` client that doesn't compress requests until it has received a response with the appropriate header and minimum version, at which point future requests _will_ be gzip encoded. This still deals with the campaign issue, since the campaign spec has to be uploaded before any changesets can be, but means that we don't necessarily get the benefits on the first request in a `src-cli` invocation.

(Spoiler: having a version field in the response is something I'll also be proposing in the versioning RFC.)

eseliger:
Your point is still valid because we do evaluate the search scope before and stuff, but I think campaign specs are created at the very end, since they're bundling changeset specs :)

LawnGnome:
> Your point is still valid because we do evaluate the search scope before and stuff, but I think campaign specs are created at the very end, since they're bundling changeset specs :)

Never trust my memory on a one coffee day. :laughing:

mrnugget:
> I approved this, but I do have a concern with the migration path: since there's no way for the Sourcegraph server to indicate its version before a request is made, we may attempt to speak `gzip` to a pre-3.21 version that isn't expecting it and doesn't know what to do with it.

How about we bring back the version checks as the first step here? We had this code before switching to the new workflow:

https://github.com/sourcegraph/src-cli/blob/d7d4bb2ae7481c745db53f7fed9a0c827253d247/cmd/src/patch_sets_create_from_patches.go#L108-L115

It does make an additional request, but I think that's negligible for now. And it gives us a back a way, today, to fence features. Once we return the version in a header we can switch out the "get the version" part and keep the check.

eseliger:
> It does make an additional request, but I think that's negligible for now.

That's probably right. Why not

internal/api/api.go
```diff
@@ -114,6 +117,35 @@ func (c *client) NewRequest(query string, vars map[string]interface{}) Request {
...
+	if err != nil {
+		return nil, err
+	}
+
+	if *c.supportsGzip {
+		req.Header.Set("Content-Encoding", "gzip")
+	}
+
+	return req, nil
+}
```
mrnugget, line 150: That means we try to gzip _every_ request — GraphQL, non-GraphQL — now, right? Doesn't that break things, since we only added gzip support to the GraphQL API?

mrnugget:
This broke the build, because `semver` is missing in `go.mod` and `go.sum`. And it also introduces a data race: multiple goroutines will now access and try to modify the `supportsGzip` field.

I'll open a PR that fixes this and also try to only use gzip in the campaigns part of src-cli, so we don't unintentionally break something.

mrnugget:
I opened a follow-up PR that addresses my comments here https://github.com/sourcegraph/src-cli/pull/343 so that we can release src-cli before branch cut.

PR #338, opened 2020-10-06
Show current worker goroutine status below progress bar

mrnugget:
GIF:

screenshot_2020-10-07_11 30 10

TODO:
- [x] Timestamp needs to not be truncated, but I'm not sure where to best put the timestamp.


mrnugget:
Timestamps truncated and moved to the right. TODOs checked off. Ready for the big review @sourcegraph/campaigns!

internal/output/progress_with_status_bars.go
```diff
@@ -0,0 +1,22 @@
...
+	barPtrs := make([]*ProgressBar, len(bars))
+	for i := range bars {
+		barPtrs[i] = &bars[i]
+	}
+
+	if !o.caps.Isatty {
+		newProgressWithStatusBarsSimple(barPtrs, statusBars, o)
+	}
+
+	return newProgressWithStatusBarsTTY(barPtrs, statusBars, o, opts)
```
mrnugget, line 21: Note to reviewers: `progressWithStatusbarsTTY` and `progressWithSatusBarsSimple` have a lot of similarities with `progressTTY` and `progressSimple` respectively. I've tried to extract some of the common stuff and embed `progressTTY` in `progressWithStatusBarsTTY`. That helped but not as much as it would've in a language where inheritance was a first class concept.

The problem in Go is this:

```go
type inner struct {
	name string
}

func (i *inner) String() string { return i.format() }
func (i *inner) format() string { return fmt.Sprintf("my name is %s", i.name) }

type outer struct {
	*inner
}

func (o *outer) format() string { return "i am outer!" }

func main() {
	i := &inner{name: "inner"}
	fmt.Println(i)

	o := &outer{inner: i}
	fmt.Println(o)
}
```

That prints:

```
my name is inner
my name is inner
```

[See this StackOverflow answer for details](https://stackoverflow.com/a/29390981). Short version: Go doesn't have virtual methods.

I might try to "invert" the code and have the structs' methods call functions, passing themselves in, but I'm not sure it's worth it.

Because right now I don't think the duplication is a problem. The code that is duplicated depends on the callee's state and doesn't contain a lot of duplicated logic. The hairy bits are not duplicated.

internal/campaigns/executor.go
```diff
@@ -60,12 +60,21 @@ type TaskStatus struct {
 	FinishedAt time.Time
 
 	// TODO: add current step and progress fields.
```
eseliger, line 62: is this todo comment related to the newly added line? If not, I'd add a newline here to avoid confusion

eseliger (approved):
Looks good! Thanks for taking on the effort of designing a TUI 👍

internal/campaigns/executor.go
mrnugget: I'll let @LawnGnome decide. He added the TODO and I wasn't sure whether he had something else in mind, like

```go
TotalSteps int
CurrentStep int
```

or something like that.

chrispine (approved):
LGTM, thanks!

LawnGnome (changes requested):
This LGTM! I think there's a lot of good stuff in here.

There is an issue right now with outputting to a non-TTY, so I'm going to mark this as changes requested, but I think there's a clear path to fixing it and getting this landed.

I also tested this on Windows. Behold:

out

LawnGnome (approved):

PR #339, opened 2020-10-07
Add 'campaigns new' command

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/14449 by adding a `src campaign new` command.

Screenshot:

screenshot_2020-10-07_15 17 25@2x

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 - The `published` flag in campaign specs may now be an array, which allows only specific changesets within a campaign to be published based on the repository name. [#294](https://github.com/sourcegraph/src-cli/pull/294)
-- A new `src campaign new` command creates a campaign spec YAML file with common values prefilled to make it easier to create a new campaign. [](https://github.com/sourcegraph/src-cli/pull/XXX)
+- A new `src campaign new` command creates a campaign spec YAML file with common values prefilled to make it easier to create a new campaign. [#339](https://github.com/sourcegraph/src-cli/pull/339)
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ All notable changes to `src-cli` are documented in this file.
 - Error reporting by `src campaign [preview|apply]` has been improved and now includes more information about which step failed in which repository. [#325](https://github.com/sourcegraph/src-cli/pull/325)
 - The default behaviour of `src campaigns [preview|apply]` has been changed to retain downloaded archives of repositories for better performance across re-runs of the command. To use the old behaviour and delete the archives use the `-clean-archives` flag. Repository archives are also not stored in the directory for temp data (see `-tmp` flag) anymore but in the cache directory, which can be configured with the `-cache` flag. To manually delete archives between runs, delete the `*.zip` files in the `-cache` directory (see `src campaigns -help` for its default location).
 - `src campaign [preview|apply]` now check whether `git` and `docker` are available before trying to execute a campaign spec's steps. [#326](https://github.com/sourcegraph/src-cli/pull/326)
+- The progress bar displayed by `src campaign [preview|apply]` has been extended by status bars that show which steps are currently being executed for each repository. [#338](https://github.com/sourcegraph/src-cli/pull/338)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 - The `published` flag in campaign specs may now be an array, which allows only specific changesets within a campaign to be published based on the repository name. [#294](https://github.com/sourcegraph/src-cli/pull/294)
+- A new `src campaign new` command creates a campaign spec YAML file with common values prefilled to make it easier to create a new campaign. [](https://github.com/sourcegraph/src-cli/pull/XXX)
 
 ### Changed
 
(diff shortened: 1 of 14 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 - The `published` flag in campaign specs may now be an array, which allows only specific changesets within a campaign to be published based on the repository name. [#294](https://github.com/sourcegraph/src-cli/pull/294)
+- A new `src campaign new` command creates a campaign spec YAML file with common values prefilled to make it easier to create a new campaign. [](https://github.com/sourcegraph/src-cli/pull/XXX)
```
eseliger, line 17: todo fill in number

cmd/src/campaigns_new.go
```diff
@@ -0,0 +1,131 @@
...
+	var (
+		fileFlag = flagSet.String("f", "campaign.yaml", "The name of campaign spec file to create.")
+	)
+
+	handler := func(args []string) error {
+		if err := flagSet.Parse(args); err != nil {
+			return err
+		}
+
+		if _, err := os.Stat(*fileFlag); !os.IsNotExist(err) {
```
eseliger, line 43: nitpick: technically, it can also be other errors than file already exists

cmd/src/campaigns_new.go
```diff
@@ -0,0 +1,131 @@
...
+			return err
+		}
+
+		author := campaigns.GitCommitAuthor{
+			Name:  "Sourcegraph",
+			Email: "campaigns@sourcegraph.com",
+		}
+
+		// Try to get better default values from git, ignore any errors.
+		if err := checkExecutable("git", "version"); err == nil {
```
eseliger, line 64: is this to test if git is available?

eseliger (approved):

cmd/src/campaigns_new.go
mrnugget: Yeah. Strictly speaking it's not necessary since we ignore the other errors anyway, but I like this check more.

cmd/src/campaigns_new.go
```diff
@@ -0,0 +1,131 @@
...
+			gitAuthorName, err := getGitConfig("user.name")
+			if err == nil && gitAuthorName != "" {
+				author.Name = gitAuthorName
+			}
+
+			gitAuthorEmail, err := getGitConfig("user.email")
+			if err == nil && gitAuthorEmail != "" {
+				author.Email = gitAuthorEmail
+			}
```
chrispine, line 73: This code makes it possible for only one of name/email to be set, and the other to come from the defaults. I kind of feel like it should either be both defaults, or both from git config.

chrispine (approved):

cmd/src/campaigns_new.go
```diff
@@ -0,0 +1,131 @@
...
+  branch: campaigns/hello-world # Push the commit to this branch.
+  commit:
+    author:
+      name: {{ .Author.Name }}
+      email: {{ .Author.Email }}
+    message: Append Hello World to all README.md files
+
+  # Change published to true once you're ready to create changesets on the code host.
+  published: false
+`
```
LawnGnome, line 131: I wonder if this spec is too opinionated: I know it's basically a copy of the hello world campaign in our docs, but I worry that it's not obvious enough where users need to dig in and change things, and if we'd be better served putting more obvious placeholders in instead for things like the name and description.

What do you think?

cmd/src/campaigns_new.go
```diff
@@ -0,0 +1,131 @@
...
+
+
+    $ src campaigns new -f campaign.spec.yaml
+
+`
+
+	flagSet := flag.NewFlagSet("new", flag.ExitOnError)
+
+	var (
+		fileFlag = flagSet.String("f", "campaign.yaml", "The name of campaign spec file to create.")
```
LawnGnome, line 35:
```suggestion
		fileFlag = flagSet.String("f", "campaign.yaml", "The name of the campaign spec file to create.")
```

LawnGnome (approved):
I'm not sure _exactly_ what the user experience we should shoot for here is, but the skeleton of the ~skeleton~ `new` subcommand looks good to me.

cmd/src/campaigns_new.go
mrnugget: I've been wondering about that too. What kept me from doing that was the fear that using meta-values like `run: <your shell script here>` or `name: campaign-name` makes it hard to learn what should go in the field (is `your shell script here` supposed to be a path? the name of a script? or a line?). I think with the values we have here it's obvious what's what and what needs to be changed. 

But I'll try something.

cmd/src/campaigns_new.go
mrnugget: Fixed!

cmd/src/campaigns_new.go
mrnugget: Done.

cmd/src/campaigns_new.go
```diff
@@ -102,23 +102,28 @@ func getGitConfig(attribute string) (string, error) {
...
 
-# Find all repositories that contain a README.md file.
+
+# "on" specifies on which repositories to execute the "steps"
 on:
+  # Example: find all repositories that contain a README.md file.
   - repositoriesMatchingQuery: file:README.md
 
-# In each repository, run this command. Each repository's resulting diff is captured.
+# steps are run in each repository. Each repository's resulting diff is captured.
```
LawnGnome, line 114: This feels inconsistent with the other comments in the template. I'm not going to hold up today's release for this, but I'll open a PR to play around with this wording some more.

PR #342, opened 2020-10-12
Show warning if 0 changeset specs have been created

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/14531 by showing a warning.

Please note that I think it should only ever be a warning, otherwise it'll be really hard to run `src campaign apply|preview` reliably in CI without some additional flags or something.

### DIFF
--- a/internal/output/emoji.go
+++ b/internal/output/emoji.go
@@ -3,6 +3,6 @@ package output
 // Standard emoji for use in output.
 const (
 	EmojiFailure = "❌"
-	EmojiWarning = "⚠️"
+	EmojiWarning = "❗️"
 	EmojiSuccess = "✅"
 )
--- a/internal/output/emoji.go
+++ b/internal/output/emoji.go
@@ -3,5 +3,6 @@ package output
 // Standard emoji for use in output.
 const (
 	EmojiFailure = "❌"
+	EmojiWarning = "⚠️"
 	EmojiSuccess = "✅"
 )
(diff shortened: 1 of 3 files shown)


internal/output/emoji.go
```diff
@@ -3,5 +3,6 @@ package output
 // Standard emoji for use in output.
 const (
 	EmojiFailure = "❌"
+	EmojiWarning = "⚠️"
```
eseliger, line 6: This triangle renders weirdly in chromium browsers, which includes VSCode and it's terminal.. But what can we do about it :| Just mentioning it because I ran into it before and looked why I keep pasting the wrong triangle in my editor. I can't find the issue anymore, though.

Chrome
image
iTerm (not chromium based)
image
vscode terminal (chromium based)
image

eseliger (approved):

internal/output/emoji.go
mrnugget: How about ❗️ instead?

mrnugget: Hmm, the problem might be 3-byte vs. 6-byte emoji?

eseliger: That might be the case! The❗️works at least.

mrnugget: Okay, switched to ❗️, that works in VS Code and here in Chrome. The emoji work in mysterious ways, I guess.

PR #343, opened 2020-10-13
Follow-up to gzip encoding of request body

mrnugget:
This is a follow-up to https://github.com/sourcegraph/src-cli/pull/336 and addresses this comment https://github.com/sourcegraph/src-cli/pull/336#discussion_r503788437 and the broken build (see https://github.com/sourcegraph/src-cli/pull/336#issuecomment-707605116) by

- making gzip compression for GraphQL requests an opt-in feature of the `api.Client` (vs. trying to encode every request)
- making the `gzip: bool` flag a property of the `api.Request` struct to fix the data race
- pulling the version/feature-flag check out of the `api.Client` and moving it into the `campaigns.Service` where it's used to determine whether gzip compression should be used or not
- adding back the the unit tests that came with the original implementation of the `sourcegraphVersionCheck` function
- reusing the existing GraphQL-request logic in `campaigns.Service` to query the version
- adding a changelog

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -10,12 +10,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"regexp"
 	"strings"
 
-	"github.com/Masterminds/semver"
 	"github.com/hashicorp/go-multierror"
-	"github.com/jig/teereadcloser"
+	ioaux "github.com/jig/teereadcloser"
 	"github.com/kballard/go-shellquote"
 	"github.com/mattn/go-isatty"
 	"github.com/pkg/errors"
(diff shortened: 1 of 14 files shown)


eseliger (approved):
🌟

internal/api/api.go
```diff
@@ -10,12 +10,10 @@ import (
 	"io/ioutil"
 	"net/http"
 	"os"
-	"regexp"
 	"strings"
 
-	"github.com/Masterminds/semver"
 	"github.com/hashicorp/go-multierror"
-	"github.com/jig/teereadcloser"
+	ioaux "github.com/jig/teereadcloser"
```
chrispine, line 16: It looks like `goimports` is what keeps adding the `ioaux` alias.

chrispine:
Thanks!

internal/api/api.go
mrnugget: And I think it's correct, since the package defined in `teereadcloser` is called `ioaux`: https://github.com/jig/teereadcloser/blob/953720c48e058a8869b07daa7db756dc7823b2e1/teereadcloser.go#L5

PR #345, opened 2020-10-13
changelog entry for validate command

uwedeportivo:

PR #347, opened 2020-10-13
doc: update changelog before release

LawnGnome:
This covers the changelog we need for #200, and also makes the wording consistent between entries.


LawnGnome:
Sorry, I missed #345.

PR #349, opened 2020-10-13
release: add a note about DEVELOPMENT.md

LawnGnome:
I'm not saying I forgot to check the checklist in DEVELOPMENT.md before releasing today, but... uh, I'm not saying I _didn't_ forget, either.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -1,12 +1,16 @@
 #!/usr/bin/env bash
 
-# Have you read DEVELOPMENT.md before running this script?
-#
-# No? Please stop and do so.
-#
-# Yes? You know what to do!
-
 set -euf -o pipefail
+
+read -p 'Have you read DEVELOPMENT.md? [y/N] ' -n 1 -r
+echo
+case "$REPLY" in
+    Y|y) ;;
+    *)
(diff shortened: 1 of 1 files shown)


release.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# If you haven't already read DEVELOPMENT.md, please stop and do so. Running
+# this script is _not_ the first thing you need to do to make a release.
```
slimsag, line 4: Can you make this a prompt instead? "Have you read DEVELOPMENT.md first? [y/N]"

slimsag (approved):

release.sh
LawnGnome: Done.

slimsag: haha, sorry, that's not exactly what I meant - I mean a Bash prompt for yes/no in case people run this script without reading the comments you have here.

e.g. https://github.com/sourcegraph/syntect_server/blob/master/publish.sh#L42-L49

LawnGnome: That makes more sense! Updated. :smile:

slimsag (approved):
🚀

eseliger (approved):
Bash is so nice .... Good work figuring out how to do that!

PR #350, opened 2020-10-14
Add draft option support for publish in changeset template

eseliger:
Using the introduced triple of true/false/"draft", users can now specify to open a PR in draft mode on the code host.

I'm opening this up for review, but I need to get https://github.com/sourcegraph/campaignutils/pull/3 merged first in order to merge this PR. Hence, the `go.mod` file contains some local hackery that I need to override (no pun intended) once merged.

Another PR in the series of https://github.com/sourcegraph/sourcegraph/issues/7998 towards the first milestone, getting the syntax support ready.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -18,7 +18,7 @@ require (
 	github.com/olekukonko/tablewriter v0.0.4 // indirect
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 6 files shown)


mrnugget (approved):
Approving, pending the removal of the line in `go.mod`.

Also: changelog entry is missing!

eseliger:
Regarding changelog entry:

Since this is a noop, I would like to keep it in the dark until it has an actual effect, I think that will reduce confusion :)

PR #352, opened 2020-10-15
Ensure that cache dir exists before downloading ZIP archive

mrnugget:
Reported by a customer.

### DIFF
--- a/internal/campaigns/archive_fetcher.go
+++ b/internal/campaigns/archive_fetcher.go
@@ -86,6 +86,10 @@ func fetchRepositoryArchive(ctx context.Context, client api.Client, repo *graphq
 		return fmt.Errorf("unable to fetch archive (HTTP %d from %s)", resp.StatusCode, req.URL.String())
 	}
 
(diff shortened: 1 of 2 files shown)


internal/campaigns/archive_fetcher.go
```diff
@@ -86,6 +86,10 @@ func fetchRepositoryArchive(ctx context.Context, client api.Client, repo *graphq
 		return fmt.Errorf("unable to fetch archive (HTTP %d from %s)", resp.StatusCode, req.URL.String())
 	}
 
+	if err := os.MkdirAll(filepath.Dir(dest), 0700); err != nil {
```
eseliger, line 89: I'm assuming this won't fail if the dest already exists?

eseliger (approved):

internal/campaigns/archive_fetcher.go
mrnugget: > // If path is already a directory, MkdirAll does nothing
> // and returns nil.

LawnGnome (approved):

sqs (approved):

PR #353, opened 2020-10-16
campaigns: move validation code to campaignutils

LawnGnome:
This turned out to be a nicer cleanup than I expected: the parse and validate helper functions in `cmd/src` were separate for mostly historical reasons, but now we can unify them.

Fixes sourcegraph/sourcegraph#14249.


cmd/src/campaigns_common.go
```diff
@@ -290,6 +286,34 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
...
+			}
+		} else {
+			// This shouldn't happen; let's just punt and let the normal
+			// rendering occur.
+			return nil, "", err
+		}
+	}
+
+	return spec, raw, nil
+}
```
eseliger, line 315: Just wondering if we should incorporate this logic into `ParseCampaignSpec`, then we wouldn't need this function that gets passed in a service, and it can't be forgotten. Are there use-cases where we explicitly would not want to have this error prettify logic?

eseliger (approved):

cmd/src/campaigns_common.go
LawnGnome: It's basically a layering thing: I don't want `internal/campaigns` (in this case, the `Service` methods) to have any specific knowledge of `internal/output`, so this code lives here to basically handle the rendering of any errors. This function is used by each of the campaign commands that have to parse a spec file, though, so it's still shared.

mrnugget (approved):

PR #355, opened 2020-10-16
Show failed tasks as errors in statusbars

mrnugget:
Yesterday I noticed that when executing steps fails in a repository we simply show "No changes" in the status bars. If you execute steps in a lot of repositories that means you'll need to wait until all of them are finished to get feedback on whether one of them failed.

This changes the behaviour from this

before

to this:

screenshot_2020-10-16_16 28 24

I'm not super happy with the `StatusText()` and `SingleLineError()` methods and how to best interface them, but then again I never found wrapped errors easy to manage. If you have any ideas: let me know!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Changed
 
+- The progress bar in `src campaign [preview|apply]` now shows when executing a step failed in a repository by styling the line red and displaying standard error output. [#355](https://github.com/sourcegraph/src-cli/pull/355)
+
 ### Fixed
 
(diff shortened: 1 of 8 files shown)


eseliger (approved):

eseliger:
> I'm not super happy with the StatusText() and SingleLineError() methods and how to best interface them, but then again I never found wrapped errors easy to manage. If you have any ideas: let me know!

No real idea off the top of my head but I also didn't find it too strange.

mrnugget:
Since there are no opposing voices, let's merge it then and see what happens.

LawnGnome:
Late to the party, but I think it's fine too. :partying_face:

PR #356, opened 2020-10-16
Fix flaky executor timeout test

mrnugget:
This gives the test a longer timeout to ensure that we've reached
`runSteps` when the timeout kicks in, even on slow CI machines.

And it changes the "sleep" to a power-nap-sleep-loop so that we only
have to wait 100ms until the test returns.


eseliger (approved):
Talked about it over zoom, lgtm!

PR #358, opened 2020-10-20
validate command: ability to create access token for first admin user

uwedeportivo:
will be used for executing automated regression tests against k8s cluster.

you can use a script similar to the one below to create the first admin and let it spit out the token

validate_script.json:
```
{
  "firstAdmin": {
    "email": "foo@sourcegraph.com",
    "username": "foo",
    "password": "{{ .foo_pw}}",
    "createAccessToken": true
  }
}
```

you can run it like so:
```
EXPORT SRC_ENDPOINT=http://localhost:3080
src validate -context foo_pw=testtesttest validate_script.json
```

### DIFF
--- a/cmd/src/validate.go
+++ b/cmd/src/validate.go
@@ -22,9 +22,10 @@ import (
 
 type validationSpec struct {
 	FirstAdmin struct {
-		Email    string
-		Username string
-		Password string
+		Email             string
+		Username          string
+		Password          string
+		CreateAccessToken bool
 	}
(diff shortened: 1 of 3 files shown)


davejrt:
Cool, maybe we could output this value to a file, so we can source it later so all the env vars are exported before we run any yarn commands.

davejrt (approved):

daxmc99 (approved):
Looking forward to using this on https://github.com/sourcegraph/sourcegraph/issues/13878

PR #361, opened 2020-10-22
Templating, lower_case variables and files

mrnugget:
This builds on top of https://github.com/sourcegraph/src-cli/pull/334 and adds **templating to campaign spec** including **lower-case variable names** and **access to STDOUT/STDERR**.

Updated campaign spec from previous prototype:

```yaml
name: templated-and-files
description: Use template variables and mount files in campaign specs

on:
  - repositoriesMatchingQuery: repo:automation-testing lang:go fmt.Sprintf("%d", :[v]) patterntype:structural -file:vendor count:5

steps:
  # Run comby over the search results in each repository:
  - run: comby -in-place -config /tmp/go-sprintf.toml -f ${{ join repository.search_result_paths "," }}
    container: comby/comby
    files:
      # Create files inside the container by specifying path and content here:
      /tmp/go-sprintf.toml: |
        [sprintf_to_strconv]
        match='fmt.Sprintf("%d", :[v])'
        rewrite='strconv.Itoa(:[v])'

  - run: echo "comby found the following problems:" >> CHANGELOG.md && cat /tmp/comby-output.txt >> CHANGELOG.md
    container: alpine:3
    files:
      # files also support templating:
      /tmp/comby-output.txt: ${{ previous_step.stdout }}

  - run: echo $MY_MODIFIED_FILES >> modified_files.txt
    container: alpine:3
    env:
      # env vars also support templating:
      MY_MODIFIED_FILES: ${{ join previous_step.modified_files "\n" }}

changesetTemplate:
  title: templated-and-files
  body: My not-first campaign!
  branch: campaign/testing-files
  commit:
    message: Code your code with templates
  published: false

```

### Details

* `${{ .Repository.SearchResultPaths }}` and `${{ repository.search_result_paths }}`
* `${{ .Repository.Name }}` and `${{ repository.name }}` and [the other fields and methods on `*graphql.Repository`](https://github.com/sourcegraph/src-cli/blob/4022b92bc73a13c0011c94cea9f630c023629d04/internal/campaigns/graphql/repository.go#L27-L58)
* `${{ .PreviousStep.ModifiedFiles }}` and `${{ previous_step.modified_files }}`
* `${{ .PreviousStep.AddedFiles }}` and `${{ previous_step.added_files }}`
* `${{ .PreviousStep.DeletedFiles }}` and `${{ previous_step.deleted_files }}`
* `${{ .PreviousStep.Stdout }}` and`${{ previous_step.stdout }}`
* `${{ .PreviousStep.Stderr }}` and `${{ previous_step.stderr}}`
* `${{ join <list> <sep> }}`, example: `${{ join repository.search_result_paths "\n" }}`
* `${{ split <string> <sep> }}`, example: `${{ split repository.name "/" }}`

### TODOs
- [x] ~Support mounting local files~ Erik and I decided against this, since that would make server-side execution harder.
- [x] Parse `env` too


chrispine:
Awesome! I like the new lower-case names. (Do we even need the longer-form names?)

mrnugget:
> Awesome! I like the new lower-case names. (Do we even need the longer-form names?)

Well, we don't need the longer-form names, but as I said in the sync meeting: the lower-case names have a bunch of drawbacks, since they're super-dynamically-typed (everything is a `map[string]interface{}`) and might end in "no value" messages instead of errors, etc.

I think for a v0.1 of templated-campaign specs that's fine as long as we flag it as experimental

LawnGnome:
> I think for a v0.1 of templated-campaign specs that's fine as long as we flag it as experimental

Agreed. I also prefer the lowercase, un-`.`-prefixed names, as we talked about last week, but I understand the implementation concerns. (I did play around with this a bit on Friday, but couldn't come up with a better approach with the standard library, and didn't want to get into the weeds of evaluating third party libraries right now!)

I think the design here is solid, and we should move on to reviewing and landing this once you're happy with the implementation. Nice job!

bobheadxi:
Question: does `changesetTemplate` get access to templating?

eseliger:
So in your scenario, the 3.21 branch would be at the head of the default branch, and then `upgrade-3.21` is a _second_ PR, that actually has changes and should target `3.21`?

bobheadxi:
> So in your scenario, the 3.21 branch would be at the head of the default branch, and then upgrade-3.21 is a second PR, that actually has changes and should target 3.21?

Yep - [where the `3.21` branch might not yet exist](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/dev/release/src/github.ts#L283:1-300:48) (will campaign create a branch? is another question)

So I imagine templating in the changeset template would be, if repo is deploy-sourcegraph, set base branch to `3.21`

eseliger:
I see, thanks for elaborating! Currently we don't have support for creating the base branch. Setting the base branch to an existing branch, however, is a [planned feature](https://github.com/sourcegraph/sourcegraph/issues/15070) (also specifying per repo is coming).

internal/campaigns/run_steps.go
```diff
@@ -271,3 +331,171 @@ func (e stepFailedErr) SingleLineError() string {
...
+			return strings.Join(list, sep)
+		},
+		"previous_step": func() map[string]interface{} {
+			result := map[string]interface{}{
+				"modified_files": stepCtx.PreviousStep.ModifiedFiles(),
+				"added_files":    stepCtx.PreviousStep.AddedFiles(),
+				"deleted_files":  stepCtx.PreviousStep.DeletedFiles(),
+			}
+
+			if stepCtx.PreviousStep.Stdout != nil {
```
eseliger, line 410: What happens if I use `${{ previous_step }}` in the first step? Would this be nil and panic?

eseliger (approved):
Nice 🌟 Very cool feature and makes using campaign specs so much simpler

mrnugget:
Here's another example campaign that can be run with this prototype (adding this here so I have a link to share):

```yaml
# Step 1: build the `ruplacer` Docker image
#   $ cat Dockerfile
#   FROM rust
#   RUN cargo install ruplacer
#
#   $ docker build -t ruplacer .
#
#
#   (Why use ruplacer? Because it supports `--subvert` which allows us to
#   replace `camelCase`, `snake_case`, `ThisCase`, `nocase`.)
#
# Step 2: use the src-cli prototype with templating support
#
#   $ cd src-cli && git fetch && git checkout mrnugget/templates-and-files
#   $ go build ./cmd/src -o ~/bin/src
#
# Step 3: update the `repositoriesMatchingQuery` to include or exclude file types.
#
# Step 4: run this campaign
#
#   $ src campaign preview -f update-language.campaign.yaml
name: update-language
description: This campaign changes occurrences of whitelist & blacklist to allowlist & denylist.

on:
  - repositoriesMatchingQuery: whitelist OR blacklist -file:scss$ -file:html$ repo:github.com/sourcegraph/sourcegraph

steps:
  - run: |
      cat /tmp/search-results | while read file;
      do
        ruplacer --subvert whitelist allowlist --go ${file} || echo "nothing to replace";
        ruplacer --subvert blacklist denylist --go ${file} || echo "nothing to replace";
      done
    container: ruplacer
    files:
      /tmp/search-results: ${{ join repository.search_result_paths "\n" }}

changesetTemplate:
  title: Replace usage of whitelist & blacklist with allowlist & denylist
  body: This replaces usages 
  branch: campaigns/allowlist-denylist
  commit:
    message: Use allowlist/denylist in wording
  published: false
```

internal/campaigns/run_steps.go
mrnugget: No. Say you have this in your spec:

```yaml
on:
  - repositoriesMatchingQuery: repo:automation-testing

steps:
  - run: |
      cat ${{ previous_step.modified_files }} >> previous_step_modified_files.txt &&\
        cat ${{ repository.search_result_paths }} >> repository_search_result_paths.txt
    container: alpine:3
```

that would produce two empty files. No panic, no errors.

internal/campaigns/campaign_spec.go
```diff
@@ -63,9 +63,10 @@ type OnQueryOrRepository struct {
 }
 
 type Step struct {
-	Run       string            `json:"run,omitempty" yaml:"run"`
-	Container string            `json:"container,omitempty" yaml:"container"`
-	Env       map[string]string `json:"env,omitempty" yaml:"env"`
+	Run       string                 `json:"run,omitempty" yaml:"run"`
```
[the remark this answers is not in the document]
mrnugget: No idea? I just copied what was already there.

mrnugget:
> I think the design here is solid, and we should move on to reviewing and landing this once you're happy with the implementation. Nice job!

That's what I meant with [requesting the review](https://github.com/sourcegraph/src-cli/pull/361#event-3921201176) 😄 But I should've been clearer:

I think the implementation as it is _now_ is fine to be merged as an experimental feature, which we will modify and tweak in the future. I checked that this branch here works with `sourcegraph@main` as well as this branch that knows about `step.files`: https://github.com/sourcegraph/sourcegraph/pull/15008. I also checked that `src-cli@main` works with https://github.com/sourcegraph/sourcegraph/pull/15008, so we don't run into "customer needs to upgrade because we merged an experimental feature"-problems.

I'd love to merge this as soon as possible.

LawnGnome:
> That's what I meant with [requesting the review](https://github.com/sourcegraph/src-cli/pull/361#event-3921201176) smile But I should've been clearer:

Sorry, I think I somehow thought this was still in draft when I looked at it Monday. Real review coming shortly!

internal/campaigns/executor_test.go
```diff
@@ -170,8 +187,13 @@ func TestExecutor_Integration(t *testing.T) {
 
 				diffsByName := map[string]*diff.FileDiff{}
 				for _, fd := range fileDiffs {
-					diffsByName[fd.OrigName] = fd
+					if fd.NewName == "/dev/null" {
+						diffsByName[fd.OrigName] = fd
```
LawnGnome, line 191: I like this affordance. :+1:

internal/campaigns/run_steps.go
```diff
@@ -87,29 +94,74 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
...
-			"--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", hostTemp, containerTemp),
-			"--entrypoint", shell,
-		)
-		for k, v := range step.Env {
-			cmd.Args = append(cmd.Args, "-e", k+"="+v)
+			"--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", runScriptFile.Name(), containerTemp),
 		}
+		for target, source := range filesToMount {
+			args = append(args, "--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", source.Name(), target))
+			defer os.Remove(source.Name())
```
LawnGnome, line 155: Is there a reason to not have this `defer` immediately after creating the temporary file above?

internal/campaigns/run_steps.go
```diff
@@ -271,3 +335,169 @@ func (e stepFailedErr) SingleLineError() string {
...
+	Added    []string
+	Deleted  []string
+}
+
+// ModifiedFiles returns the files modified by a step, whitespace-separated in a single string.
+func (r StepResult) ModifiedFiles() string {
+	if len(r.Files.Modified) == 0 {
+		return ""
+	}
+	return strings.Join(r.Files.Modified, " ")
```
LawnGnome, line 458: This pattern feels problematic: if the paths include spaces, then I presume users won't be able to distinguish a file called `a b` against files called `a` and `b`.

Fortunately, I think the solution might already be at hand: could we return `[]string` from these functions instead and require the use of the `join` function in the template?

(Additional thought: if we do this, we might want to add another function that can transform a `[]string` into a string with appropriately `sh`-quoted arguments, but that isn't on the critical path for this PR.)

internal/campaigns/run_steps.go
```diff
@@ -271,3 +335,169 @@ func (e stepFailedErr) SingleLineError() string {
...
+		file := line[3:]
+
+		switch line[0] {
+		case 'M':
+			result.Modified = append(result.Modified, file)
+		case 'A':
+			result.Added = append(result.Added, file)
+		case 'D':
+			result.Deleted = append(result.Deleted, file)
+		}
```
LawnGnome, line 499: It's a niche case, and probably irrelevant for an MVP, but would we also want to provide ways to access files that are renamed or copied as well by checking for [`R` and `C`, respectively](https://git-scm.com/docs/git-status#_short_format)?

chrispine (approved):
Excited to see this land!

internal/campaigns/run_steps.go
mrnugget: > Fortunately, I think the solution might already be at hand: could we return []string from these functions instead and require the use of the join function in the template?

Very good point! Will do that.

> (Additional thought: if we do this, we might want to add another function that can transform a []string into a string with appropriately sh-quoted arguments, but that isn't on the critical path for this PR.)

Very good idea! Something like `${{ to_shell_args repository.search_result_paths }}`

internal/campaigns/run_steps.go
mrnugget: No, you're right. I'll do that. I simply forgot about those.

internal/campaigns/executor_test.go
mrnugget: 👍 goes to @eseliger, I simply typed down what he told me :)

internal/campaigns/run_steps.go
mrnugget: Yes: I forgot. I kinda reorganised that code a bit and I think I left this lying around here. Good catch! Fixed.

internal/campaigns/run_steps.go
mrnugget: I added renames but I didn't add `C`, because ... I couldn't even get git to show a copied file: https://stackoverflow.com/questions/27198414/git-status-git-diff-c-not-detecting-file-copy-in-index And I would assume that other users then also expect new files to show up under `added` and not under copied.

If we get feedback to the contrary, we can add it.

PR #364, opened 2020-11-03
Do not marshal docker command to JSON when logging

mrnugget:
With the command being in JSON it's really hard to copy and run it to
debug something.

And since we don't ever need this output in JSON, I changed the code to
log log the un-rendered version of run.Step and then the full command.


eseliger (approved):
Why did we ever do this 🤔

chrispine (commented):
LGTM

PR #366, opened 2020-11-04
campaigns: widen permissions on mounted paths

### LINKED ISSUE #365: campaigns: create files/directories that might be mounted into the container with global permissions
### Scenario

1. The user is running on an environment where UIDs are mapped directly into Docker containers (so Linux, not macOS).
2. The user is not running as root, but as a normal user (eg `1000`).
3. The container being run does _not_ run as root, but runs as a different user that does not match the user's UID outside the container. (eg `2000`)

Therefore, anything we naïvely bind mount into the container may have permission issues.

### Shell scripts

As part of executing steps, `src-cli` creates a temporary file with the shell script to be executed and volume mounts it into the Docker container. This file is created using `ioutil.TempFile`, which [creates all files with permission `0600` as the current user](https://golang.org/src/io/ioutil/tempfile.go?s=1429:1487#L41).

Therefore, we end up in a scenario where there's a file owned by `1000`, with permission `0600`, that is being executed by a shell in a container running as `2000`, and that fails with a permission error.

There's no particular reason that temporary file needs to be `0600`, so we should create it as `0644` and sidestep this issue entirely.

### Workspaces

Similarly, `src-cli` unzips the repository being modified into a temporary directory. This directory and its children are created with minimal permissions, and are subject to the umask, both of which are likely to prevent writing to the workspace in a mixed-UID environment like that described above.

This one's a bit ickier: the easiest fix here (which is what we'll do to start with) is to replicate the user permissions out to the group/other bits on the workspace's files and directories, which essentially matches what Git does ([check the user executable bit and nothing else](https://medium.com/@tahteche/how-git-treats-changes-in-file-permissions-f71874ca239d)), but it has the potential for subtle breakage if the steps that are being run assume more about the permission structure of the repository.

It _may_ be possible to probe the container's UID and use ACLs to get more fine-grained behaviour, but this has some pretty tricky scenarios if there are multiple UIDs at play: if the first step runs as `2000` and the second step as `2001`, we'd have to manipulate the ACL on the fly.

Finally, the other option I considered was user namespaces, but since these [make changes to the global behaviour of Docker and require a `dockerd` option](https://docs.docker.com/engine/security/userns-remap/), that's a non-starter.

LawnGnome:
This fixes #365 by ensuring that files and workspaces mounted into campaign containers are world readable, writable, and executable as appropriate.

### DIFF
--- a/internal/campaigns/archive_fetcher.go
+++ b/internal/campaigns/archive_fetcher.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"strings"
 
+	"github.com/hashicorp/go-multierror"
 	"github.com/pkg/errors"
 	"github.com/sourcegraph/src-cli/internal/api"
 	"github.com/sourcegraph/src-cli/internal/campaigns/graphql"
@@ -67,6 +68,11 @@ func unzipToTempDir(ctx context.Context, zipFile, tempDir, tempFilePrefix string
 	if err != nil {
 		return "", err
 	}
+
+	if err := os.Chmod(volumeDir, 0777); err != nil {
+		return "", err
+	}
+
 	return volumeDir, unzip(zipFile, volumeDir)
 }
 
@@ -86,6 +92,10 @@ func fetchRepositoryArchive(ctx context.Context, client api.Client, repo *graphq
 		return fmt.Errorf("unable to fetch archive (HTTP %d from %s)", resp.StatusCode, req.URL.String())
 	}
 
+	// Unlike the mkdirAll() calls elsewhere in this file, this is only giving
+	// us a temporary place on the filesystem to keep the archive. Since it's
+	// never mounted into the containers being run, we can keep these
+	// directories 0700 without issue.
 	if err := os.MkdirAll(filepath.Dir(dest), 0700); err != nil {
 		return err
 	}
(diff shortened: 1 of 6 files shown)


internal/campaigns/archive_fetcher.go
```diff
@@ -67,6 +68,11 @@ func unzipToTempDir(ctx context.Context, zipFile, tempDir, tempFilePrefix string
 	if err != nil {
 		return "", err
 	}
+
+	if err := os.Chmod(volumeDir, 0777); err != nil {
```
eseliger, line 72: No action required:
I think ultimately we might even want to make this configurable (not needed in the next few months, but.. maybe some day:) )
..for ssh tokens and what not

internal/campaigns/archive_fetcher.go
```diff
@@ -145,6 +155,16 @@ func unzip(zipFile, dest string) error {
 			return err
 		}
 
+		// Since the container might not run as the same user, we need to ensure
+		// that the file is globally writable. If the execute bit is normally
+		// set on the zipped up file, let's ensure we propagate that to the
+		// group and other permission bits too.
+		if f.Mode()&0111 != 0 {
```
eseliger, line 162: Nice catch

internal/campaigns/run_steps.go
```diff
@@ -97,6 +97,12 @@ func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repositor
 		}
 		defer os.Remove(runScriptFile.Name())
 
+		// This file needs to be readable within the container regardless of
+		// the user the container is running as.
+		if err := runScriptFile.Chmod(0644); err != nil {
```
eseliger, line 102: does it not need to be executable?

eseliger (approved):
Nice 👍

internal/campaigns/archive_fetcher.go
LawnGnome: Yeah, this definitely gets more complicated if and when we're passing more in here than just the Git repo.

internal/campaigns/run_steps.go
LawnGnome: Counter intuitively, no! The Docker command basically boils down to this:

```sh
docker run --entrypoint /bin/bash -- sha256:CONTAINER_ID /tmp/some-horrible-script-name
```

Since the shell is the entrypoint, only that needs to be executable, and the script being run is just a regular old command line parameter.

LawnGnome:
OK, it looks like we have some Windows issues, so please hold while I figure out how much they matter. (My suspicion is: not much, given the different Docker execution model on Windows.)

chrispine (approved):
Thanks for getting to this so quickly!

internal/campaigns/archive_fetcher_windows_test.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/pkg/errors"
+)
+
+func mustHavePerm(t *testing.T, path string, want os.FileMode) error {
+	t.Helper()
+
+	have := mustGetPerm(t, path)
+
+	// Go maps Windows file attributes onto Unix permissions in a fairly trivial
```
eseliger, line 15: 🤯 
Thanks for finding this!

mrnugget (approved):

mrnugget:
Good stuff!

PR #369, opened 2020-11-05
Fix killing of unresponsive containers by adding init process

mrnugget:
Fixes https://github.com/sourcegraph/sourcegraph/issues/15314

This adds the `--init` flag to the `docker run` command so that the
entrypoint process (`bash` in our case) is started by a init process.
That fixes the behaviour of the interrupt signal simply having no
effect.

From what I could gather that's because PID 1 (which is the
`--entrypoint` in a Docker container) is protected and cannot be killed.
But `--init` is given the PID1 is an init process, which forwards the
signal to our entrypoint and kills it, exiting afterwards.

Resources:

- https://stackoverflow.com/questions/45718967/how-do-you-kill-a-docker-containers-default-command-without-killing-the-entire-c
- https://unix.stackexchange.com/questions/457649/unable-to-kill-process-with-pid-1-in-docker-container
- https://stackoverflow.com/questions/31538314/stopping-docker-container-from-inside

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- Unresponsive containers started by `src campaign [apply|preview]` can now be killed by hitting Ctrl-C. Previously the signal wasn't properly forwarded to the process in the container.
+
 ## 3.21.5
 
 ### Added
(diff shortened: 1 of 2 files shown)


eseliger (approved):
Call 1 800 555 5555 for Thorsten Ball container extermination services.

mrnugget:
This is a campaign spec to reproduce the issue:

```yaml
name: sleep
description: sleep

on:
  - repositoriesMatchingQuery: repo:automation-testing count:1

steps:
  - run: sleep 40
    container: alpine:3

changesetTemplate:
  title: SLEEP
  body: SLEEP
  branch: campaign/sleep
  commit:
    message: Sleep
  published: false
```

This couldn't be killed before with Ctrl-C, now it does.

LawnGnome (approved):
Nice!

I'm going to merge this once #368 lands so I can do a release.

PR #370, opened 2020-11-05
campaigns: restore pre-3.20 instance compatibility

LawnGnome:
The author name and e-mail fields were always sent, but the schema Sourcegraph 3.19 uses to validate changeset specs don't include these fields. Since we're doing version detection for feature flags (quirks) now anyway, let's use that to not set the author name and e-mail if they're not explicitly provided when talking to a 3.19 instance.

If the user provides an author name and/or e-mail and sends a changeset spec to a pre-3.20 instance, the request will fail, but since those are 3.20 features that's OK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


chrispine (approved):
LGTM, thanks @LawnGnome

mrnugget (commented):
Good stuff! Nice little struct :)

PR #372, opened 2020-11-10
if namespace is omitted, default to username

chrispine:
partial fix for https://github.com/sourcegraph/sourcegraph/issues/15058

### DIFF
--- a/cmd/src/campaigns_apply.go
+++ b/cmd/src/campaigns_apply.go
@@ -21,6 +21,8 @@ Usage:
 Examples:
 
     $ src campaigns apply -f campaign.spec.yaml
+  
+    $ src campaigns apply -f campaign.spec.yaml -namespace myorg
 
 `
 
--- a/cmd/src/campaigns_apply.go
+++ b/cmd/src/campaigns_apply.go
@@ -16,11 +16,11 @@ instance, creating or updating the described campaign if necessary.
 
 Usage:
 
-    src campaigns apply -f FILE -namespace NAMESPACE [command options]
+    src campaigns apply -f FILE [command options]
 
 Examples:
 
-    $ src campaigns apply -f campaign.spec.yaml -namespace myorg
+    $ src campaigns apply -f campaign.spec.yaml
 
 `
 
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -79,7 +79,7 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 	)
 	flagSet.StringVar(
 		&caf.namespace, "namespace", "",
-		"The user or organization namespace to place the campaign within. Default is the current authenticated user.",
+		"The user or organization namespace to place the campaign within. Default is the currently authenticated user.",
 	)
 	flagSet.StringVar(&caf.namespace, "n", "", "Alias for -namespace.")
 
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -169,10 +169,6 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 		defer specFile.Close()
 	}
 
-	if flags.namespace == "" {
-		errs = multierror.Append(errs, &usageError{errors.New("a namespace must be provided with -namespace")})
-	}
-
 	opts := campaigns.ExecutorOpts{
 		Cache:      svc.NewExecutionCache(flags.cacheDir),
 		Creator:    svc.NewWorkspaceCreator(flags.cacheDir, flags.cleanArchives),
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -79,7 +79,7 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 	)
 	flagSet.StringVar(
 		&caf.namespace, "namespace", "",
-		"The user or organization namespace to place the campaign within.",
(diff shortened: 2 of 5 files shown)


internal/campaigns/service.go
```diff
@@ -312,7 +312,31 @@ query NamespaceQuery($name: String!) {
...
+				CurrentUser struct {
+					ID string `json:"id"`
+				} `json:"currentUser"`
+			} `json:"data"`
+		}
+		if ok, err := svc.client.NewRequest(usernameQuery, nil).DoRaw(ctx, &resp); err != nil || !ok {
+			return "", errors.WithMessage(err, "failed to resolve namespace: no user logged in")
+		}
+
+		return resp.Data.CurrentUser.ID, nil
```
chrispine, line 337: Do I need a nil check for `CurrentUser` or `CurrentUser.ID`?

eseliger: Hmm, I think currentUser is nilable. That would mean "not logged in, I guess"

cmd/src/campaigns_common.go
```diff
@@ -169,10 +169,6 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 		defer specFile.Close()
```
eseliger, line 169: maybe in l.82 a doc change would be nice. Like it defaults to the authenticated user's namespace or so

eseliger (approved):
LGTM, besides your nil check

cmd/src/campaigns_common.go
chrispine: This is the ticket for the doc changes: https://github.com/sourcegraph/sourcegraph/pull/15664

Did you mean something else?

eseliger: This little docstring I meant https://github.com/sourcegraph/src-cli/blob/ec0a7ab72e2bc3d258b744260d8a3871057766ee/cmd/src/campaigns_common.go#L82

internal/campaigns/service.go
chrispine: Actually, they can't be nil because they aren't pointers, so nil-checks won't even compile. (I'm assuming this is fine as is?)

eseliger: Ah, then I'd either make CurrentUser a pointer, or check for the empty string, but I think it is a valid case that the API returns an empty current user, so I think we should cover that here.
```suggestion
                 if resp.Data.CurrentUser.ID == "" {
                     return nil, errors.New("cannot resolve current user")
                 }
		return resp.Data.CurrentUser.ID, nil
```

cmd/src/campaigns_common.go
chrispine: Ah, good call. I added a note about the default.

cmd/src/campaigns_apply.go
```diff
@@ -16,11 +16,11 @@ instance, creating or updating the described campaign if necessary.
 
 Usage:
 
-    src campaigns apply -f FILE -namespace NAMESPACE [command options]
+    src campaigns apply -f FILE [command options]
 
 Examples:
 
-    $ src campaigns apply -f campaign.spec.yaml -namespace myorg
+    $ src campaigns apply -f campaign.spec.yaml
```
mrnugget, line 23: I think it might make sense to keep this in the examples:

```suggestion
    $ src campaigns apply -f campaign.spec.yaml
  
    $ src campaigns apply -f campaign.spec.yaml -namespace myorg
```

cmd/src/campaigns_common.go
```diff
@@ -79,7 +79,7 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 	)
 	flagSet.StringVar(
 		&caf.namespace, "namespace", "",
-		"The user or organization namespace to place the campaign within.",
+		"The user or organization namespace to place the campaign within. Default is the current authenticated user.",
```
mrnugget, line 82: Isn't it...?

```suggestion
		"The user or organization namespace to place the campaign within. Default is the currently authenticated user.",
```

internal/campaigns/service.go
```diff
@@ -312,7 +312,34 @@ query NamespaceQuery($name: String!) {
 }
 `
 
+const usernameQuery = `
+query {
```
mrnugget, line 316: I think it's good practice to name your queries, but might be wrong:

```suggestion
query GetCurrentUserID {
```

internal/campaigns/service.go
```diff
@@ -332,7 +359,7 @@ func (svc *Service) ResolveNamespace(ctx context.Context, namespace string) (str
 	if result.Data.Organization != nil {
 		return result.Data.Organization.ID, nil
 	}
-	return "", errors.New("no user or organization found")
+	return "", errors.New("failed to resolve namespace `" + namespace + "`: no user or organization found")
```
mrnugget, line 362: I don't know if you know about `%q` yet, but in case you don't: you're in luck!

```suggestion
	return "", fmt.Errorf("failed to resolve namespace %q: no user or organization found", namespace)
```

mrnugget (approved):
Approve so. you can merge it after you've done it, but

- [x] please add a changelog entry!

**Before you merge your documentation changes in `sourcegraph/sourcegraph`:**

- [x] merge this
- [x] release a new src-cli version (see `DEVELOPMENT.md`)
- [ ] make sure to bump the version in `sourcegraph/sourcegraph`

🌶 Hot Pro Tip: 🌶 include the version bump in our documentation PR so that the docs are updated with the bump.

internal/campaigns/service.go
chrispine: I did know about it; just didn't have a strong preference between the two in this case.

Issue #371, opened 2020-11-10
Local git config may break campaigns


eseliger:
A customer had a local git config to include ANSI color codes in the git diff output. That config is obviously also picked up when we shell out to git, and the result was a git diff decorated with ANSI color codes, that is not parsable. We should ignore any local configs (or at least ones that affect the output).

PR #373, opened 2020-11-11
Don't use git config that might be present on the host

eseliger:
This can cause unexpected git behavior, like a different diff tool, ANSI coloring in the diff and probably other bad magic.

Closes https://github.com/sourcegraph/src-cli/issues/371


mrnugget (approved):
SWEET! Yes! That's what I call a "nice catch"!

mrnugget:
I added a changelog entry and am going to merge this and cut a release.

PR #374, opened 2020-11-12
Cache results in repository even if diff was empty

mrnugget:


LawnGnome (approved):
Approved, but since I was on Zoom with Thorsten when this was being developed, someone else should probably take a look at this too. :smile:

chrispine (approved):
Nice!

PR #377, opened 2020-11-12
CI: remove last vestiges of Travis

LawnGnome:
As @chrispine noticed when releasing 3.21.9, we have a dangling reference to Travis in our release checklist. We also still had the old CI badge kicking around, so let's fix them both at once.


chrispine (approved):
thanks!

PR #378, opened 2020-11-13
Fix out-of-bounds panic in progress printer

mrnugget:
We ran into an out-of-bounds panic when we had more tasks reporting as
"currently running" than the number of status bars we had.

That was due to a lack of defensive checks in the face of a possible
race condition.

The race condition leads the `PrintStatuses` method seeing more
"currently running" tasks than there can technically be.

The race condition is between the `PrintStatus` method that prints the
slices of pointers to `*TaskStatus` and the tasks themselves updating
the status non-atomatically.

At least that's what it seems like. I'm not 100% sure about the cause
yet, but I could reproduce the issue by simply injecting more statuses
at the top of `PrintStatuses`. The code here fixes it.

---

I'm still trying to find out exactly why it happens, but I do think this code should be in there in _any case_, that's why I'm opening up this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- Fixed a bug that could cause `src campaign [apply|preview]` to crash in rare circumstances when executing a campaign spec due to a bug in the logic for the progress bar. [#378](https://github.com/sourcegraph/src-cli/pull/378)
+
 ### Removed
 
 ## 3.21.9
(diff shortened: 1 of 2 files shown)


eseliger (approved):
Seems a little hacky but I don't know another simple way to get it in on a Friday evening, so LGTM

chrispine (approved):
LGTM

LawnGnome (approved):
Yeah, I also don't understand exactly how this happened, but as you say, let's get the defensive check in regardless.

mrnugget:
HA! Yesterday, in the shower, I figured out how it happens!

It's a race condition between the `PrintStatuses` method and the `(*executor).Start` method enqueuing tasks _one-by-one_. While the executor uses a semaphore to ensure that a maximum of `NumParallelism` jobs are running at the same time, the tasks themselves can update their status independently of any locking mechanism. That's the important bit.

So what happens is this.

Say you have tasks A, B, C, D.  A and B are currently running.

`PrintStatus` iterates over the tasks. `PrintStatuses` sees that A and B are running. Then it looks at C. Not running. And here's the race condition: before `PrintStatus` looks at job D, task A is finished (which can happen really fast, if results are cached, as was the case for the customer that ran into this) and task D is started. So now `PrintStatus` thinks D is running. End result: 3 tasks marked as running, even though only 2 should be.

A solution would be a lock around the `[]*campaigns.TaskStatus` slice. I'll try to see if I can make that work. But I do think that this out-of-bounds check here is still something we should keep.

PR #380, opened 2020-11-13
api: add User-Agent header

LawnGnome:
TODO: actually get the version.

This would fix sourcegraph/sourcegraph#15769 when complete.


mrnugget:

LawnGnome:
Bah, I forgot I had this open when I implemented better version number handling for the feature flags... uh, feature in `src` all those months ago, and that was really the only blocker. Reopening with a complete implementation.

Issue #383, opened 2020-11-17
campaigns validate hangs


evanweible-wf:
src version: 3.22.0

I've tried running `src campaigns validate` on a couple different campaign specs (including the hello world example from the docs) and it just hangs, no output whatsoever. Waited a couple of minutes before giving up.

mrnugget:
@evanweible-wf did you use the `-f` flag with `src campaigns validate`? If you don't provide a `-f` flag, `src campaigns validate` will wait for input on standard input.

mrnugget:
I opened https://github.com/sourcegraph/src-cli/pull/384 to help with the problem of `src campaign [validate|preview|apply]` silently waiting for input if the user forgot a `-f` flag but provided an additional arg (a filename)

evanweible-wf:
Whoops, yes I did forget the `-f` :) Sorry about that! I'll go ahead and close this, and thanks for filing the other issue!

PR #384, opened 2020-11-17
Return error if additional args passed to campaigns command

mrnugget:
This should stop users from running into https://github.com/sourcegraph/src-cli/issues/383


eseliger (approved):
I feel like we might want to move away from defaulting to stdin... maybe only explicitly using -f -

mrnugget:
> I feel like we might want to move away from defaulting to stdin... maybe only explicitly using -f -

Agree.

PR #386, opened 2020-11-18
codeintel: Update utils package; remove auth from redirect URL

efritz:


mrnugget (approved):
Does this need a CHANGELOG entry?

LawnGnome (approved):
I think a `go mod tidy` would get rid of a few dangling checksums in `go.sum` after the dependency update, but LGTM.

efritz:
> Does this need a CHANGELOG entry?

Nah! This only affects auto-indexing and debugging.

mrnugget:
> Nah! This only affects auto-indexing and debugging.

Then feel free to merge

PR #387, opened 2020-11-18
cmd: add a hidden doc command

LawnGnome:
This generates ugly, but readable Markdown output that can be used within the Sourcegraph doc directory.

For an example of this in action, see https://github.com/sourcegraph/sourcegraph/pull/15906.

### DIFF
--- a/cmd/src/doc.go
+++ b/cmd/src/doc.go
@@ -35,6 +35,13 @@ Examples:
 	handler := func(args []string) error {
 		flagSet.Parse(args)
 
+		out := output.NewOutput(flagSet.Output(), output.OutputOpts{Verbose: *verbose})
+		if *outputFlag == "" {
+			out.WriteLine(output.Line(output.EmojiFailure, output.StyleWarning, "output directory must be set via -o"))
+			flagSet.Usage()
+			return &exitCodeError{exitCode: 1}
+		}
+
 		dr, err := newDocRenderer()
 		if err != nil {
 			return err
@@ -52,8 +59,6 @@ Examples:
 			"repos":        &reposCommands,
 			"users":        &usersCommands,
 		}
-
-		out := output.NewOutput(flagSet.Output(), output.OutputOpts{Verbose: *verbose})
 
 		pending := out.Pending(output.Line("", output.StylePending, "Rendering Markdown..."))
 		count := 0
--- a/cmd/src/doc.go
+++ b/cmd/src/doc.go
@@ -1,0 +1,269 @@
+package main
+
+import (
+	"bytes"
+	"flag"
+	"fmt"
+	"io"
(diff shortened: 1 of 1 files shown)


cmd/src/doc.go
```diff
@@ -0,0 +1,269 @@
...
+|------|-------------|---------------|
+{{- range .Flags -}}
+{{- "\n" -}}
+| ` + "`" + `-{{ .Name }}` + "`" + ` | {{ .Usage }} | {{ if .DefValue }}` + "`" + `{{ .DefValue }}` + "`" + `{{ end }} |
+{{- end }}
+{{ end }}
+
+## Usage
+
+` + "```" + `
```
mrnugget, line 254: Holy quote!

cmd/src/doc.go
```diff
@@ -0,0 +1,269 @@
...
+		usageFunc: func() {
+			fmt.Fprintln(flag.CommandLine.Output(), usage)
+		},
+	})
+}
+
+func openDocFile(base, fqcn string) (interface {
+	io.StringWriter
+	io.WriteCloser
+}, error) {
```
mrnugget, line 132: Not a blocker, just curious: why use this interface? Using `*os.File` here would make the function signature easier to read.

cmd/src/doc.go
```diff
@@ -0,0 +1,269 @@
...
+				return err
+			}
+			defer file.Close()
+
+			file.WriteString(content)
+			count++
+		}
+
+		return nil
+	}
```
mrnugget, line 118: Do we need to keep track of all the files that were created and then delete those in the output directory we didn't touch?

It's not a blocker, just something to think about, since we might run into the case where old commands leave documentation lying around.

mrnugget (approved):
Cool idea!

cmd/src/doc.go
LawnGnome: I thought about doing that, and could be convinced to add a flag, but I think your cursed `go generate` suggestion in https://github.com/sourcegraph/sourcegraph/pull/15906#discussion_r525911538 might be the right layer to handle that type of cleanup in.

cmd/src/doc.go
LawnGnome: Personal scarring, I think: I've found `os.File` hard to test around in the past, and just requiring the interfaces that we actually need would make it easier to write tests using this function in the future, but this might be a premature optimisation. I'd be easily convinced to just return `*os.File`.

cmd/src/doc.go
LawnGnome: The more I write raw strings in Go, the more I appreciate raw strings in... uh, most other languages.

eseliger (approved):
🤯

PR #388, opened 2020-11-18
Clean up partially-downloaded ZIP files if aborted

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/15865

This fixes a bug where src-cli would leave partially-downloaded ZIP
files behind when the user hit Ctrl-C while a download was ongoing.

The `fetchRepositoryArchive` would then run into a context-cancelation
error when doing the request or the `io.Copy` and not clean up the file.

The code here checks whether the cause of the error is a cancelation
and, if so, cleans up the zip file.

### DIFF
--- a/internal/campaigns/archive_fetcher.go
+++ b/internal/campaigns/archive_fetcher.go
@@ -34,7 +34,14 @@ func (wc *WorkspaceCreator) Create(ctx context.Context, repo *graphql.Repository
 	}
 
 	if !exists {
-		if err := fetchRepositoryArchive(ctx, wc.client, repo, path); err != nil {
+		err = fetchRepositoryArchive(ctx, wc.client, repo, path)
+		if err != nil {
+			if errors.Is(err, context.Canceled) || errors.Cause(err) == context.Canceled {
+				// If the context got cancelled while we were downloading the
+				// file, we remove the partially downloaded file.
+				os.Remove(path)
+			}
+
 			return "", errors.Wrap(err, "fetching ZIP archive")
 		}
 	}
--- a/internal/campaigns/archive_fetcher.go
+++ b/internal/campaigns/archive_fetcher.go
@@ -36,11 +36,10 @@ func (wc *WorkspaceCreator) Create(ctx context.Context, repo *graphql.Repository
 	if !exists {
 		err = fetchRepositoryArchive(ctx, wc.client, repo, path)
 		if err != nil {
-			if errors.Is(err, context.Canceled) || errors.Cause(err) == context.Canceled {
-				// If the context got cancelled while we were downloading the
-				// file, we remove the partially downloaded file.
-				os.Remove(path)
-			}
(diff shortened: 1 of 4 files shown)


efritz (approved):

eseliger (approved):
Nice test 🤓

internal/campaigns/archive_fetcher.go
```diff
@@ -34,7 +34,14 @@ func (wc *WorkspaceCreator) Create(ctx context.Context, repo *graphql.Repository
 	}
 
 	if !exists {
-		if err := fetchRepositoryArchive(ctx, wc.client, repo, path); err != nil {
+		err = fetchRepositoryArchive(ctx, wc.client, repo, path)
+		if err != nil {
+			if errors.Is(err, context.Canceled) || errors.Cause(err) == context.Canceled {
```
keegancsmith, line 39: Why not always remove in the case of an error?

keegancsmith (approved):
LGTM. However, this is relying on your program not being sigkilled so cleanup can run. To make downloading more robust I'd recommend downloading to `filepath.Join(dst, ".tmp")` (or something) and renaming once download is complete. This technically still won't be "safe", since you haven't fflushed, but that is probably fine. You can use something like the renameio pkg to handle doing this safely.

LawnGnome:
> To make downloading more robust I'd recommend downloading to `filepath.Join(dst, ".tmp")` (or something) and renaming once download is complete.

I think I agree with this too, with the caveat that I'd still keep the logic to clean up wherever possible, except applied to that temp file.

> You can use something like the renameio pkg to handle doing this safely.

Unfortunately, renameio doesn't work on Windows. _shakes fist_

internal/campaigns/archive_fetcher.go
mrnugget: You're right. There might be other errors that leave a partial file lying around.

mrnugget:
Regarding download-to-temp-file and rename: I don't think this is necessary, since I just changed the code to delete the file in case of _any error_ and nothing else will access the file while that method is running. It doesn't need to be an atomic operation. I'm going to go ahead and merge this, but if someone can tell me what I'm missing and why this needs to be an atomic download-and-swap I'll try to change it.

LawnGnome:
>  I'm going to go ahead and merge this, but if someone can tell me what I'm missing and why this needs to be an atomic download-and-swap I'll try to change it.

I think the concern here is the one Keegan called out: if `src` receives a `SIGKILL`, then the cleanup code never gets called, since you can't handle that signal.

mrnugget:
Ah, thanks! Now I get it. I'll think about that a bit more, because we also don't want to leave tmp files lying around and probably want to clean those up when starting. Implementing that could go hand-in-hand with my thoughts about making the archive-caching a bit more efficient (right now, if turned on, we'd cache all archives, which means you can accumulate a large number if your default branch updates often)

PR #389, opened 2020-11-19
Fix status bar updates getting lost

mrnugget:
This is super embarrassing, but ... well, here we are.


eseliger (approved):
OMG SO EMBARASSING

LGTM

LawnGnome:
I don't think I understand why this makes a difference. Wasn't `numStatusBars` still in scope at the point the later check occurred anyway? :confused:

mrnugget:
The `PrintStatuses` method is used as a callback and is called every ~second to print the current status.

The first time it's called, the progress bar and the status bars haven't been initialised, so the code behind `if p.progress == nil` is executed.

The second, third, ..., nth time it is called, the bars have been initialised, which means `if p.progress == nil` is _false_ and 🐛 the `numStatusBars` is stays `0` 🐛 .

LawnGnome:
> The `PrintStatuses` method is used as a callback and is called every ~second to print the current status.

Aha! That was the bit of context I'd forgotten. Thanks for that!

PR #390, opened 2020-11-19
More information in verbose mode to provide faster feedback loop

mrnugget:
Thoughts/opinions/ideas?

screenshot_2020-11-19_16 23 17

screenshot_2020-11-20_17 59 46

TODOs: 
- [x] Clean up the code

### DIFF
--- a/cmd/src/campaign_progress_printer.go
+++ b/cmd/src/campaign_progress_printer.go
@@ -138,18 +138,23 @@ func (p *campaignProgressPrinter) PrintStatuses(statuses []*campaigns.TaskStatus
 	}
 
 	for _, ts := range newlyCompleted {
+		var fileDiffs []*diff.FileDiff
+
+		if ts.ChangesetSpec != nil {
+			var err error
+			fileDiffs, err = diff.ParseMultiFileDiff([]byte(ts.ChangesetSpec.Commits[0].Diff))
+			if err != nil {
+				p.progress.Verbosef("%-*s failed to display status: %s", p.maxRepoName, ts.RepoName, err)
+				continue
+			}
+		}
+
 		if p.verbose {
 			p.progress.WriteLine(output.Linef("", output.StylePending, "%s", ts.RepoName))
 
 			if ts.ChangesetSpec == nil {
 				p.progress.Verbosef("  No changes")
 			} else {
-				fileDiffs, err := diff.ParseMultiFileDiff([]byte(ts.ChangesetSpec.Commits[0].Diff))
-				if err != nil {
-					p.progress.Verbosef("%-*s failed to display status: %s", p.maxRepoName, ts.RepoName, err)
-					continue
-				}
-
 				lines, err := verboseDiffSummary(fileDiffs)
 				if err != nil {
 					p.progress.Verbosef("%-*s failed to display status: %s", p.maxRepoName, ts.RepoName, err)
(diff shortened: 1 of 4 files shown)


eseliger (approved):
Since I've been around when this was made, I think I shouldn't approve, but I'm feeling wild so I do it anyways 🦊

cmd/src/campaign_progress_printer.go
```diff
@@ -134,13 +134,30 @@ func (p *campaignProgressPrinter) PrintStatuses(statuses []*campaigns.TaskStatus
...
+		p.progress.WriteLine(output.Linef("", output.StylePending, "%s", ts.RepoName))
+		lines, err := diffLines(ts)
+		if err != nil {
+			p.progress.Verbosef("%-*s failed to display status: %s", p.maxRepoName, ts.RepoName, err)
+			continue
+		}
+		for _, line := range lines {
+			p.progress.Verbosef("\t%s", line)
+		}
+
```
LawnGnome, line 146: Do we want to gate this on whether verbose mode is enabled, so we don't have to calculate the diffstat otherwise?

cmd/src/campaign_progress_printer.go
```diff
@@ -134,13 +134,30 @@ func (p *campaignProgressPrinter) PrintStatuses(statuses []*campaigns.TaskStatus
...
 
-		p.progress.Verbosef("%-*s %s", p.maxRepoName, ts.RepoName, statusText)
+		verboseDiffStat, err := finalDiffStatLine(ts)
+		if err != nil {
+			p.progress.Verbosef("%-*s failed to display status: %s", p.maxRepoName, ts.RepoName, err)
+			continue
+		}
+
+		p.progress.Verbosef("  %s", verboseDiffStat)
+		p.progress.Verbose("")
```
LawnGnome, line 160: Same question as above: could we gate this on the verbose flag and not do the work if it's disabled?

cmd/src/campaign_progress_printer.go
```diff
@@ -223,3 +240,76 @@ func taskStatusText(ts *campaigns.TaskStatus) (string, error) {
...
+
+	return lines, nil
+}
+
+func finalDiffStatLine(ts *campaigns.TaskStatus) (string, error) {
+	if !ts.IsCompleted() || ts.ChangesetSpec == nil {
+		return "", nil
+	}
+
+	fileDiffs, err := diff.ParseMultiFileDiff([]byte(ts.ChangesetSpec.Commits[0].Diff))
```
LawnGnome, line 285: Can we combine `finalDiffStatLine` and `diffStats` above, so we don't have to parse and iterate twice?

LawnGnome (approved):
I think this is great!

I have a few code level suggestions that are probably already encompassed by your "clean up the code" TODO, but I think the feature is awesome.

mrnugget:
> I have a few code level suggestions that are probably already encompassed by your "clean up the code" TODO, but I think the feature is awesome.

2/3 I thought of, so I appreciate the third one!

cmd/src/campaign_progress_printer.go
mrnugget: Done!

cmd/src/campaign_progress_printer.go
mrnugget: Done!

cmd/src/campaign_progress_printer.go
mrnugget: Done!

mrnugget:
@LawnGnome @eseliger I cleaned up the code as best as I could today. We still parse the file diffs twice, if a task is completed (once for the verbose log, once for the status bar), but I couldn't make that go away without making the code hard to understand. I'd be happy to give it another shot next week. For now: please review :)

LawnGnome (approved):
> I cleaned up the code as best as I could today. We still parse the file diffs twice, if a task is completed (once for the verbose log, once for the status bar), but I couldn't make that go away without making the code hard to understand.

That's fair enough — this version's quite a bit cleaner, and I think I'm less concerned about the possible cost of parsing the diffs twice with the verbose checks up front.

Ship it!

PR #392, opened 2020-11-21
campaigns: allow env usage in campaign steps

LawnGnome:
This is the src-cli component of sourcegraph/sourcegraph#15822.

There's also a new feature flag in here to maintain backward compatibility with older Sourcegraph versions, at the cost of extra work when parsing the campaign spec. I think that's the right tradeoff.

### DIFF
--- a/internal/campaigns/campaign_spec.go
+++ b/internal/campaigns/campaign_spec.go
@@ -3,6 +3,9 @@ package campaigns
 import (
 	"fmt"
 
+	"github.com/hashicorp/go-multierror"
+	"github.com/pkg/errors"
+	"github.com/sourcegraph/campaignutils/env"
 	"github.com/sourcegraph/campaignutils/overridable"
 	"github.com/sourcegraph/campaignutils/yaml"
 	"github.com/sourcegraph/src-cli/schema"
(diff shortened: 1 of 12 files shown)


internal/campaigns/campaign_spec.go
```diff
@@ -65,18 +68,31 @@ type OnQueryOrRepository struct {
...
 	var spec CampaignSpec
 	if err := yaml.UnmarshalValidate(schema.CampaignSpecJSON, data, &spec); err != nil {
 		return nil, err
 	}
 
+	if !features.allowArrayEnvironments {
+		var errs *multierror.Error
+		for i, step := range spec.Steps {
+			if !step.Env.IsStatic() {
+				errs = multierror.Append(errs, errors.Errorf("step %d includes one or more dynamic environment variables, which are unsupported in this Sourcegraph version", i+1))
```
mrnugget, line 87: Nice touch!

internal/campaigns/execution_cache.go
```diff
@@ -24,6 +24,33 @@ type ExecutionCacheKey struct {
...
+	}
+
+	return json.Marshal(struct {
+		*Task
+		Environments []map[string]string
+	}{
+		Task:         key.Task,
+		Environments: envs,
+	})
+}
```
mrnugget, line 52: It took me a while to realise that we don't persist the the plain text version of this on disk, but take a `sha256` sum and the `base64` of it in `(ExecutionDiskCache).cacheFilePath`.

And now, with more logic in `MarshalJSON` maybe we should make the behaviour more explicit and just give `ExecutionCacheKey` a `String()` method that contains all of it — the json marshalling, the sha sum, the base64 encoding — instead of the logic being split in two methods?

mrnugget (approved):
Good stuff! Really neat :)

eseliger (approved):

chrispine:
This is awesome, thanks!

internal/campaigns/execution_cache.go
LawnGnome: Yeah, I had a similar thought. I think I'll go ahead and do that in this PR, since I agree it makes sense.

PR #393, opened 2020-11-23
Query and use the repository.branch when defined in spec

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/15070 by changing the evaluation of the `on.repository` field in the campaign spec to actually query for and use the `branch` attribute.

It queries the GraphQL API for a commit that matches the exact rev and returns an error if no commit was found.

If the campaign spec uses `repositoriesMatchingQuery` and `repository`, like this

```yaml
  - repositoriesMatchingQuery: repo:automation-testing
  - repository: github.com/sourcegraph/automation-testing
    branch: thorstens-test-branch
```

then the second branch, `thorstens-test-branch`, is checked out and used as the base branch for the changeset.

What we don't support yet is using a revspec in the search query, like `repo:github.com/sourcegraph/automation-testing@thorstens-test-branch foobar`, and using that to check out the branch. We can tackle that in a separate ticket/PR.

For now, this fixes the original problem and allows users to run the steps for a specified subset of repositories in specific branches.

(Side note: the search results in the template variable `${{ repository.search_result_paths }}` are populated by the query, even though they might not match any results in the second branch — that's undefined behaviour for now, but I want to document it in the docs.)

### DIFF
--- a/internal/campaigns/graphql/repository.go
+++ b/internal/campaigns/graphql/repository.go
@@ -22,9 +22,42 @@ fragment repositoryFields on Repository {
 }
 `
 
+const RepositoryWithBranchFragment = `
+fragment repositoryFieldsWithBranch on Repository {
+    id
+    name
+    url
+    externalRepository {
+        serviceType
+    }
+    defaultBranch {
+        name
+        target {
+            oid
+        }
+    }
+	branches(query: $branch, first: 1) @include(if:$queryBranch){
+	    nodes {
+		    name
+            target {
+                oid
+            }
+		}
+	}
+}
+`
+
+type Target struct {
+	OID string
+}
+
 type Branch struct {
 	Name   string
-	Target struct{ OID string }
+	Target Target
+}
+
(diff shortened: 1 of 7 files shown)


eseliger (approved):

chrispine:
Thanks for tackling this!

internal/campaigns/graphql/repository.go
```diff
@@ -19,29 +19,61 @@ fragment repositoryFields on Repository {
...
 `
 
+type Target struct {
+	OID string
+}
+
 type Branch struct {
 	Name   string
-	Target struct{ OID string }
+	Target Target
```
LawnGnome, line 34: Roger roger.

LawnGnome (approved):
LGTM. I appreciate the naming symmetry with types and fields in the backend where possible!

LawnGnome:
I hit Control-Enter too early, but the one thing that we do need is a changelog!

PR #395, opened 2020-11-24
Add -skip-errors flag to src campaign apply|preview

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/15790 by adding the `-skip-errors` flag.

Note: I made the conscious decision to only skip "runtime errors", i.e.: we can skip the failed step execution, or we can skip the failed importing of a changeset (see "after" screenshot). That means validation errors etc. are not skipped. Because I don't think that makes sense.

### Before
before

### After

after


eseliger (approved):

LawnGnome (approved):
LGTM!

> Note: I made the conscious decision to only skip "runtime errors", i.e.: we can skip the failed step execution, or we can skip the failed importing of a changeset (see "after" screenshot). That means validation errors etc. are not skipped. Because I don't think that makes sense.

Agreed. This is only about things the user isn't necessarily in direct control of — validation errors, server errors, etc, shouldn't be subject to this.

PR #396, opened 2020-11-25
Add stats to progress bar

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/15607.

### GIF

after

### DIFF
--- a/cmd/src/campaign_progress_printer.go
+++ b/cmd/src/campaign_progress_printer.go
@@ -55,7 +55,7 @@ func (p *campaignProgressPrinter) initProgressBar(statuses []*campaigns.TaskStat
 	}
 
 	p.progress = p.out.ProgressWithStatusBars([]output.ProgressBar{{
-		Label: fmt.Sprintf("Executing ... (  0/%d,   0 errored)", len(statuses)),
+		Label: fmt.Sprintf("Executing ... (0/%d, 0 errored)", len(statuses)),
 		Max:   float64(len(statuses)),
 	}}, statusBars, nil)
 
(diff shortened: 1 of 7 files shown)


cmd/src/campaign_progress_printer.go
```diff
@@ -55,7 +55,7 @@ func (p *campaignProgressPrinter) initProgressBar(statuses []*campaigns.TaskStat
 	}
 
 	p.progress = p.out.ProgressWithStatusBars([]output.ProgressBar{{
-		Label: fmt.Sprintf("Executing steps in %d repositories", len(statuses)),
+		Label: fmt.Sprintf("Executing ... (  0/%d,   0 errored)", len(statuses)),
```
LawnGnome, line 58: Just to avoid an initial bounce as the first 9 changesets are executed (since we use `%d` below without a width specifier), this might be better without the leading spaces:

```suggestion
		Label: fmt.Sprintf("Executing ... (0/%d, 0 errored)", len(statuses)),
```

LawnGnome (approved):
I know this is in draft, but this LGTM. A good win in terms of providing more detail without being confusing!

cmd/src/campaign_progress_printer.go
mrnugget: Hehe, this is why it's a 🙈 draft 🙈 What I was trying to do here is to add enough buffer whitespace to the `Label` so that it's not cut off later once the the numbers get bigger.

I might also add a "SetLabelAndRecalc" method to the progress bar.

cmd/src/campaigns_common.go
```diff
@@ -25,7 +25,6 @@ import (
 var (
 	campaignsPendingColor = output.StylePending
 	campaignsSuccessColor = output.StyleSuccess
-	campaignsErrorColor   = output.StyleWarning
```
eseliger, line 28: we don't have errors anymore? 🤔

eseliger (approved):

cmd/src/campaigns_common.go
mrnugget: It was unused 🤷

PR #397, opened 2020-11-25
campaigns: skip changeset spec creation for cached empty diffs

LawnGnome:
It's totally valid and normal for empty diffs to be created when
executing campaign specs: sometimes you just don't want anything to
change, even though the repo matched the initial query. When this
happens, #313 added a check that prevents the changeset spec from being
created, and print a verbose mode message indicating that the repo was
skipped:

https://sourcegraph.com/github.com/sourcegraph/src-cli@d29ad54eff678d96fb7ebdf75ff95890dce6a1cf/-/blob/internal/campaigns/executor.go?utm_source=VSCode-1.1.0#L273-278

So far, so good. In #374, we made our empty diff handling even better by
caching the empty diff: this means that we don't have to recalculate
that nothing happened. Unfortunately, the check that exists in the cache
miss code path to skip changeset spec creation doesn't exist in the
cache hit code path, which means that on subsequent applications of the
campaign, a changeset spec with an empty diff will be uploaded, and
gitserver will ultimately be very grumpy.

By applying the same logic to the cache hit code path, we can filter out
these problematic changeset specs.

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -210,9 +210,20 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 				diff = result.Commits[0].Diff
 			}
 
+			status.Cached = true
+
+			// If the cached result resulted in an empty diff, we don't need to
+			// add it to the list of specs that are displayed to the user and
+			// send to the server. Instead, we can just report that the task is
+			// complete and move on.
+			if len(diff) == 0 {
+				status.FinishedAt = time.Now()
+				x.updateTaskStatus(task, status)
+				return
+			}
+
 			spec := createChangesetSpec(task, diff, x.features)
 
-			status.Cached = true
 			status.ChangesetSpec = spec
 			status.FinishedAt = time.Now()
 			x.updateTaskStatus(task, status)
--- a/internal/campaigns/executor_test.go
+++ b/internal/campaigns/executor_test.go
@@ -4,6 +4,7 @@ import (
 	"archive/zip"
 	"bytes"
 	"context"
+	"encoding/json"
 	"fmt"
 	"io/ioutil"
 	"log"
@@ -116,6 +117,21 @@ func TestExecutor_Integration(t *testing.T) {
 				srcCLIRepo.ID: []string{"main.go", "modified-main.go.md", "added-modified-main.go.md"},
 			},
 		},
+		{
+			name:  "empty",
+			repos: []*graphql.Repository{srcCLIRepo},
+			archives: []mockRepoArchive{
+				{repo: srcCLIRepo, files: map[string]string{
+					"README.md": "# Welcome to the README\n",
+					"main.go":   "package main\n\nfunc main() {\n\tfmt.Println(     \"Hello World\")\n}\n",
+				}},
+			},
+			steps: []Step{
+				{Run: `true`, Container: "doesntmatter:13"},
+			},
+			// No changesets should be generated.
+			wantFilesChanged: map[string][]string{},
+		},
 	}
 
 	for _, tc := range tests {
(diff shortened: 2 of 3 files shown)


internal/campaigns/executor.go
```diff
@@ -210,9 +210,20 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
...
 			}
 
+			status.Cached = true
+
+			// If the cached result resulted in an empty diff, we don't need to
+			// add it to the list of specs that are displayed to the user and
+			// send to the server. Instead, we can just report that the task is
+			// complete and move on.
+			if len(diff) == 0 {
+				status.FinishedAt = time.Now()
```
eseliger, line 220: I think this line could also be moved up and deduped

eseliger (approved):
Good catch. LGTM!

internal/campaigns/executor.go
LawnGnome: I thought about it, but kept it there because `createChangesetSpec()` really marks the end of executing the task in the normal case, and I'd like to account for that time as well.

LawnGnome:
Just asking for a quick re-review here to glance over the test changes, since they're moderately significant.

internal/campaigns/executor_test.go
```diff
@@ -116,6 +118,21 @@ func TestExecutor_Integration(t *testing.T) {
...
+				{repo: srcCLIRepo, files: map[string]string{
+					"README.md": "# Welcome to the README\n",
+					"main.go":   "package main\n\nfunc main() {\n\tfmt.Println(     \"Hello World\")\n}\n",
+				}},
+			},
+			steps: []Step{
+				{Run: `true`, Container: "doesntmatter:13"},
+			},
+			// No changesets should be generated.
+			wantFilesChanged: map[string][]string{},
```
eseliger, line 134: I don't 100% understand how this tests the behavior is right but I trust you that it does what we look for. Regardless this is a good test. 
I assume this would have failed before the fix because the hot run is going to yield a change? Even though it was empty?  Not sure why it would show any files changed when there was an empty diff.

eseliger (approved):
Great work, I like that. Just a question re the test preventing this from happening again.

internal/campaigns/executor_test.go
LawnGnome: Each entry in `wantFilesChanged` represents a changeset spec. In this case, since `true` won't change anything, the test _should_ hit the empty diff code paths and not generate changeset specs, even though the cache is populated.

This does indeed fail on the old `main`:

```

~/trees/sourcegraph/src-cli on  main via 🐹 v1.15.5 ❯ go test -race ./...
ok  	github.com/sourcegraph/src-cli/cmd/src	0.224s
ok  	github.com/sourcegraph/src-cli/internal/api	(cached)
--- FAIL: TestExecutor_Integration (0.46s)
    --- FAIL: TestExecutor_Integration/empty (0.04s)
        executor_test.go:189: wrong number of changeset specs. want=0, have=1
        --- FAIL: TestExecutor_Integration/empty/warm_cache (0.00s)
            testing.go:1048: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test
FAIL
FAIL	github.com/sourcegraph/src-cli/internal/campaigns	0.565s
?   	github.com/sourcegraph/src-cli/internal/campaigns/graphql	[no test files]
ok  	github.com/sourcegraph/src-cli/internal/codeintel	(cached)
?   	github.com/sourcegraph/src-cli/internal/output	[no test files]
ok  	github.com/sourcegraph/src-cli/internal/servegit	(cached)
?   	github.com/sourcegraph/src-cli/schema	[no test files]
FAIL
```

eseliger: Ahh okay makes sense. Thanks for confirming!

mrnugget:
Ah man, what a bug! Thanks for fixing this so diligently!

PR #398, opened 2020-11-30
Introduce `transformChanges` to campaign spec (RFC 265)

mrnugget:
This is the implementation of the suggested solution for **Option B** in [RFC 265: Multiple changesets per repository](https://docs.google.com/document/d/1cXEuUdYJeK-uwZpowhorPGSAUfE87QkaLbf9r_EVRfI/edit) to produce multiple changesets in a single repository.

It depends on https://github.com/sourcegraph/sourcegraph/pull/16235. It fixes https://github.com/sourcegraph/sourcegraph/issues/14970.

**IMPORTANT**: this _only_ implements the first step in ["Defintion of success"](https://docs.google.com/document/d/1cXEuUdYJeK-uwZpowhorPGSAUfE87QkaLbf9r_EVRfI/edit#bookmark=id.rs92im1omcfo). It does **not** yet implement addressing multiple changesets in the same repository separately (that's Part2 in the RFC).

What this does is to introduce the `transformChanges` property to campaign specs (see https://github.com/sourcegraph/sourcegraph/pull/16235 for schema change on server).

Here is an example campaign spec using it:

```yaml
name: multi-changesets
description: This is a testing campaign that should have multiple changesets in a single repository.

on:
  - repositoriesMatchingQuery: repo:src-cli|automation-testing$

steps:
  - run: echo ".swp" >> .gitignore
    container: alpine:3
  - run: if [[ -e gopkg/main.go ]]; then echo "func unused() int { return 99 }" >> gopkg/main.go; fi
    container: alpine:3
  - run: mkdir foobar && echo "Hey there" foobar/hello.txt
    container: alpine:3

# Transform the changes produced in each repository.
transformChanges:
  # Group the file diffs by directory and produce one additional changeset per group.
  group:
    - directory: gopkg
      branch: thorsten/multi-changesets-backend # will replace the `branch` in the `changesetTemplate`
    - directory: foobar
      repository: github.com/sourcegraph/automation-testing # optional: only apply the rule in this repository
      branch: thorsten/multi-changesets-foobar # will replace the `branch` in the `changesetTemplate`

changesetTemplate:
  title: multi-changesets
  body: multi-changesets
  branch: thorsten/multi-changesets
  commit:
    message: multi-changesets
  published: false
```

#### Changelog of this PR 

- 02 Dec: Updated the code and the PR description to use the new `branch` property instead of `branchSuffix` to incorporate feedback on RFC 265.
- 02 Dec: Added the optional `repository` attribute to the `group`.
- 10 Dec: Added validation to make sure that multiple changesets cannot be created with the _same_ branch in the _same_ repository. Updated PR description.

### DIFF
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -264,12 +264,17 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 	ids := make([]campaigns.ChangesetSpecID, len(specs))
 
 	if len(specs) > 0 {
+		var label string
+		if len(specs) == 1 {
+			label = "Sending changeset spec"
+		} else {
+			label = fmt.Sprintf("Sending %d changeset specs", len(specs))
+		}
+
 		progress := out.Progress([]output.ProgressBar{
-			{
-				Label: fmt.Sprintf("Sending %d changeset specs", len(specs)),
-				Max:   float64(len(specs)),
-			},
+			{Label: label, Max: float64(len(specs))},
 		}, nil)
+
 		for i, spec := range specs {
 			id, err := svc.CreateChangesetSpec(ctx, spec)
 			if err != nil {
@@ -279,7 +284,6 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 			progress.SetValue(0, float64(i+1))
 		}
 		progress.Complete()
-
 	} else {
 		if len(repos) == 0 {
 			out.WriteLine(output.Linef(output.EmojiWarning, output.StyleWarning, `No changeset specs created`))
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -265,7 +265,10 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 
 	if len(specs) > 0 {
 		progress := out.Progress([]output.ProgressBar{
-			{Label: "Sending changeset specs", Max: float64(len(specs))},
+			{
+				Label: fmt.Sprintf("Sending %d changeset specs", len(specs)),
+				Max:   float64(len(specs)),
+			},
 		}, nil)
 		for i, spec := range specs {
 			id, err := svc.CreateChangesetSpec(ctx, spec)
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -216,25 +216,17 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 			return
 		}
 	} else {
-		var result *ChangesetSpec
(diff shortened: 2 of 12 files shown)


mrnugget:
Just updated this to use the `branch` attribute in a `group` instead of `branchSuffix` based on the feedback in RFC 265.

mrnugget:
@sourcegraph/campaigns @rrhyne This is now ready for review. Please see the PR description for what's included and what's not. Documentation and schema update in `sourcegraph/sourcegraph` are here: https://github.com/sourcegraph/sourcegraph/pull/16235

cmd/src/campaigns_common.go
```diff
@@ -265,7 +265,10 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 
 	if len(specs) > 0 {
 		progress := out.Progress([]output.ProgressBar{
-			{Label: "Sending changeset specs", Max: float64(len(specs))},
+			{
+				Label: fmt.Sprintf("Sending %d changeset specs", len(specs)),
```
LawnGnome, line 269: This should probably have a special case to handle `len(specs) == 1` so we don't output `Sending 1 changeset specs`.

internal/campaigns/executor.go
```diff
@@ -97,6 +108,34 @@ func (ts *TaskStatus) ExecutionTime() time.Duration {
...
+// FileDiffs returns the file diffs produced by the Task in the given
+// repository.
+// If no file diffs were produced, the task resulted in an error, or the task
+// hasn't finished execution yet, the second return value is false.
+func (ts *TaskStatus) FileDiffs() ([]*diff.FileDiff, bool, error) {
+	if !ts.IsCompleted() || len(ts.ChangesetSpecs) == 0 || ts.Err != nil {
+		return nil, false, nil
+	}
+
+	ts.fileDiffsOnce.Do(func() {
```
LawnGnome, line 120: Nice bit of memoisation here.

internal/campaigns/executor.go
```diff
@@ -374,24 +412,149 @@ func createChangesetSpec(task *Task, diff string, features featureFlags) *Change
...
+
+	// For each file diff...
+	for _, f := range fileDiffs {
+		name := f.NewName
+		if name == "/dev/null" {
+			name = f.OrigName
+		}
+
+		// .. we check whether it matches one of the given directories in the
+		// group transformations, starting with the longest one first:
```
LawnGnome, line 527: I think this comment is now inaccurate after the RFC change to use the last match.

```suggestion
		// group transformations, with the last match winning:
```

LawnGnome (approved):
LGTM

chrispine (approved):
👍🏻

cmd/src/campaigns_common.go
mrnugget: Ah, good catch!

PR #399, opened 2020-11-30
Fix race conditions when printing current status of tasks

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/16195 by doing two things:

1. It threads every update and read of `*TaskStatus` in `executor` through a mutex to provide a consistent view of all `[]*TaskStatus`. While we had a `sync.Map` in place it didn't protect us from race conditions, since we passed around _pointers_ to `TaskStatus` freely: the `executor` goroutines would happily write to a `*TaskStatus` and the `campaignProgressPrinter.PrintStatus` would read `[]*TaskStatus` _while they're being updated_. The `sync.Map` was useless in practice. So instead of that, I added a mutex that's used when updating a `*TaskStatus` and when reading _all_ `[]*TaskStatus` when printing the status.
2. It uses a 1-capacity semaphore to control access to the `campaignProgressPrinter` through the `PrintStatuses` callback: only one execution at a time and if another execution is already in progress it returns.

### DIFF
--- a/internal/campaigns/executor.go
+++ b/internal/campaigns/executor.go
@@ -40,10 +40,16 @@ func (e TaskExecutionErr) StatusText() string {
 }
 
 type Executor interface {
-	AddTask(repo *graphql.Repository, steps []Step, template *ChangesetTemplate) *TaskStatus
+	AddTask(repo *graphql.Repository, steps []Step, template *ChangesetTemplate)
 	LogFiles() []string
 	Start(ctx context.Context)
 	Wait() ([]*ChangesetSpec, error)
+
+	// LockedTaskStatuses calls the given function with the current state of
+	// the task statuses. Before calling the function, the statuses are locked
+	// to provide a consistent view of all statuses, but that also means the
+	// callback should be as fast as possible.
+	LockedTaskStatuses(func([]*TaskStatus))
 }
 
 type Task struct {
@@ -74,6 +80,11 @@ type TaskStatus struct {
 	Err           error
 }
 
+func (ts *TaskStatus) Clone() *TaskStatus {
+	clone := *ts
+	return &clone
+}
+
 func (ts *TaskStatus) IsRunning() bool {
 	return !ts.StartedAt.IsZero() && ts.FinishedAt.IsZero()
 }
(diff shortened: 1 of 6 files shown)


internal/campaigns/executor.go
```diff
@@ -302,8 +319,26 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
...
+}
+
+func (x *executor) LockedTaskStatuses(callback func([]*TaskStatus)) {
+	x.statusesMu.RLock()
+	defer x.statusesMu.RUnlock()
+
+	var s []*TaskStatus
+	for _, status := range x.statuses {
+		s = append(s, status)
+	}
```
eseliger, line 339: why the copy of a pointer slice?

internal/campaigns/executor.go
```diff
@@ -168,24 +185,19 @@ func (x *executor) Wait() ([]*ChangesetSpec, error) {
...
 	defer func() {
-		status.FinishedAt = time.Now()
-		status.CurrentlyExecuting = ""
-		status.Err = err
-		x.updateTaskStatus(task, status)
+		x.updateTaskStatus(task, func(status *TaskStatus) {
+			status.FinishedAt = time.Now()
+			status.CurrentlyExecuting = ""
+			status.Err = err
+		})
```
eseliger, line 194: I wish there was a readonly type so we could never forget this 😢

eseliger (approved):
Code LGTM, I trust you that it fixes the underlying issue :)

internal/campaigns/executor.go
mrnugget: Because `x.statuses` is a map and not a slice and I don't want to expose the `*Task` (key of the map) to the consumers.

eseliger: Ah, I didn't see that. Makes sense

LawnGnome (approved):
LGTM. Thanks for fixing this bug! (I'll spare us all the Rust evangelism here.)

mrnugget:
> LGTM. Thanks for fixing this bug! (I'll spare us all the Rust evangelism here.)

I know :| But to be fair to Go, what we're doing here is communicating between two parts by sharing memory, instead of [sharing by communicating](https://blog.golang.org/codelab-share). (This is orthogonal to Rust's memory checks that would've probably avoided this, but a thought I had yesterday)

PR #400, opened 2020-12-01
Snapshot test the terminal UI

mrnugget:
I started with one thought: "I just wish I had _one_ test to make sure that this whole thing [the terminal UI] doesn't blow up every time I make a change" What I ended with is this. It includes a tiny VT100 "emulator".

And here's the kicker: I actually like this. It's fast to run and easy to extend and it gives me confidence that the basic rendering functionality works. Of course, the hacky injections need to be cleaned up before merging. And maybe we can get it to pass on Windows.

Opinions?

(This fixes https://github.com/sourcegraph/sourcegraph/issues/16330)

### DIFF
--- a/cmd/src/campaign_progress_printer_test.go
+++ b/cmd/src/campaign_progress_printer_test.go
@@ -81,6 +81,7 @@ func TestCampaignProgressPrinterIntegration(t *testing.T) {
 		"├── github.com/sourcegraph/sourcegraph  echo Hello World > README.md          0s",
 		"├── github.com/sourcegraph/src-cli      Downloading archive                   0s",
 		"└── github.com/sourcegraph/automati...  echo Hello World > README.md          0s",
+		"",
 	}
 	if !cmp.Equal(want, have) {
 		t.Fatalf("wrong output:\n%s", cmp.Diff(want, have))
@@ -128,6 +129,7 @@ func TestCampaignProgressPrinterIntegration(t *testing.T) {
 		"├── github.com/sourcegraph/sourcegraph  echo Hello World > README.md          0s",
 		"├── github.com/sourcegraph/src-cli      Downloading archive                   0s",
 		"└── github.com/sourcegraph/automati...  3 files changed ++++               0s",
+		"",
 	}
 	if !cmp.Equal(want, have) {
 		t.Fatalf("wrong output:\n%s", cmp.Diff(want, have))
@@ -157,6 +159,11 @@ func (t *ttyBuf) Write(b []byte) (int, error) {
 		case '\n':
 			t.line++
 			t.column = 0
+
+			if len(t.lines) == t.line {
+				t.lines = append(t.lines, []byte{})
+			}
+
 		case '\x1b':
 			// First of all: forgive me.
 			//
--- a/cmd/src/campaign_progress_printer_test.go
+++ b/cmd/src/campaign_progress_printer_test.go
@@ -1,0 +1,151 @@
+package main
+
+import (
+	"strconv"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/sourcegraph/src-cli/internal/campaigns"
+	"github.com/sourcegraph/src-cli/internal/output"
+)
+
+func TestCampaignProgressPrinterIntegration(t *testing.T) {
+	buf := &ttyBuf{}
+
+	out := output.NewOutput(buf, output.OutputOpts{
+		ForceTTY:   true,
+		ForceColor: true,
+		Verbose:    true,
+	})
+
+	printer := newCampaignProgressPrinter(out, true, 4)
+	printer.forceNoSpinner = true
+
+	printer.PrintStatuses([]*campaigns.TaskStatus{
+		{
+			RepoName:           "github.com/sourcegraph/sourcegraph",
+			StartedAt:          time.Now(),
+			CurrentlyExecuting: "echo Hello World > README.md",
+		},
+		{
+			RepoName:           "github.com/sourcegraph/src-cli",
+			StartedAt:          time.Now().Add(time.Duration(-5) * time.Second),
+			CurrentlyExecuting: "Downloading archive",
+		},
+	})
+
+	have := buf.Lines()
+	want := []string{
+		"⠋  Executing... (0/2, 0 errored)  ",
+		"│                                                                                                                     ",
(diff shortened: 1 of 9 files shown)


cmd/src/campaign_progress_printer_test.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+type ttyBuf struct {
+	lines [][]byte
+
+	line   int
+	column int
+}
+
+func (t *ttyBuf) Write(b []byte) (int, error) {
```
keegancsmith, line 66: nice!

mrnugget:
btw. I know that "test coverage" itself doesn't mean a lot, but with the latest commit this bumps the test coverage for `campaign_progress_printer.go` from 0% to 80%. That's pretty cool!

cmd/src/campaign_progress_printer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	var cur int
+
+	for cur < len(b) {
+		switch b[cur] {
+		case '\n':
+			t.line++
+			t.column = 0
+		case '\x1b':
+			// First of all: forgive me.
```
LawnGnome, line 161: Ten Hail ANSIs should do it.

cmd/src/campaign_progress_printer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+	if !cmp.Equal(want, have) {
+		t.Fatalf("wrong output:\n%s", cmp.Diff(want, have))
+	}
+
+	// Now mark the last task as completed
+	statuses[len(statuses)-1] = &campaigns.TaskStatus{
+		RepoName:           "github.com/sourcegraph/automation-testing",
+		StartedAt:          time.Now().Add(time.Duration(-5) * time.Second),
+		FinishedAt:         time.Now().Add(time.Duration(5) * time.Second),
```
LawnGnome, line 93: This could be very subtly flaky if (somehow) the `StartedAt` and `FinishedAt` calls are separated by more than half a second. It might be better to calculate this above this struct as something like:

```go
startedAt := time.Now().Add(time.Duration(-5) * time.Second)
finishedAt := startedAt.Add(time.Duration(10) * time.Second)
```

Of course, I might also just be paranoid.

internal/output/output.go
```diff
@@ -50,7 +50,15 @@ type Output struct {
...
 	// ForceColor ignores all terminal detection and enabled coloured output.
 	ForceColor bool
-	Verbose    bool
+	// ForceTTY ignores all terminal detection and enables TTY output.
+	ForceTTY bool
+
+	// ForceHeight ignores all terminal detection and sets the height to this value.
+	ForceHeight int
+	// ForceWidth ignores all terminal detection and sets the width to this value.
+	ForceWidth int
```
LawnGnome, line 59: Obviously don't do it now, but these _might_ actually be something we'd want to expose at some point for more than just testing.

cmd/src/campaign_progress_printer_test.go
```diff
@@ -0,0 +1,249 @@
...
+--- x/x.txt
++++ x/x.txt
+@@ -1 +1 @@
+-this is x
++this is x (or is it?)
+`
+
+func TestCampaignProgressPrinterIntegration(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Something emits different escape codes on windows.")
```
LawnGnome, line 41: Can you create a follow up issue for us to figure this out at some point, please? I don't really want to leave this skipped forever, and there _should_ be a way to use the same approach on Windows.

LawnGnome (approved):
It's hideous. I love it. :heart:

cmd/src/campaign_progress_printer_test.go
```diff
@@ -0,0 +1,249 @@
...
+			// First of all: forgive me.
+			//
+			// Now. Looks like we ran into a VT100 escape code.
+			// They follow this structure:
+			//
+			//      \x1b [ <digit> <command>
+			//
+			// So we jump over the \x1b[ and try to parse the digit.
+
+			cur = cur + 2 // cur == '\x1b', cur + 1 == '['
```
chrispine, line 170: Do you want to verify that `cur+1 == '['`?

cmd/src/campaign_progress_printer_test.go
```diff
@@ -0,0 +1,249 @@
...
+	want = []string{
+		"github.com/sourcegraph/automation-testing",
+		"\tREADME.md   | 3 +++",
+		"\ta/b/c/c.txt | 1 -",
+		"\tx/x.txt     | 2 +-",
+		"  3 files changed, 4 insertions, 2 deletions",
+		"  Execution took 10s",
+		"",
+		"⠋  Executing... (1/3, 0 errored)  ███████████████▍",
+		"│                                                                                 0s", // Not sure why this is here, bug?
```
chrispine, line 127: Did you ever figure out what this was there? (Just so I'm clear: the bug is the extra `0s` that doesn't appear to correspond to any task?)

chrispine (approved):
Very cool!

internal/output/output.go
mrnugget: Yeah, had the same thought!

cmd/src/campaign_progress_printer_test.go
mrnugget: Turns out it was a bug in our code! See: 550dafa I never cleared the current line when printing the `|        ` blank line. And the `ttyBuf` implementation keeps all previously written things in memory, so this was a leftover. The reason why this wasn't visible in the actual TUI I'm not sure, but I suspect it's because the `ttyBuf` handles lines differently

cmd/src/campaign_progress_printer_test.go
mrnugget: Done!

cmd/src/campaign_progress_printer_test.go
mrnugget: Done! Yeah, I also think so. https://github.com/sourcegraph/sourcegraph/issues/16353

cmd/src/campaign_progress_printer_test.go
mrnugget: Not paranoid! Good point :)

cmd/src/campaign_progress_printer_test.go
chrispine: Cool, thanks!

PR #403, opened 2020-12-11
Actually skip unsupported repos if -allow-unsupported is not set

mrnugget:
I think multiple changes to this method lead to a regression where
unsupported repositories were correctly being reported as unsupported
but were still added to the final list of repositories, regardless of
whether `-allow-unsupported` was set or not.


eseliger (approved):
Nice finally a test for this!

PR #409, opened 2020-12-15
extensions: copy operation create client pointing at dot com

uwedeportivo:
withCfg swaps in a new config that points at dot-com but the request uses the client created above the swap and captured in the closure.


eseliger (approved):
Oof. Good catch! I bet that was awful to debug, sorry about that.

LawnGnome (approved):

PR #410, opened 2020-12-15
campaigns: suggest alternatives when campaigns are unlicensed

LawnGnome:
In and around sourcegraph/sourcegraph#16687, we had a fairly lengthy discussion around how to present licensing errors to the user. Doing it in a way that is actionable for the user is fairly tricky: we don't have a simple limit directive that could be applied to the repository search, so the user is left with a campaign spec that they might still want to trial but don't know how to get to a trialable state.

This PR uses the new error code returned from sourcegraph/sourcegraph#16687 to special case licensing errors. When those are detected, a suggestion is made for how the campaign spec could be modified to be compliant with the unlicensed behaviour.

Here's how it looks:

image

Linked to sourcegraph/sourcegraph#15715, but it's more a bonus adventure than that issue.

### DIFF
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -436,10 +436,16 @@ func prettyPrintCampaignsUnlicensedError(out *output.Output, err error) error {
 			} else if code == "ErrCampaignsUnlicensed" {
 				// OK, let's print a better message, then return an
 				// exitCodeError to suppress the normal automatic error block.
-				block := out.Block(output.Line("🪙", output.StyleWarning, "Campaigns is a paid feature of Sourcegraph. All users can create sample campaigns with up to 5 changesets without a license."))
-				block.WriteLine(output.Linef("", output.StyleWarning, "Contact Sourcegraph sales at %shttps://about.sourcegraph.com/contact/sales/%s to obtain a trial license.", output.StyleSearchLink, output.StyleWarning))
+				// Note that we have hand wrapped the output at 80 (printable)
+				// characters: having automatic wrapping some day would be nice,
+				// but this should be sufficient for now.
+				block := out.Block(output.Line("🪙", output.StyleWarning, "Campaigns is a paid feature of Sourcegraph. All users can create sample"))
+				block.WriteLine(output.Linef("", output.StyleWarning, "campaigns with up to 5 changesets without a license. Contact Sourcegraph sales"))
+				block.WriteLine(output.Linef("", output.StyleWarning, "at %shttps://about.sourcegraph.com/contact/sales/%s to obtain a trial license.", output.StyleSearchLink, output.StyleWarning))
 				block.Write("")
-				block.WriteLine(output.Linef("", output.StyleWarning, "To proceed with this campaign, you will need to create five or fewer changesets. To do so, you could try adding %scount:5%s to your %srepositoriesMatchingQuery%s search, or reduce the number of changesets in %simportChangesets%s.", output.StyleSearchAlertProposedQuery, output.StyleWarning, output.StyleReset, output.StyleWarning, output.StyleReset, output.StyleWarning))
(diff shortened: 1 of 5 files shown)


cmd/src/campaigns_common.go
```diff
@@ -331,6 +330,12 @@ func campaignsParseSpec(out *output.Output, svc *campaigns.Service, input io.Rea
 // printExecutionError is used to print the possible error returned by
 // campaignsExecute.
 func printExecutionError(out *output.Output, err error) {
+	// exitCodeError shouldn't generate any specific output, since it indicates
+	// that this was done deeper in the call stack.
+	if _, ok := err.(*exitCodeError); ok {
```
mrnugget, line 335: Hmm, are you sure about this? I'm asking because I vaguely remember something related to killed-by-interrupt and `exitCodeError` but I am not sure. If you're sure, though, and this still works when and reports errors correctly when killing it with Ctrl-C, then that's okay.

cmd/src/campaigns_suggestions.go
```diff
@@ -0,0 +1,89 @@
...
+	return err
+}
+
+func outputUnlicensedError(ctx context.Context, out *output.Output, svc *campaigns.Service, spec *campaigns.CampaignSpec) error {
+	// OK, let's print a better message, then return an exitCodeError to
+	// suppress the normal automatic error block.
+	//
+	// First, we're going to print a generic message about the restrictions on
+	// deployments without a campaigns license.
+	block := out.Block(output.Line("🪙", output.StyleWarning, "Campaigns is a paid feature of Sourcegraph. All users can create sample campaigns with up to five changesets without a license."))
```
mrnugget, line 47:
```suggestion
	block := out.Block(output.Line("🪙", output.StyleWarning, "Campaigns is a paid feature of Sourcegraph. All users can create sample campaigns with up to 5 changesets without a license."))
```

Usually I wouldn't recommend this (I remember a rule saying that every number up to 12 should be written out), but in this case I think the number should stand out.

cmd/src/campaigns_suggestions.go
```diff
@@ -0,0 +1,89 @@
+package main
+
+import (
+	"context"
+
+	"github.com/sourcegraph/src-cli/internal/api"
+	"github.com/sourcegraph/src-cli/internal/campaigns"
+	"github.com/sourcegraph/src-cli/internal/output"
+	"gopkg.in/yaml.v2"
```
mrnugget, line 9: I think if you use `v3` then the `"on":` would be printed as `on:` instead.

mrnugget (approved):
Code looks good to me and I do appreciate the work you put into this, which seems like a lot, but I'm also skeptical of whether we should do this.

It adds a lot of complexity (errors being wrapped, errors being printed differently, multiple ways to get suggestions, additional requests being made) that I'm not sure is worth it, because in the end the suggestions help to make the error go away, but they might not be what the user is after. And more often than not, adding a `count:x` fixes the problem. Or using a `repo:` filter, or a `lang:` filter, all of which is already suggested in our web UI to drill down on search results.

cmd/src/campaigns_common.go
LawnGnome: In response to this, I tested that specifically, and it does still report errors as expected with Ctrl-C.

cmd/src/campaigns_suggestions.go
LawnGnome: Implemented in the rebase.

cmd/src/campaigns_suggestions.go
LawnGnome: Now moot.

LawnGnome:
This has been refactored into the simpler version that just gives you a couple of static hints, as discussed at today's sync, and I've dismissed the existing review, since the PR has changed a fair bit.

eseliger (approved):
🌟

PR #411, opened 2020-12-16
Use api.Client for http request in src version

eseliger:
This fixes us not using any authentication in the src version command, because the api.Client adds the Authorization header to the request automatically


LawnGnome (approved):
This is probably worth a changelog entry, but otherwise this LGTM.

chrispine (approved):
LGTM, thanks for the quick turnaround!

PR #412, opened 2020-12-25
campaigns: add and use volume mounts by default on Intel macOS

LawnGnome:
### Background

As we now know all too well, the default bind mount behaviour we use to manage workspaces when executing campaigns is slow on macOS: any file I/O in the workspace has to go through the loopback interface, and this causes problems with I/O heavy operations such as pretty much anything to do with `npm` or `yarn`.

The steps we perform within a workspace are extremely simple, however: we unzip the repository archive and run a variety of `git` commands to generate the diff. It's handy for this to be done on the host filesystem, since it allows the user to inspect what happens to their repository if a campaign step fails, but it's not a technical requirement.

### What's in the ~box~ PR

This PR generalises our existing concept of a workspace, and expands it significantly. Workspace modes are now defined by implementing a `WorkspaceCreator` interface, which knows how to create `Workspace` implementations which implement the common operations we need to set up, diff, and tear down workspaces. This is used to add a new workspace mode that uses a Docker volume to contain the workspace, rather than bind mounting the workspace from the host. This is controlled by the new `-workspace` flag, and the default for Intel macOS has been switched to the new volume mode.

For this to work, we have to perform the workspace management commands — the aforementioned unzipping and `git` malarkey — within a container, since that's the only way to access the volume. As such, part of this PR adds a Docker image that extends `alpine` to always have `git` available and configured appropriately, along with `unzip` and `curl`. A GitHub Action has been added that will rebuild and push this image on `src` release.

There's also a _lot_ of new testing stuff: the volume workspace code is copiously unit tested. In order to support this, there's a (very) minimal implementation of a mock API client, and an expanded ~ripoff~ implementation of the method Go's `os/exec` test suite uses to mock external commands, which means the new tests don't need `dummydocker` and work on Windows.

#### Why is CI failing?

The `os/exec` method for mocking external commands basically redirects `exec.Cmd` calls back into the test binary, which gains a little bit of logic in its `TestMain` to handle those correctly.

This works fine cross-platform, but something about AppVeyor specifically makes this not work on Windows. The errors are fairly inscrutable. However, the approach doesn't have any systemic issue on Windows: it works fine both for me locally, and in GitHub Actions, which now has Windows runner support.

I've opened #415, which is included in this PR as well, to remove our AppVeyor support and only use GitHub Actions for CI. If we decide to go ahead with that, then we can remove the AppVeyor integration and this PR will start passing.

#### How much faster is this on macOS?

It depends.

For campaigns that generate small-to-medium sized diffs and do little I/O, there isn't much difference: maybe a few percent here and there. For campaigns that generate large diffs and do little I/O, this can actually be slower. (That feels like a _really_ weird case, though.)

For campaigns that perform lots of I/O, this is a big win. A test campaign that upgrades TypeScript in Sourcegraph completes in approximately a quarter of the time in volume mode compared to bind mode. (~11 minutes compared to ~42.)

#### Why change the default on macOS only?

Volume mode isn't universally better. On Linux (assuming nothing weird like a remote Docker server), there's no reason not to use bind mount mode: the performance is the same, and you have the advantage that you can inspect what happened in your workspace if a step fails.

#### Why change the default on _Intel_ macOS only?

Conservatism. I've put in a bunch of effort to make multi-architecture builds of the Docker image needed by volume mode work, but I don't have an M1 Mac sitting around to test this. It _should_ work, but I'd prefer to find that out by testing it using `-workspace volume` explicitly, rather than finding out that it doesn't work after making it the default.

#### What about Windows?

My suspicion is that this will provide similar performance improvements on Windows, but I don't have good numbers on this right now, don't have an environment ready to go to prove that out, and I'd prefer to focus on macOS for now. It's easy enough to change the default later.

That said, I think there's another reason we should consider this for Windows sooner rather than later: it effectively removes the requirement to have `git` in your PATH, and removes any potential for Windows-specific `git` weirdness.

#### What future work could we do to improve this further?

Glad you asked! I had more ideas, but this PR was more than big enough already.

1. Bind mode tests: now that we have comprehensive tests for volume workspaces, we could retrofit the same techniques onto the bind workspace tests.
2. Volume mode debugging: right now, if a step fails, the volume is removed, which means the user would have to do some hackery using something like `sleep` to `exec` into the container to inspect the workspace if that was necessary. We should probably provide some sort of interactive debugging mode where you get dropped into a container with the workspace already set up and running on error. (I mean, this is probably a good idea for bind mode, too.)
3. Volume container communication: I have a suspicion that the large diff slowness noted above might be more to do with extracting the diff from the Docker container via `stdout` than anything endemic to the approach. Another way of doing this would be to have the utility container running the whole time while executing a campaign on a repository, and provide some sort of RPC interface to perform setup and teardown and (most importantly) get diffs without having to go through Docker.
4. Exec output: to allow mocking external commands, the volume workspace uses the new `internal/exec` package instead of `os/exec` directly. If we migrate other modules to use this, then we could use that down the track to provide verbose logging of every command that's run, since command execution will go through a central place.

### PR links

Skeletal end user documentation is provided by sourcegraph/sourcegraph#16979.

Fixes sourcegraph/sourcegraph#16809.


internal/campaigns/bind_workspace_test.go
```diff
@@ -54,7 +55,7 @@ func TestWorkspaceCreator_Create(t *testing.T) {
 
 		testTempDir := workspaceTmpDir(t)
 
-		creator := &WorkspaceCreator{dir: testTempDir, client: client}
+		creator := &dockerBindWorkspaceCreator{dir: testTempDir, client: client}
```
mrnugget, line 58: I can't comment on it, but I think the function should be renamed from `func TestWorkspaceCreator_Create` to `func TestDockerBindWorkspaceCreator_Create`

internal/campaigns/bind_workspace_test.go
```diff
@@ -329,10 +330,15 @@ func isDir(t *testing.T, path string) bool {
...
+func readWorkspaceFiles(workspace Workspace) (map[string]string, error) {
+	w, ok := workspace.(*dockerBindWorkspace)
+	if !ok {
+		return nil, errors.Errorf("unknown workspace of type %T: %+v", workspace, workspace)
+	}
+
 	files := map[string]string{}
 
-	err := filepath.Walk(workspace, func(path string, info os.FileInfo, err error) error {
+	err := filepath.Walk(w.dir, func(path string, info os.FileInfo, err error) error {
```
mrnugget, line 341: Do you even need that assertion? I think with the interface in place you could simply call `w.WorkDir()` and walk that, right?

(I'd also understand if you want to assert the type here, so feel free to ignore this comment)

internal/campaigns/run_steps.go
```diff
@@ -478,7 +436,7 @@ func (stepCtx *StepContext) ToFuncMap() template.FuncMap {
 // StepResult represents the result of a previously executed step.
 type StepResult struct {
 	// Files are the changes made to files by the step.
-	Files StepChanges
+	Files *StepChanges
```
mrnugget, line 439: If we make this a pointer I think we should also lower-case `Files` to `files` so that users can't use it with the templating functionality. The original reason why I didn't make it a pointer was because those can lead to nil-pointer panics when playing around with templating, but after that we introduced the helper methods, so I think we don't need to to expose `Files` anymore, but best to check in the docs.

internal/campaigns/service.go
```diff
@@ -199,19 +202,68 @@ func (svc *Service) NewExecutor(opts ExecutorOpts) Executor {
...
+		}
+	} else {
+		// Either append the digest pointer to an existing map entry, or add a
+		// new entry if required.
+		if digests, ok := dis[image]; ok {
+			dis[image] = append(digests, digestPtr)
+		} else {
+			dis[image] = []*string{digestPtr}
+		}
+	}
```
mrnugget, line 232: I get the feeling this was written by someone who's used to working with pointers 😄

internal/campaigns/workspace.go
```diff
@@ -0,0 +1,94 @@
...
+	Create(context.Context, *graphql.Repository) (Workspace, error)
+
+	// DockerImages returns any Docker images required to use workspaces created
+	// by this creator.
+	DockerImages() []string
+}
+
+// Workspace implementations manage per-changeset storage when executing
+// campaign step.
+type Workspace interface {
```
mrnugget, line 30: This is a really neat abstraction 👏 I've grown to be skeptical when I see interfaces (maybe even: abstractions in general? am I getting old?), but I think this one here falls in place just right.

internal/campaigns/workspace.go
```diff
@@ -0,0 +1,94 @@
...
+	// Prepare is called once, immediately before the first step is executed.
+	// Generally, this would perform whatever Git or other VCS setup is required
+	// to establish a base upon which to calculate changes later.
+	Prepare(ctx context.Context) error
```
mrnugget, line 34: This is the only one I'm unsure about: since we do have a `WorkspaceCreator` and _specific implementations_: is there a reason why we shouldn't the things that are being done in `Prepare` into the `Create` method of the corresponding workspace creators?

I think that could be nice, because it would mean that (1) `WorkspaceCreator` creates a ready-to-go workspace and (2) the `Workspace` interface becomes smaller and easier to use (you don't have to manipulate a workspace before you can use it)

internal/campaigns/volume_workspace.go
```diff
@@ -0,0 +1,199 @@
...
+	"os"
+
+	"github.com/pkg/errors"
+
+	"github.com/sourcegraph/src-cli/internal/api"
+	"github.com/sourcegraph/src-cli/internal/campaigns/graphql"
+	"github.com/sourcegraph/src-cli/internal/exec"
+)
+
+// dockerVolumeWorkspaceCreator creates dockerWorkspace instances.
```
mrnugget, line 16:
```suggestion
// dockerVolumeWorkspaceCreator creates dockerVolumeWorkspace instances.
```

internal/campaigns/volume_workspace.go
```diff
@@ -0,0 +1,199 @@
...
+// and are therefore transparent to the host filesystem. This has performance
+// advantages if bind mounts are slow, such as on Docker for Mac, but could make
+// debugging harder and is slower when it's time to actually retrieve the diff.
+type dockerVolumeWorkspace struct {
+	volume string
+}
+
+var _ Workspace = &dockerVolumeWorkspace{}
+
+func (w *dockerVolumeWorkspace) init(ctx context.Context, client api.Client, repo *graphql.Repository) error {
```
mrnugget, line 42: Feel free to ignore: I would've expected the content of this method to be in `dockerVolumeWorkspaceCreator.Create` instead of being a method on this. The `.init` method is more OOP-like, I think, and it comes down to taste here, but just wanted to mention this thought in case you were unsure about code organisation and needed a bit of input.

internal/campaigns/volume_workspace.go
```diff
@@ -0,0 +1,199 @@
...
+	// That means we need to strip away the `a/` and `/b` prefixes with `--no-prefix`.
+	// See: https://github.com/sourcegraph/sourcegraph/blob/82d5e7e1562fef6be5c0b17f18631040fd330835/enterprise/internal/campaigns/service.go#L324-L329
+	//
+	// Also, we need to add --binary so binary file changes are inlined in the patch.
+	script := `#!/bin/sh
+	
+exec git diff --cached --no-prefix --binary
+`
+
+	out, err := w.runScript(ctx, "/work", script)
```
mrnugget, line 151: Not sure how you'll feel about this, I might be conflating implementation-detail and interface here, but since you're explicitly passing the script to `sh`, I think you don't need the shebang line, right?

```suggestion
	out, err := w.runScript(ctx, "/work", "exec git diff --cached --no-prefix --binary")
```

internal/campaigns/volume_workspace.go
```diff
@@ -0,0 +1,199 @@
...
+	}, nil
+}
+
+func (w *dockerVolumeWorkspace) WorkDir() *string { return nil }
+
+func (w *dockerVolumeWorkspace) Prepare(ctx context.Context) error {
+	script := `#!/bin/sh
+	
+set -e
+set -x
```
mrnugget, line 103: Where do users see the output of `set -x`? Only in the case of error, right?

internal/campaigns/volume_workspace_test.go
```diff
@@ -0,0 +1,392 @@
...
+			t.Run(name, func(t *testing.T) {
+				expect.Commands(
+					t,
+					expect.NewGlob(
+						expect.Behaviour{Stdout: bytes.TrimSpace([]byte(tc.stdout))},
+						"docker", "run", "--rm", "--init", "--workdir", "/work",
+						"--mount", "type=bind,source=*,target=/run.sh,ro",
+						"--mount", "type=volume,source="+volumeID+",target=/work",
+						dockerWorkspaceImage,
+						"sh", "/run.sh",
```
mrnugget, line 235: Neat!

internal/exec/expect/expect.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// ExitCode is the exit code that the process will exit with.
+	ExitCode int
+}
+
+// Commands defines a set of expected commands for the given test. Commands may
+// be called from any number of nested subtests, but must only be called once
+// from a single test function, as it uses (*testing.T).Cleanup to manage
+// per-test state.
+func Commands(t *testing.T, exp ...*Expectation) {
```
mrnugget, line 54: This is really cool stuff, man!

mrnugget (approved):
Fantastic, thorough work - docs, tests, scripts, code. And things can get up to 4 times faster. Good start into the new year :)

DEVELOPMENT.md
```diff
@@ -46,3 +46,19 @@ We use AppVeyor to test `src-cli` on Windows.
 Configure the AppVeyor builds by editing the `appveyor.yml` file and logging in to AppVeyor and changing the settings there.
 
 Login with your GitHub account, switch to the `sourcegraph` account and change the settings here: https://ci.appveyor.com/project/sourcegraph/src-cli/settings/environment
+
+## Dependent Docker images
+
+`src campaign apply` and `src campaign preview` use a Docker image published as `sourcegraph/src-campaign-volume-workspace` for utility purposes when the volume workspace is selected, which is the default on macOS. This [Docker image](./docker/campaign-volume-workspace/Dockerfile) is built by [a Python script](./docker/campaign-volume-workspace/push.sh) invoked by the GitHub Action workflow described in [`docker.yml`](.github/workflows/docker.yml).
```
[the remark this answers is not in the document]
LawnGnome: :snake:

internal/api/mock/mock.go
```diff
@@ -0,0 +1,61 @@
...
+var _ http.Handler = &mockHandler{}
+
+func (ms *mockHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	if ms.t != nil {
+		ms.t.Logf("handling request: %+v", r)
+	}
+
+	w.WriteHeader(200)
+	w.Write(ms.data)
+}
```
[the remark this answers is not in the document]
LawnGnome: Because it used to be a simpler handler until I wanted to log stuff, and I didn't step back at the end to consider simplifying it. :smiley:

I'm going to adopt your suggestion and then tweak it as needed. Thanks!

internal/campaigns/volume_workspace.go
LawnGnome: You are correct: this was a stylistic choice rather than a technical one. I like having the shebang to reinforce that we're using Bourne shell (as opposed to bash) here, but I also accept that it makes the source code uglier. I'm not particularly tied to this, though — if you feel strongly about this, I'm happy to change it.

internal/campaigns/volume_workspace.go
LawnGnome: Yep, which I think is also the only case we'll be interested in it.

internal/campaigns/bind_workspace.go
```diff
@@ -18,19 +16,29 @@ import (
...
 	if err != nil {
-		return "", err
+		return nil, err
+	}
+
+	// Unlike the mkdirAll() calls elsewhere in this file, this is only giving
+	// us a temporary place on the filesystem to keep the archive. Since it's
+	// never mounted into the containers being run, we can keep these
+	// directories 0700 without issue.
+	if err := os.MkdirAll(filepath.Dir(path), 0700); err != nil {
```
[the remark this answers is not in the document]
LawnGnome: Good catch!

internal/campaigns/bind_workspace_test.go
LawnGnome: Yep, agreed.

internal/campaigns/bind_workspace_test.go
LawnGnome: No, that's betraying `WorkDir()` being a relatively late addition to the interface. I'll change that.

internal/campaigns/run_steps.go
LawnGnome: Agreed, and the fact that changing it only affects this file and its associated test file probably reinforces that.

internal/campaigns/workspace.go
LawnGnome: I went back and forth on this: originally, this survived because I hadn't fully fleshed out the creator interface and was trying to avoid changing too much about how the bind workspace operated (since it had an analogue for `Prepare`), but in the end, I left it mostly out of inertia and a vague feeling that there might be scenarios where it makes sense to keep these separate.

But, realistically, this is used in one place right now, immediately after creating the workspace. So I think you're right, and I'll change it.

internal/campaigns/volume_workspace.go
LawnGnome: I refactored this as part of the removal of `Prepare()` to be on the creator, and with a more broken down set of methods.

LawnGnome:
OK, so this has changed around a bit in response to @mrnugget's review. Most notably: there's now a `RepoFetcher` type that is responsible for fetching repos (gasp!) and is called by `runSteps()`, which means that repo fetching logic is no longer within the workspace types (and, as a side benefit, means that the volume workspace uses the archive cache correctly).

PTAL!

eseliger (approved):
Very nice work 🌟

cmd/src/campaigns_common.go
```diff
@@ -100,6 +101,20 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 		"If true, errors encountered while executing steps in a repository won't stop the execution of the campaign spec but only cause that repository to be skipped.",
 	)
 
+	// We default to bind workspaces on everything except AMD64 macOS at
```
mrnugget, line 104:
```suggestion
	// We default to bind workspaces on everything except ARM64 macOS at
```

Right? Maybe we should use "macOS on M1 chips" as the wording 😄

cmd/src/campaigns_common.go
```diff
@@ -210,10 +227,10 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
 
 	imageProgress := out.Progress([]output.ProgressBar{{
 		Label: "Preparing container images",
-		Max:   float64(len(campaignSpec.Steps)),
+		Max:   1.0,
```
mrnugget, line 230: Ohhhh, nice catch

internal/campaigns/repo_fetcher_test.go
```diff
@@ -0,0 +1,203 @@
...
+		wantZipFile := "github.com-sourcegraph-src-cli-d34db33f.zip"
+		ok, err := dirContains(rf.dir, wantZipFile)
+		if err != nil {
+			t.Fatal(err)
+		}
+		if !ok {
+			t.Fatalf("temp dir doesnt contain zip file")
+		}
+
+		// We can't use path.Join() here because it hardcodes / as the path
```
mrnugget, line 74: What about filepath.Join?

internal/campaigns/run_steps.go
```diff
@@ -17,51 +17,20 @@ import (
 	"github.com/sourcegraph/src-cli/internal/campaigns/graphql"
 )
 
-func runSteps(ctx context.Context, wc *WorkspaceCreator, repo *graphql.Repository, steps []Step, logger *TaskLogger, tempDir string, reportProgress func(string)) ([]byte, error) {
+func runSteps(ctx context.Context, rf RepoFetcher, wc WorkspaceCreator, repo *graphql.Repository, steps []Step, logger *TaskLogger, tempDir string, reportProgress func(string)) ([]byte, error) {
```
mrnugget, line 20: I like the separation of `RepoFetcher` and `WorkspaceCreator` - makes a lot of sense now looking at them 👍

internal/campaigns/volume_workspace.go
mrnugget: Makes sense!

internal/campaigns/workspace.go
```diff
@@ -0,0 +1,88 @@
...
+	// Changes is called after each step is executed, and should return the
+	// cumulative file changes that have occurred since Prepare was called.
+	Changes(ctx context.Context) (*StepChanges, error)
+
+	// Diff should return the total diff for the workspace. This may be called
+	// multiple times in the life of a workspace.
+	Diff(ctx context.Context) ([]byte, error)
+}
+
+// We'll put some useful utility functions below here that tend to be reused
```
mrnugget, line 54: I think these could now we moved to the `repo_fetcher.go` file. But: not a blocker :)

mrnugget (approved):

cmd/src/campaigns_common.go
LawnGnome: Ugh, thanks. I definitely did this a few other times (that I caught) when I was writing this.

I'm trying to avoid M1, since there will presumably be more CPUs in the future and we don't know Apple's naming scheme yet, but this is definitely a drawback to using the more technically correct name.

internal/campaigns/repo_fetcher_test.go
LawnGnome: I... uh, don't know why I forgot that existed. Thanks!

internal/campaigns/workspace.go
LawnGnome: Good call; done.

PR #413, opened 2020-12-27
config: fix subjects flag, add example for editing global config

bobheadxi:
Currently, providing a `-subject` flag to `src config` does not apply it.

```
+ src config edit -subject=U2l0ZToic2l0ZSI= -property=notices '-value=[]'
GraphQL errors: 1 error occurred:
        * {
  "locations": [
    {
      "column": 39,
      "line": 2
    }
  ],
  "message": "Variable \"subject\" has invalid value null.\nExpected type \"ID!\", found null."
}
```

Also, the docs state that this command can edit global config, but does not imply how, so I added an example with a workaround - let me know if there's a proper way to do this!

### DIFF
--- a/cmd/src/config_edit.go
+++ b/cmd/src/config_edit.go
@@ -34,8 +34,8 @@ Examples:
 
     	$ src config edit -subject=$(src orgs get -f '{{.ID}}' -name=abc-org) -overwrite -value '{"motd":["Hello!"]}'
 
-
   Change global settings:
+
     	$ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
 `
 
--- a/cmd/src/config_edit.go
+++ b/cmd/src/config_edit.go
@@ -36,7 +36,7 @@ Examples:
 
   Change global settings:
 		
-        $ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
+      $ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
 `
 
 	flagSet := flag.NewFlagSet("edit", flag.ExitOnError)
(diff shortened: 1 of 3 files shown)


cmd/src/config_edit.go
```diff
@@ -34,6 +34,9 @@ Examples:
 
     	$ src config edit -subject=$(src orgs get -f '{{.ID}}' -name=abc-org) -overwrite -value '{"motd":["Hello!"]}'
 
+  Change global settings:
+		
+        $ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
```
LawnGnome, line 39: The indentation on this was slightly off compared to the earlier examples:

```suggestion
      $ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
```

LawnGnome (approved):
Good catch!

cmd/src/config_edit.go
```diff
@@ -34,6 +34,9 @@ Examples:
 
     	$ src config edit -subject=$(src orgs get -f '{{.ID}}' -name=abc-org) -overwrite -value '{"motd":["Hello!"]}'
 
+  Change global settings:
+		
+      $ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
```
bobheadxi, line 39:
```suggestion
    	$ src config edit -subject=$(echo 'query { site { id } }' | src api | jq .data.site.id --raw-output)
```

cmd/src/config_edit.go
```diff
@@ -34,6 +34,9 @@ Examples:
 
     	$ src config edit -subject=$(src orgs get -f '{{.ID}}' -name=abc-org) -overwrite -value '{"motd":["Hello!"]}'
 
+  Change global settings:
+
```
bobheadxi, line 38:
```suggestion

```

cmd/src/config_edit.go
```diff
@@ -34,6 +34,8 @@ Examples:
 
     	$ src config edit -subject=$(src orgs get -f '{{.ID}}' -name=abc-org) -overwrite -value '{"motd":["Hello!"]}'
 
+  Change global settings:
```
bobheadxi, line 37:
```suggestion

  Change global settings:
```

PR #414, opened 2020-12-27
login: default to SRC_ENDPOINT

bobheadxi:
Allows `src login` without an argument to validate against `SRC_ENDPOINT` by default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Changed
 
+- `src login` now defaults to validating against `SRC_ENDPOINT` if configured.
+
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
This LGTM, but I'm curious if @sqs has any thoughts on whether this makes sense for the intended use of `src login`.

sqs:
Seems reasonable to me. As is, the command doesn’t ever automatically update your config, it just tells you what to do. So, it’s not as though running it with a URL will update your config and running it without one won’t (ie, there is no meaningful difference in behavior between the two forms).

 If a user has no idea what self-hosted means and is just using Sourcegraph.com, “src login” with this new default behavior would do exactly what they expect.

PR #415, opened 2020-12-30
ci: migrate all CI to GitHub Actions

LawnGnome:
I ran into some Windows CI issues on #412, and it got me thinking. Nature abhors a monoculture, but GitHub Actions can now do everything AppVeyor could for us, and more (macOS support), and less jankily (_why_ can't `go test` invocations call back into themselves, only on AppVeyor?), and we're not really tying ourselves to it any more, since we're just executing `go test`. Why not unify everything on GitHub Actions?

(As a bonus, this also means that my shiny new tests in #412 will pass, but I promise that's not the only reason I'm doing this.)

This will fail CI right now because AppVeyor will try to build even without an `appveyor.yml`, but everything else should pass. If so, I think we should go ahead and do this.


mrnugget (approved):
Ah, this is nice. Yes, let's do it.

chrispine (approved):
LGTM 👍🏻

eseliger (commented):
nice! Even macOS CI now

PR #422, opened 2021-01-12
campaigns: fall back to bind workspaces when running steps with mixed users

LawnGnome:
This fixes sourcegraph/sourcegraph#17172. I'm inclined to get this in before I try releasing 3.23.2 again.

The only real drawback of this PR (besides the extra lines of code) is that `src` will now always ensure the `sourcegraph/src-campaign-volume-workspace` container is available, since we don't know at the point of downloading Docker images which workspace we'll use. (To figure that out, we'd need the images, and then the snake eats itself.) Since Docker is good at caching, I'm not too concerned: this is only one extra `docker image inspect` call in any case except for the very first cold run with a new version. (Also, we can _definitely_ optimise those more later.)

Speaking of versioning, there's one other change to make this work as well: when I implemented #412, I missed that newer versions of the workspace image wouldn't be pulled when new `src` versions were released. I've fixed this now (otherwise all my careful Docker image tagging would be for naught!), but note there's a touch of extra complexity in here around moving the `buildTag` to another package so we can access it. I think it's fine, but it's worth an extra set of eyes.

### DIFF
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -102,8 +102,8 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 	)
 
 	flagSet.StringVar(
-		&caf.workspace, "workspace", "default",
-		`Workspace mode to use ("default", "bind", or "volume")`,
+		&caf.workspace, "workspace", "auto",
+		`Workspace mode to use ("auto", "bind", or "volume")`,
 	)
 
 	flagSet.BoolVar(verbose, "v", false, "print verbose output")
--- a/cmd/src/campaigns_common.go
+++ b/cmd/src/campaigns_common.go
@@ -101,18 +101,9 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 		"If true, errors encountered while executing steps in a repository won't stop the execution of the campaign spec but only cause that repository to be skipped.",
 	)
 
-	// We default to bind workspaces on everything except ARM64 macOS at
-	// present. In the future, we'll likely want to switch the default for ARM64
-	// macOS as well, but this requires access to the hardware for testing.
-	var defaultWorkspace string
-	if runtime.GOOS == "darwin" && runtime.GOARCH == "amd64" {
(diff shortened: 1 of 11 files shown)


cmd/src/campaigns_common.go
```diff
@@ -101,18 +101,9 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
...
-	if runtime.GOOS == "darwin" && runtime.GOARCH == "amd64" {
-		defaultWorkspace = "volume"
-	} else {
-		defaultWorkspace = "bind"
-	}
 	flagSet.StringVar(
-		&caf.workspace, "workspace", defaultWorkspace,
-		`Workspace mode to use ("bind" or "volume")`,
+		&caf.workspace, "workspace", "default",
+		`Workspace mode to use ("default", "bind", or "volume")`,
```
mrnugget, line 106: Would "auto" be a better name for what's happening?

```suggestion
		&caf.workspace, "workspace", "auto",
		`Workspace mode to use ("auto", "bind", or "volume")`,
```

internal/campaigns/service.go
```diff
@@ -243,20 +252,18 @@ func (dis dockerImageSet) add(image string, digestPtr *string) {
...
 	for i, step := range spec.Steps {
 		images.add(step.Container, &spec.Steps[i].image)
 	}
 
-	// The workspace creator may have its own dependencies.
-	for _, image := range creator.DockerImages() {
-		images.add(image, nil)
-	}
+	// We also need to ensure we have our own utility images available.
+	images.add(dockerVolumeWorkspaceImage, nil)
```
mrnugget, line 266: 👍 on reducing the interface and hardcoding the image here.

mrnugget (approved):
Looks good to me! Left just a tiny suggestion regarding the default CLI param value.

cmd/src/campaigns_common.go
LawnGnome: Yeah, I wrestled with that too: the `best` function name was an artefact of a different iteration on the name.

`auto` feels better than `default` to me, so I'm going to commit this. Thanks!

eseliger (approved):
nice workaround 👍

PR #424, opened 2021-01-12
Enable dynamic changesetTemplates by adding templating & steps outputs

mrnugget:
### What?

This PR enables **dynamic `changesetTemplates`** in campaign specs by doing two things:

1. it adds templating support to the `title`, `body`, `branch` and `commit.message` fields in `changesetTemplate`.
2. it adds `outputs` to `steps`, which are values that are set after the step has been executed and that have _access to template variables, including the current step's output_.

This solves the problem of not being able to customize the `changesetTemplate` per repository (only `published` could be configured per repository).

### Show me

With this PR, the following campaign spec works:

```yaml
name: dynamic-changeset-template
description: Hello Outputs

on:
  - repository: github.com/sourcegraph/src-cli

steps:
  - run: echo "Hello there!" >> message.txt && cat message.txt
    container: alpine:3
    # NEW:
    outputs:
      friendlyMessage:
        value: "${{ step.stdout }}"
        # format: text is implicit
  - run: cat .goreleaser.yml >&2
    container: alpine:3
    # NEW:
    outputs:
      goreleaserConfig:
        value: "${{ step.stderr }}"
        # we will parse the value here as YAML, meaning that we can access it as an object
        format: yaml
      goreleaserConfigExists:
        # we can use the power of Go's text/template engine to dynamically produce complex values
        value: "exists: ${{ gt (len step.stderr) 0 }}"
        format: yaml

changesetTemplate:
  # title allows templating
  title: "The friendly message is: `${{ outputs.friendlyMessage }}`"
  # body allows templating
  body: |
    This repository has a `gorelaserConfig`: ${{ outputs.goreleaserConfigExists.exists }}.

    (Since these are just Go `text/templates` and `goreleaserConfig` was parsed as YAML we can access every field)

    The `goreleaser.yml` defines the following `before.hooks`:

    ${{ range $index, $hook := outputs.goreleaserConfig.before.hooks }}
    - `${{ $hook }}`
    ${{ end }}

    These files were added by the steps:

    ```
    ${{ join steps.added_files " " }}
    ```

  # branch allows templating
  branch: thorsten/variables-${{ outputs.goreleaserConfig.dist }}
  published: draft
  commit:
    # commit message allows templating
    message: "The friendly message is: `${{ outputs.friendlyMessage }}`"
```

It created [this pull request](https://github.com/sourcegraph/src-cli/pull/428):

screenshot_2021-01-14_15 38 19@2x

### Meta

PR in `sourcegraph/sourcegraph` that adds the schema: https://github.com/sourcegraph/sourcegraph/pull/17243

### What's missing?

- [x] BIG FAT **TODO**: caching! we need to cache the outputs too now, not just the diff.
- [x] ~Documentation. I want to add docs for this in https://github.com/sourcegraph/sourcegraph/pull/17243 and also better document templating support, since we now lean on it more heavily.~ PR is here: https://github.com/sourcegraph/sourcegraph/pull/17286

### Possible follow-ups

- [ ] Can be a follow-up PR: making `transformChanges.group.branch` optional by making `transform_changes.group` available as a template variable in `changesetTemplate.branch`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,6 +30,7 @@ require (
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
 	golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4
 	gopkg.in/yaml.v2 v2.3.0
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
 	jaytaylor.com/html2text v0.0.0-20200412013138-3577fbdbcff7
 )
 
--- a/internal/campaigns/execution_cache.go
+++ b/internal/campaigns/execution_cache.go
@@ -62,8 +62,8 @@ func (key ExecutionCacheKey) Key() (string, error) {
 }
 
 type ExecutionCache interface {
-	Get(ctx context.Context, key ExecutionCacheKey) (diff string, found bool, err error)
-	Set(ctx context.Context, key ExecutionCacheKey, diff string) error
+	Get(ctx context.Context, key ExecutionCacheKey) (diff string, outputs map[string]interface{}, found bool, err error)
+	Set(ctx context.Context, key ExecutionCacheKey, diff string, outputs map[string]interface{}) error
 	Clear(ctx context.Context, key ExecutionCacheKey) error
 }
 
@@ -77,13 +77,20 @@ func (c ExecutionDiskCache) cacheFilePath(key ExecutionCacheKey) (string, error)
 		return "", errors.Wrap(err, "calculating execution cache key")
 	}
 
-	return filepath.Join(c.Dir, keyString+".diff"), nil
+	return filepath.Join(c.Dir, keyString+".v3.json"), nil
 }
 
-func (c ExecutionDiskCache) Get(ctx context.Context, key ExecutionCacheKey) (string, bool, error) {
+type executionResult struct {
+	Diff    string                 `json:"diff"`
+	Outputs map[string]interface{} `json:"outputs"`
+}
+
+func (c ExecutionDiskCache) Get(ctx context.Context, key ExecutionCacheKey) (string, map[string]interface{}, bool, error) {
+	outputs := map[string]interface{}{}
+
 	path, err := c.cacheFilePath(key)
 	if err != nil {
-		return "", false, err
+		return "", outputs, false, err
 	}
 
 	data, err := ioutil.ReadFile(path)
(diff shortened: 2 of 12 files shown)


chrispine:
This is so cool! I'm thrilled to see this coming together!

mrnugget:
This is ready for review. I'll add the docs in a PR to sourcegraph/sourcegraph tomorrow, @sourcegraph/campaigns.

eseliger (approved):
N i c e, 🚢 it!!!11

internal/campaigns/execution_cache.go
```diff
@@ -77,58 +77,91 @@ func (c ExecutionDiskCache) cacheFilePath(key ExecutionCacheKey) (string, error)
 		return "", errors.Wrap(err, "calculating execution cache key")
 	}
 
-	return filepath.Join(c.Dir, keyString+".diff"), nil
+	return filepath.Join(c.Dir, keyString+".v3.json"), nil
```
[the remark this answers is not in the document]
mrnugget: Ah, I got you: `v3.0.1.json` coming up soon

go.mod
```diff
@@ -30,6 +30,7 @@ require (
 	golang.org/x/sync v0.0.0-20190911185100-cd5d95a43a6e
 	golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4
 	gopkg.in/yaml.v2 v2.3.0
+	gopkg.in/yaml.v3 v3.0.0-20200615113413-eeeca48fe776
```
LawnGnome, line 33: Not for this PR, but do you think it's worth us migrating our other YAML code to v3 so we don't have to pull in both versions?

internal/campaigns/execution_cache.go
```diff
@@ -77,58 +77,91 @@ func (c ExecutionDiskCache) cacheFilePath(key ExecutionCacheKey) (string, error)
...
-	// To be backwards compatible, we keep reading these:
-	if strings.HasSuffix(path, ".json") {
-		var result ChangesetSpec
+	// There are now three different cache versions out in the wild and to be
+	// backwards compatible we read all of them.
+	// 2-step plan for taming this:
+	//  1) February 2021: deprecate old caches by deleting the files when
+	//                    detected and reporting as cache-miss.
+	//  2) May 2021 (two releases later): remove handling of these files from
+	//                                    this function
```
LawnGnome, line 105: :+1:, particularly since the worst case here is that some campaign steps have to be re-executed.

Can you please open a GitHub issue to cover the follow up action, assuming you haven't already?

internal/campaigns/execution_cache.go
```diff
@@ -77,58 +77,91 @@ func (c ExecutionDiskCache) cacheFilePath(key ExecutionCacheKey) (string, error)
...
+	switch {
```
LawnGnome, line 106: Can we hit any of the branches here other than `v3.json`, given that's what `cacheFilePath()` always returns?

internal/campaigns/executor.go
```diff
@@ -422,11 +458,11 @@ func createChangesetSpecs(task *Task, completeDiff string, features featureFlags
...
-				Title:          task.Template.Title,
-				Body:           task.Template.Body,
+				Title:          title,
+				Body:           body,
 				Commits: []GitCommitDescription{
 					{
-						Message:     task.Template.Commit.Message,
+						Message:     message,
 						AuthorName:  authorName,
 						AuthorEmail: authorEmail,
```
LawnGnome, line 467: Should we make all the fields eligible for templates for consistency?

internal/campaigns/templating.go
```diff
@@ -40,44 +40,64 @@ func renderMap(m map[string]string, stepCtx *StepContext) (map[string]string, er
 	return rendered, nil
 }
 
-// StepContext represents the contextual information available when executing a
-// step that's defined in a campaign spec.
+// StepContext represents the contextual information available when render a
```
LawnGnome, line 43:
```suggestion
// StepContext represents the contextual information available when rendering a
```

internal/campaigns/templating.go
```diff
@@ -40,44 +40,64 @@ func renderMap(m map[string]string, stepCtx *StepContext) (map[string]string, er
...
+	Outputs map[string]interface{}
+	// Step is the result of the current step. Empty when evaluating the "run" field
+	// but filled when evaluating the "outputs" field.
+	Step StepResult
+	// PreviousStep is the result of the previous step. Empty when there is no
+	// previous step.
 	PreviousStep StepResult
-	Repository   graphql.Repository
+	// Repository is the Sourcegraph repository in which the steps are executed.
+	Repository graphql.Repository
```
LawnGnome, line 55: Good docs!

LawnGnome (commented):
I have a question around forward compatibility of cached changeset specs (which I think was probably already present, but is more obvious now), but otherwise this looks good! :+1:

internal/campaigns/execution_cache.go
mrnugget: Oh, wow. Nice catch. Turns out I'm an idiot: this code was never backwards compatible. I'll try to see if I can make it work with globbing. If I can't: I think it shouldn't be a problem if we bust the cache instead of keeping backwards compatibility, since this was more of a "ah, shouldn't be hard to make it work" thing. What do you think? @eseliger?

internal/campaigns/executor.go
mrnugget: Yeah, good idea. Just did that in 585abbd

internal/campaigns/execution_cache.go
mrnugget: Done: https://github.com/sourcegraph/sourcegraph/issues/17306

go.mod
mrnugget: Yeah, I think so. I just migrated all of our campaigns code to v3 (two other places in `_test.go` files) and the only other place that uses yaml is in `cmd/src/validate.go`. Ticket: https://github.com/sourcegraph/src-cli/issues/429

mrnugget:
@LawnGnome @eseliger After Adam spotted in bug in [this comment](https://github.com/sourcegraph/src-cli/pull/424#discussion_r557674000) I reworked the `ExecutionDiskCache` a bit so that

1. it actually works and reads old cache files
2. new: writes a new cache and deletes the old one if an old one was read
3. new: handles multiple old cache files

I went a bit overboard with the tests, maybe, but they were a joy to write, so... reason enough to up the test coverage :)

Since we do have a branch cut later today, it would be great if @eseliger can review this and we merge this before then. @LawnGnome you can then also take another look and if there are issues we can follow up, but I want to cut a 3.24 release with this in it and bump the `MiniumVersion` in `sourcegraph/sourcegraph` before branch cut. We can always merge follow-up fixes and release a src-cli 3.24.1 then.

eseliger (approved):
:shipit:

mrnugget:
@LawnGnome We just reviewed this together, Erik and I. And Erik, of course, found something: the conversion from old caches to new caches is not lossless, since we don't recreate the `steps.{modified,added,deleted,renamed}_files` template var. Means in practice: if a user ran a campaign, populated caches, then thinks: "ahh, new template vars, let's add `steps.added_files` to my campaign spec and run again", then that wouldn't work.

LawnGnome:
> @LawnGnome you can then also take another look

I don't see anything problematic from where I sit, besides the cache lossiness issue Erik saw (which I'd be _amazed_ if anyone ever hit). :+1:

PR #425, opened 2021-01-12
Adding environment variable clarification to readme

emchap:
@LawnGnome Taking a stab at updating the readme to reduce the confusion I ran into when setting up. Do you mind taking a look and confirming it looks okay?


LawnGnome (approved):
LGTM! Thanks for doing that.

PR #426, opened 2021-01-13
Add verbosity levels to lsif upload

efritz:
Add trace logging to lsif upload actions.

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -4,14 +4,15 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
+	"sort"
 	"strings"
 	"sync"
 	"time"
 
 	"github.com/efritz/pentimento"
-	"github.com/mattn/go-isatty"
 	"github.com/pkg/browser"
 	"github.com/pkg/errors"
 	"github.com/sourcegraph/codeintelutils"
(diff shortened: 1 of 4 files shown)


cmd/src/lsif_upload.go
```diff
@@ -67,6 +70,7 @@ Examples:
 	flags.maxPayloadSizeMb = flagSet.Int("max-payload-size", 100, `The maximum upload size (in megabytes). Indexes exceeding this limit will be uploaded over multiple HTTP requests.`)
 	flags.ignoreUploadFailures = flagSet.Bool("ignore-upload-failure", false, `Exit with status code zero on upload failure.`)
 	flags.uploadRoute = flagSet.String("upload-route", "/.api/lsif/upload", "The path of the upload route.")
+	flags.rawVerbosity = flagSet.Int("v", 0, "-v=0 shows no logs; -v=1 shows requests and response metadata; -v=2 shows headers, -v=3 shows response body")
```
eseliger, line 73: is there no issue with this subcommand defining a "v" flag, and the main command doing so, too?

eseliger (approved):

cmd/src/lsif_upload.go
efritz: Does the main command currently define this flag?

eseliger: Yes it does. Okay, seems fine then.

 I think this is possible now, but doesn't matter I think: 

`src -v lsif upload -v=-1` :D

efritz: Ah. I'll rename to avoid the confusion.

PR #430, opened 2021-01-18
codeintel: add an --associated-index-id flag for autoindexer usage

gbrik:
Need to wait for the codeintelutils update before this works.

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -70,7 +70,7 @@ Examples:
 	flags.noProgress = flagSet.Bool("no-progress", false, `Do not display a progress bar.`)
 	flags.maxPayloadSizeMb = flagSet.Int("max-payload-size", 100, `The maximum upload size (in megabytes). Indexes exceeding this limit will be uploaded over multiple HTTP requests.`)
 	flags.ignoreUploadFailures = flagSet.Bool("ignore-upload-failure", false, `Exit with status code zero on upload failure.`)
-	flags.uploadRoute = flagSet.String("upload-route", "/.api/lsif/upload", "The path of the upload route.")
+	flags.uploadRoute = flagSet.String("upload-route", "/.api/lsif/upload", "The path of the upload route. For internal use only.")
 	flags.rawVerbosity = flagSet.Int("trace", 0, "-trace=0 shows no logs; -trace=1 shows requests and response metadata; -trace=2 shows headers, -trace=3 shows response body")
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/pkg/errors v0.9.1
 	github.com/sourcegraph/campaignutils v0.0.0-20201124055807-2f9cfa9317e2
-	github.com/sourcegraph/codeintelutils v0.0.0-20210113171425-9ec641b48a8e
+	github.com/sourcegraph/codeintelutils v0.0.0-20210118225004-6e6891d301ca
```
efritz, line 24: Can we bump this to a new commit if we handle https://github.com/sourcegraph/codeintelutils/pull/10#discussion_r559824707? (I think it was misread, we should call it `associatedIndexId` instead of `associatedIndexID` to be consistent with the existing parameters.

cmd/src/lsif_upload.go
```diff
@@ -71,6 +72,7 @@ Examples:
 	flags.ignoreUploadFailures = flagSet.Bool("ignore-upload-failure", false, `Exit with status code zero on upload failure.`)
 	flags.uploadRoute = flagSet.String("upload-route", "/.api/lsif/upload", "The path of the upload route.")
```
efritz, line 73: Should we add 'internal use only' here too?

efritz (approved):

cmd/src/lsif_upload.go
gbrik: Sure, if someone's changed this in a forked Sourcegraph instance they'll happily ignore that instruction.

cmd/src/lsif_upload.go
```diff
@@ -70,7 +70,7 @@ Examples:
 	flags.noProgress = flagSet.Bool("no-progress", false, `Do not display a progress bar.`)
 	flags.maxPayloadSizeMb = flagSet.Int("max-payload-size", 100, `The maximum upload size (in megabytes). Indexes exceeding this limit will be uploaded over multiple HTTP requests.`)
 	flags.ignoreUploadFailures = flagSet.Bool("ignore-upload-failure", false, `Exit with status code zero on upload failure.`)
-	flags.uploadRoute = flagSet.String("upload-route", "/.api/lsif/upload", "The path of the upload route.")
+	flags.uploadRoute = flagSet.String("upload-route", "/.api/lsif/upload", "The path of the upload route. For internal use only.")
```
efritz, line 73: image

efritz (approved):

PR #433, opened 2021-01-19
RIDICULOUS: Set chown in workspace volume and create temp file

mrnugget:


LawnGnome:
OK, so I said I'd polish this off today in the sync. I was, of course, wildly optimistic, because I realised only this afternoon that there's another issue: if the user _explicitly_ selects `-workspace volume`, all that UID detection code doesn't fire, because it's in the best workspace detector. :boom:

I've opened #434, which builds on top of this PR, with an approach I explored. It _may_ be more complicated than we want. Please provide feedback!

LawnGnome:
We ended up building on this in #434, so this is now moot.

PR #435, opened 2021-01-20
Fix all errors reported by golint-ci

mrnugget:
I used VS Code again and got sick of all the warnings popping up. So, here we are. I fixed all the errors reported by running `golint-ci run` in the root of `src-cli`.

### DIFF
--- a/cmd/src/doc.go
+++ b/cmd/src/doc.go
@@ -34,7 +34,9 @@ Examples:
 	`
 
 	handler := func(args []string) error {
-		flagSet.Parse(args)
+		if err := flagSet.Parse(args); err != nil {
+			return err
+		}
 
 		out := output.NewOutput(flagSet.Output(), output.OutputOpts{Verbose: *verbose})
 		if *outputFlag == "" {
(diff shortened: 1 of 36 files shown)


internal/campaigns/executor.go
```diff
@@ -92,11 +92,6 @@ type TaskStatus struct {
 	fileDiffsOnce sync.Once
 }
 
-func (ts *TaskStatus) Clone() *TaskStatus {
```
mrnugget, line 95: Unused function that's also broken because it copies a `sync.Once`.

efritz (approved):

eseliger (approved):

cmd/src/doc.go
```diff
@@ -97,7 +99,9 @@ Examples:
 					}
 					defer file.Close()
 
-					file.WriteString(content)
+					if _, err := file.WriteString(content); err != nil {
+						return err
```
Strum355, line 103: This would break out of the loop, which is _maybe_ what we want, but just double checking the assumption :slightly_smiling_face:

Strum355 (approved):

cmd/src/doc.go
mrnugget: I think that's fine, yeah.

LawnGnome (approved):
So fresh, so clean! Thanks for doing this.

PR #436, opened 2021-01-21
campaigns: use the temp dir when mounting scripts in volume mode 

LawnGnome:
@eseliger noticed this one: volume workspaces were mounting their run scripts from the OS temporary directory, which could cause problems on macOS 11 + Docker 3.1. (And probably other configurations, but it worked for me on macOS 10.15.)

It should be using the same temporary directory as everything else, so let's do that.


eseliger (approved):
🌟

PR #437, opened 2021-01-21
campaigns: temporary switch back to bind for non-root

LawnGnome:
This can be reverted once we land #434.


internal/campaigns/workspace.go
```diff
@@ -128,7 +128,7 @@ func detectBestWorkspaceCreator(ctx context.Context, steps []Step) workspaceCrea
 		}
 
 		uids[uid] = struct{}{}
-		if len(uids) > 1 {
+		if len(uids) > 1 || uid != 0 {
```
LawnGnome, line 131: This is ugly, but simplifies the revert later.

eseliger (approved):

PR #438, opened 2021-01-21
Fix flaky test by using slice for stable iteration

mrnugget:
I don't know what changed but before this fix I could pretty reliably get `TestBestWorkspaceCreator` to fail by running:

```
go test -run TestBestWorkspaceCreator/different_user -count=10 ./internal/campaigns
```

It would fail with

```
--- FAIL: TestBestWorkspaceCreator (0.01s)
    --- FAIL: TestBestWorkspaceCreator/different_user (0.01s)
        expect.go:64: unexpected command: 1 error occurred:
                * unexpected argument at position 4: have="bar" want="foo"
            
        expect.go:108: unexpected number of commands executed: have=1 want=4
FAIL
FAIL    github.com/sourcegraph/src-cli/internal/campaigns       0.536s
FAIL
```

With the fix here I can't get it to fail anymore, so I'm 99.9% sure that the error was due to `range` over `map` not being stable and sometimes setting up the wrong assertions which would then fail.

### DIFF
--- a/internal/campaigns/workspace_test.go
+++ b/internal/campaigns/workspace_test.go
@@ -12,8 +12,12 @@ func TestBestWorkspaceCreator(t *testing.T) {
 	ctx := context.Background()
 	isOverridden := !(runtime.GOOS == "darwin" && runtime.GOARCH == "amd64")
 
+	type imageBehaviour struct {
+		image     string
+		behaviour expect.Behaviour
+	}
 	for name, tc := range map[string]struct {
-		behaviours map[string]expect.Behaviour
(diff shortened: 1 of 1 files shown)


eseliger (approved):
Nice catch! How many "whoat" moments did that take?

mrnugget:
> How many "whoat" moments did that take?

I was honestly surprised with myself by how fast I could find it 😎

LawnGnome:
Sorry, that's totally on me. Years of working with languages with inherently ordered map types has made me lazy. Thanks!

PR #440, opened 2021-01-22
testing: fix expect error when the command is not on the PATH

LawnGnome:
The comment in the diff basically describes this: we had an unintentional coupling in mocked commands with the host's binaries on its `PATH`, which is obviously bad if you're trying to mock away the environment in the first place.


eseliger (approved):
sounds reasonable :)

Issue #180, opened 2020-04-21
Add commands for checking for upgrade and upgrading src CLI in place


ryan-blunden:
As a starting point, I like @slimsag 's suggestion from https://github.com/sourcegraph/src-cli/issues/18#issuecomment-425635219:

```
src update check -> checks for updates, doesn't apply them.
src update -> checks for updates, self-applies them. Will prompt for root/admin privileges.
```

I also liked @felixfbecker's suggestion of having a prompt to let the user know there is an upgrade available (screenshot taken from https://www.npmjs.com/package/update-notifier), but add logic so this is only displayed once per day, not on every invocation.

<img width="462" alt="Screen Shot 2020-04-22 at 9 15 17 AM" src="https://user-images.githubusercontent.com/133014/79924005-1f0c5580-847a-11ea-87c2-a05fa4f94889.png">

joshuabezaleel:
Hi @ryan-blunden , will this be open to public contribution? :slightly_smiling_face: 

I think for `src update check` I can use [google/go-github](https://github.com/google/go-github) to get the latest version available.

I am still not sure for the second one, `src update`, though. What comes in my mind is using Go's `os` and `os/exec` to do some "scripting" for "automatically" download the binary and put it at the appropriate folder like the one in the README's installation steps. What do you think?

Thank you very much in advance! :slightly_smiling_face:

ryan-blunden:
@joshuabezaleel Absolutely! We would love you to take this on. @slimsag What do you think about the suggested approach?

joshuabezaleel:
Hi @slimsag , if it would be okay I want to ask questions on this particular issue.
For `src update check`, how would it be different with the current [`src version`](https://github.com/sourcegraph/src-cli/blob/master/cmd/src/version.go#L31-L40) since the `version` command also output the recommended latest version from `/.api/src-cli/version`? Thank you in advance! :slightly_smiling_face:

mrnugget:
Hey @joshuabezaleel, I think the `src version` command does what `src update check` is supposed to do. But `src version` is newer.

burmudar:
Closing this issue as part of repository maintenance. We are no longer using GitHub issues to track src-cli requests or reports. For future questions or support requests, please contact Sourcegraph Support (mailto:support@sourcegraph.com).

PR #441, opened 2021-01-22
Add command to self-update

joshuabezaleel:
Fixes: #180 

**What and Why**
Add command `src update` to self-update the version of the installed `src-cli` to latest release/recommended version according to the `src version` output.
It works fine in the cases of (1) the directory is writable already or (2) the directory is not writable but the command is run using elevated permission (`sudo` in Linux or via Administrator Powershell at Windows). I attached two videos below that show the scenario of the directory is not writable and asks user for elevated permission. 

**How**
I use the [tj/go-update](https://github.com/tj/go-update) package for updating the binary (retrieving releases, download, installation) since it's the only one existed that I can found. There are some shortage of it though like (1) how I need to use the [InstallTo](https://github.com/tj/go-update/blob/master/update.go#L57) instead of just [Install](https://github.com/tj/go-update/blob/master/update.go#L57) since it didn't support the absolute path for Windows yet and (2) it uses the [c4milo/unpackit](https://github.com/c4milo/unpackit) package which will outputs some error when updating the permission of the files of the release because of chmod on this [particular line](https://github.com/c4milo/unpackit/blob/4ed373e9ef1c895fe04468d813cb628902d89a0a/unpackit.go#L259) (it still works on updating the binary, can be seen on the 2nd video below). 
To ask for the permission elevation on Windows, I use the technique for UAC prompt through the `runas` ShellExecute function which was explained [here](https://gist.github.com/jerblack/d0eb182cc5a1c1d92d92a4c4fcc416c6) and [here](https://www.codeproject.com/Articles/320748/Elevating-During-Runtime).

**Additional context**
This is definitely still a work-in-progress POC, and if this issue feels like too big of a feature to be done by external contributor, feel free to decline 🙂  Looking forward to the inputs on the review. Thank you so much in advance! 😄 

Linux demo:

https://user-images.githubusercontent.com/7043511/105504708-a97c3100-5cfa-11eb-98c6-01771cde3a1d.mp4

Windows demo:

https://user-images.githubusercontent.com/7043511/105504711-abde8b00-5cfa-11eb-88fe-c0cee2155cd6.mp4


joshuabezaleel:
Paging @mrnugget since I can't seem to ask review from reviewers on this PR 🙁

mrnugget:
@joshuabezaleel Thanks so much! What a contribution! I haven't looked at the code yet, but the PR itself looks fantastic 🙂 One of is going to take a closer look in the next few days.

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		if runtime.GOOS == "windows" {
+			updateManager.Command += ".exe"
+		}
+		releases, err := updateManager.LatestReleases()
+		if err != nil {
+			return err
+		}
+		if len(releases) == 0 {
+			return fmt.Errorf("No latest src-cli release update")
```
mrnugget, line 82:
```suggestion
			return fmt.Errorf("No src-cli releases available")
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+		}
+		latestRelease := releases[0]
+		if latestRelease.Version != recommendedVersion {
+			return fmt.Errorf("Mismatch of recommended version")
+		}
```
mrnugget, line 87: Should this be an error? If you're on an older Sourcegraph instance and we recommend an older version of src-cli, why look at the latest release?

I think the check would need to be a bit more nuanced: if there's a newer patch version of the recommended version it's fine to go up. If there's a newer major/patch version it would probably be best to ask the user whether they want to upgrade to that.

(It looks like you intended to do this — there's an `Errorf` here — but the error message should include the versions, I think)

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+
+		if err := updateManager.InstallTo(tarball, dir); err != nil {
+			return fmt.Errorf("Failed installing src-cli: %v\n", err)
+		}
+
+		fmt.Printf("src-cli has been updated from version %s to %s\n", buildTag, recommendedVersion)
+
+		// Keep the terminal installation window open after Windows' UAC privilege elevation.
+		if runtime.GOOS == "windows" {
+			fmt.Println("Press Enter key to exit")
```
mrnugget, line 131: Isn't the famous any key enough here? 😄 
image

```suggestion
			fmt.Println("Press any key to exit")
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+			flagSet.PrintDefaults()
+			fmt.Println(usage)
+		},
+	})
+}
+
+func isDirWritable(dir string) error {
+	f := filepath.Join(dir, ".touch")
+	if err := ioutil.WriteFile(f, []byte(""), PrivateFileMode); err != nil {
+		return fmt.Errorf("Does not have write permission to %v directory: %v\n", dir, err)
```
mrnugget, line 153:
```suggestion
		return fmt.Errorf("writing to directory %s failed: %s", dir, err)
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/sourcegraph/src-cli/internal/api"
+
+	"github.com/tj/go-update"
+	"github.com/tj/go-update/progress"
+	"github.com/tj/go-update/stores/github"
+)
+
+const (
+	// PrivateFileMode grants owner to read/write a file.
+	PrivateFileMode = 0600
```
mrnugget, line 22: Does this need to be exported?

```suggestion
	privateFileMode = 0600
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+		if err != nil {
+			return err
+		}
+		err = isDirWritable(currentDirectory)
+		if err != nil {
+			runWithElevatedPrivilege()
+			return nil
+		}
+
+		fmt.Printf("Current version: %s\n", buildTag)
```
mrnugget, line 49: This moved (which is also one of the reasons why the build is failing on this branch):

```suggestion
		fmt.Printf("Current version: %s\n", version.BuildTag)
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+
+		client := cfg.apiClient(apiFlags, flagSet.Output())
+		recommendedVersion, err := getRecommendedVersion(context.Background(), client)
+		if err != nil {
+			return err
+		}
+		if recommendedVersion == "" {
```
mrnugget, line 56: These lines are exactly the same as in `cmd/src/version.go` - maybe put them in a `getCurrentAndRecommendVersion` function?

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+			fmt.Println("Recommended Version: <unknown>")
+			return nil
+		}
+		if buildTag == recommendedVersion {
```
mrnugget, line 60:
```suggestion
		if version.BuildTag == recommendedVersion {
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+			return nil
+		}
+
+		// Retrieving latest release information.
+		updateManager := &update.Manager{
+			Command: "src",
+			Store: &github.Store{
+				Owner:   "sourcegraph",
+				Repo:    "src-cli",
+				Version: buildTag,
```
mrnugget, line 71:
```suggestion
				Version: version.BuildTag,
```

cmd/src/update_elevated_linux.go
```diff
@@ -0,0 +1,16 @@
...
+
+package main
+
+import (
+	"os"
+	"os/exec"
+)
+
+func runWithElevatedPrivilege() error {
+	cmd := exec.Command("sudo", "/bin/sh", "-c", "./src update")
```
mrnugget, line 11: I don't think this will work, since `src` might not be in the current directory (`./`).

```suggestion
	cmd := exec.Command("sudo", "/bin/sh", "-c", "./src update")
```

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+	var apiFlags = api.NewFlags(flagSet)
+
+	handler := func(args []string) error {
+		currentDirectory, err := os.Getwd()
+		if err != nil {
+			return err
+		}
+		err = isDirWritable(currentDirectory)
+		if err != nil {
+			runWithElevatedPrivilege()
```
mrnugget, line 45: Two questions:

1. Why check whether the _current_ directory is writable? I would've assumed that we'd need to get the absolute path of the currently-executed `src` binary and then its dir?
2. Why run it with elevated privileges automatically? Couldn't we print an error message and tell the user something like "Hey, your `src` is located in `/usr/bin`, and the current user is lacking permissions to update it, please run with adminstrator/sudo permissions"?

cmd/src/update.go
```diff
@@ -0,0 +1,156 @@
...
+		if latestRelease.Version != recommendedVersion {
+			return fmt.Errorf("Mismatch of recommended version")
+		}
+
+		updateAsset := latestRelease.FindTarball(runtime.GOOS, runtime.GOARCH)
+		if updateAsset == nil {
+			return fmt.Errorf("src-cli binary has not been installed in this machine")
+		}
+		fmt.Printf("Downloading src-cli version %s\n", recommendedVersion)
+		tarball, err := updateAsset.DownloadProxy(progress.Reader)
```
mrnugget, line 94: We could use our own `output` package here and save the dependency on `progress`. I just looked at what `progress.Reader` is and what `DownloadProxy` expects and it wouldn't need a lot of code.

Hacky version:

```suggestion
		out := output.NewOutput(flagSet.Output(), output.OutputOpts{})
		tarball, err := updateAsset.DownloadProxy(newProgressReader(out, "Downloading src-cli"))
```

And then somewhere else (ideally in the `output` package?)

```go
func newProgressReader(out *output.Output, message string) update.Proxy {
	return func(size int, r io.ReadCloser) io.ReadCloser {
		progress := out.Progress([]output.ProgressBar{{
			Label: message,
			Max:   float64(size),
		}}, nil)
		return &customThing{bars: progress, r: r}
	}
}

var _ io.ReadCloser = &customThing{}

type customThing struct {
	bars    output.Progress
	r       io.ReadCloser
	written int
}

func (c *customThing) Read(b []byte) (int, error) {
	n, err := c.r.Read(b)
	c.written += n
	c.bars.SetValue(0, float64(c.written))
	return n, err
}

func (c *customThing) Close() error {
	c.bars.Complete()
	return c.r.Close()
}
```

Works:


https://user-images.githubusercontent.com/1185253/106127224-9fc15600-615e-11eb-933d-225deffeac3f.mp4

mrnugget (commented):
Thanks again for opening this PR! I left some comments for suggestions/ideas.

The most important ones:
1. I don't think we should try to run with more priviliges without asking the user. I'd rather print a message and ask the user to use `sudo` or run the Windows Terminal as admin.
2. Build is broken for me locally.

Happy to do another round of reviews once we've gone through the feedback here!

mrnugget:
Hey @joshuabezaleel! Again, thanks for opening this PR and doing a lot of the work. We're closing it for now, but feel free to re-open if you want to address the review comments from above.

PR #442, opened 2021-01-25
Automatic workspace discovery

mrnugget:
### What?

This adds _automatic workspace discovery_ to src-cli. It allows users to
1. define where projects in large repositories live and
2. run the campaign specs `steps` in those project folders, turning them into _workspaces_.

### How?

Users define workspaces like so:

```yaml
workspaces:
  - rootAtLocationOf: go.mod
    in: github.com/sourcegraph/sourc*
```

That means: in every repository that starts with `github.com/sourcegraph/sourc` projects have a `go.mod` at its root and those folders should be used as `workspaces` for the execution of campaign spec `steps`.

src-cli uses Sourcegraph search under the hood to search for the locations of the `rootAtLocationOf` file, which means it doesn't need to download the repository first and search the file system.

`workspaces` can also contain multiple definitions, matching different repositories (but a repository cannot be matched by multiple definitions):

```yaml
on:
  - repositoriesMatchingQuery: repo:automation-testing$
  - repository: github.com/sourcegraph/sourcegraph

workspaces:
  - rootAtLocationOf: go.mod
    in: github.com/sourcegraph/sourcegraph*
  - rootAtLocationOf: package.json
    in: "*automation-testing"
```

Since multiple workspaces per repository means that multiple changesets will be produced in a single repository, the `changesetTemplate.branch` needs to use templating to avoid name clashes.

For that, users can access the template variable `steps.path` and use helper functions to generate a unique branch name per changeset. Example:

```yaml
changesetTemplate:
  # [...]
  branch: ${{ join_if "-" "thorsten/workspace-discovery" (replace steps.path "/" "-") }}
  published: false
  commit:
    message: Automatic workspace discovery
```

_(The `join_if` and the `replace` helpers are new. `join_if` joins the given list of strings, but ignoring the blank strings. `replace` is an alias for [`strings.ReplaceAll`](https://golang.org/pkg/strings/#ReplaceAll))_

Users can, of course, also user other ways to generate a unique branch name per changeset. With `outputs`, for example:

```yaml
steps:
  # [...]
  - run:  if [[ -f "package.json" ]]; then cat package.json | jq -j .name; fi
    container: jiapantw/jq-alpine:latest
    outputs:
      projectName:
        value: ${{ step.stdout }}

changesetTemplate:
  # [...]
  branch: thorsten/workspace-discovery-${{ outputs.projectName }}
```

Or, in combination:

```yaml
steps:
  # [...]
  - run:  if [[ -f "package.json" ]]; then cat package.json | jq -j .name; fi
    container: jiapantw/jq-alpine:latest
    outputs:
      projectName:
        value: ${{ step.stdout }}

changesetTemplate:
  # If we have an `outputs.projectName` we use it, otherwise we append the path
  # of the workspace. If the path is emtpy (as is the case in the root folder),
  # we ignore it.
  branch: |
    ${{ if eq outputs.projectName "" }}
    ${{ join_if "-" "thorsten/workspace-discovery" (replace steps.path "/" "-") }}
    ${{ else }}
    thorsten/workspace-discovery-${{ outputs.projectName }}
    ${{ end }}
```

## Details & Edge Cases

* Repositories matching multiple workspace definitions produce an error.
* If a repository is yielded by `on` and **not** matched by an `workspaces.in:` glob, the `steps` will be executed in its root folder.
* If a repository is yielded by `on` and **matches** a `workspace.in:` glob, but there are **no workspaces in it that contain the file in `rootAtLocationOf`** then the `steps` won't be executed in the repository.
* If a repository contains multiple workspaces where one is the subdirectory of another, then the steps will be executed in both and the parent directory also includes the sub directory. I think this is the most intuitive choice.

## Dependency

This requires the addition of `workspaces` to the campaign spec schema, which means it requires changes to the Sourcegraph server.

The PR is here: https://github.com/sourcegraph/sourcegraph/pull/17757

## What's not included

src-cli still downloads a complete archive of every matched repository, even if the steps should only be included in subdirectories. Only downloading archives of the workspace directories is something we should implement in the near future to make support for large monorepos better.

## Full campaign spec to try this at home

There you go:

```yaml
name: workspace-discovery
description: Automatic workspace discovery

on:
  - repositoriesMatchingQuery: repo:automation-testing$
  - repository: github.com/sourcegraph/sourcegraph

workspaces:
  - rootAtLocationOf: go.mod
    in: github.com/sourcegraph/sourcegraph*
  - rootAtLocationOf: package.json
    in: "*automation-testing"

steps:
  - run: "echo \"pwd: $(pwd)\nfiles: $(ls)\" >> message.txt"
    container: alpine:3
  - run:  if [[ -f "package.json" ]]; then cat package.json | jq -j .name; fi
    container: jiapantw/jq-alpine:latest
    outputs:
      projectName:
        value: ${{ step.stdout }}

changesetTemplate:
  title: Automatic workspace discovery
  body:  Automatic workspace discovery

  # If we have an `outputs.projectName` we use it, otherwise we append the path
  # of the workspace. If the path is emtpy (as is the case in the root folder),
  # we ignore it.
  branch: |
    ${{ if eq outputs.projectName "" }}
    ${{ join_if "-" "thorsten/workspace-discovery" (replace steps.path "/" "-") }}
    ${{ else }}
    thorsten/workspace-discovery-${{ outputs.projectName }}
    ${{ end }}
  published: false
  commit:
    message: Automatic workspace discovery
```


cmd/src/campaigns_common.go
```diff
@@ -85,8 +85,8 @@ func newCampaignsApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *ca
 	flagSet.StringVar(&caf.namespace, "n", "", "Alias for -namespace.")
 
 	flagSet.IntVar(
-		&caf.parallelism, "j", 0,
-		"The maximum number of parallel jobs. (Default: GOMAXPROCS.)",
+		&caf.parallelism, "j", runtime.GOMAXPROCS(0),
+		"The maximum number of parallel jobs. Default is GOMAXPROCS.",
```
mrnugget, line 89: Note for reviewers: this shoudln't be strictly part of this PR, but I wanted to clean up something.

cmd/src/campaigns_common.go
```diff
@@ -236,16 +246,11 @@ func campaignsExecute(ctx context.Context, out *output.Output, svc *campaigns.Se
...
 		TempDir:     flags.tempDir,
-	}
-	if flags.parallelism <= 0 {
-		opts.Parallelism = runtime.GOMAXPROCS(0)
-	} else {
-		opts.Parallelism = flags.parallelism
+		Parallelism: flags.parallelism,
 	}
 
-	executor := svc.NewExecutor(opts)
```
mrnugget, line 246: I plan of changing how `*Service` and `*Executor` are instantiated in this function and this was a first step before I realised that doing it would make the PR here super noisy.

internal/campaigns/debug.go
```diff
@@ -0,0 +1,7 @@
+package campaigns
+
+import "github.com/sourcegraph/src-cli/internal/output"
+
+// DebugOut can be used to print debug messages in development to the TUI.
+// For that it needs to be set to an actual *output.Output.
+var DebugOut output.Writer = output.NoopWriter{}
```
mrnugget, line 7: I need this and I'll fight for this to stay in 😛

Why do I need this? Because when I want to debug something my `fmt.Print` output gets eaten by our TUI. With this I can set `campaigns.DebugOut = out` and then `campaigns.DebugOut.Writef("foobar")`. Doesn't always work, but this is a good first step.

internal/campaigns/executor_test.go
```diff
@@ -321,7 +321,8 @@ repository_name=github.com/sourcegraph/src-cli`,
 				}
 
 				for _, r := range tc.repos {
-					executor.AddTask(r, tc.steps, tc.transform, template)
+					params := &Task{Repository: r, Path: ".", Steps: tc.steps, TransformChanges: tc.transform, Template: template}
+					executor.AddTask(params)
 				}
```
mrnugget, line 327: Note to myself: I need to write an integration test for executor.

internal/campaigns/repo_fetcher.go
```diff
@@ -80,16 +113,19 @@ func (rf *repoFetcher) Fetch(ctx context.Context, repo *graphql.Repository) (Rep
...
+	return zip, nil
 }
 
 func (rz *repoZip) Close() error {
-	if rz.fetcher.deleteZips {
+	rz.mu.Lock()
+	defer rz.mu.Unlock()
+
+	rz.references -= 1
+	if rz.references == 0 && rz.fetcher.deleteZips {
```
eseliger, line 125: Wouldn't this mean when using 1 worker thread, there would be no caching at all? and can there be the case where the workers are utilized like the following

```
[repo A:/path1]
[repo B]
[repo C]
[repo D]

=>

[repo A:/path2]
[repo B]
[repo C]
[repo D]
```

where the worker would close the zip for repo A at path1, and then need to refetch for path2?

internal/campaigns/templating.go
```diff
@@ -178,6 +198,18 @@ func (tmplCtx *ChangesetTemplateContext) ToFuncMap() template.FuncMap {
 	return template.FuncMap{
 		"join":  strings.Join,
 		"split": strings.Split,
+		"replace": func(s, old, replacement string) string {
+			return strings.ReplaceAll(s, old, replacement)
+		},
```
eseliger, line 203:
```suggestion
		"replace": strings.ReplaceAll,
```

might that work?

mrnugget:
Heads up reviewers: just noticed that I forgot to add a validation that produces an error and human readable message when multiple changesets with the same branch have been produced in the same repository. I'll add that, but that shouldn't stop anyone from reviewing this.

eseliger (approved):
Looks good to me 🌟 :shipit:

internal/output/null_writer.go
```diff
@@ -0,0 +1,10 @@
+package output
```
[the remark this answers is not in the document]
mrnugget: Nope! Forgot about it, thanks!

internal/campaigns/templating.go
mrnugget: Yes! I also forgot about this.

internal/campaigns/repo_fetcher.go
mrnugget: Nice catch. You're right. I'll address in follow-up PR.

mrnugget:
There are two things that I need to do:

- [ ] Validate and throw an error if multiple changesets have the same branch.
- [ ] Make sure that repository archives are not downloaded multiple times if tasks in the same repository run sequentially and not concurrently.

I'll address those in follow-up PRs to make it easier for review.

PR #443, opened 2021-01-26
Handle ctx.Done() when waiting for steps to finish

mrnugget:
The `.Wait()` method didn't check `ctx.Done()` before and could instead
block on `x.par.Wait()`, waiting for a mutex.

I ran into this just now when my Docker became unresponsive and the
`docker run` in `(*dockerVolumeWorkspace).runScript` would block.

Ctrl-C didn't help and did *not* kill the docker process (why that
didn't happen is a separate question). But src-cli
also didn't exit because it was blocked waiting for that mutex.

So, this solves *another* case of src-cli becoming unresponsive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


eseliger (approved):
nice catch :) 🌟

LawnGnome:
Definitely a good catch! :+1:

PR #448, opened 2021-02-01
Check for branch duplicates after creating changeset specs

mrnugget:
This is a follow-up to #442 and ensures that changeset specs are not
getting silently lost by validating that multiple changeset specs in the
same repository have different branches.

I decided to make this a separate step _after_ the execution of the
steps so that users can leverage the cache. That allows them to change
the campaign spec and then rerun the command after they get this error,
vs. the execution being aborted after running into this error (if we'd
do the check inside executor).

Screenshot:

image

### DIFF
--- a/internal/campaigns/service.go
+++ b/internal/campaigns/service.go
@@ -448,26 +448,22 @@ func (svc *Service) ValidateChangesetSpecs(repos []*graphql.Repository, specs []
 	}
 
 	if len(duplicates) > 0 {
-		return &duplicateBranchErr{fooduplicates: duplicates}
+		return &duplicateBranchesErr{duplicates: duplicates}
 	}
 
 	return nil
 }
 
-type duplicateBranchErr struct {
-	repository *graphql.Repository
-	headRef    string
-	duplicates int
-
-	fooduplicates map[*graphql.Repository]map[string]int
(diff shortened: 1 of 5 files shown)


internal/campaigns/service.go
```diff
@@ -415,6 +415,70 @@ func (svc *Service) ExecuteCampaignSpec(ctx context.Context, opts ExecutorOpts,
...
+			if _, ok := duplicates[r]; !ok {
+				duplicates[r] = make(map[string]int)
+			}
+
+			duplicates[r][branch] = len(specs)
+		}
+	}
+
+	if len(duplicates) > 0 {
+		return &duplicateBranchErr{fooduplicates: duplicates}
```
eseliger, line 451: `fooduplicates`? 😆

eseliger (approved):

internal/campaigns/service.go
LawnGnome: I mean, that _is_ a unique name. :smiley:

LawnGnome (approved):

internal/campaigns/service.go
mrnugget: Argh. I gotta clean up my stuff better.

PR #449, opened 2021-02-01
Keep repo archives cached for later executed tasks

mrnugget:
This is a follow-up to fix the issue discovered by @eseliger here: https://github.com/sourcegraph/src-cli/pull/442#discussion_r566840914

Short version: the previous implementation would only avoid deleting an
archive if there were *currently active tasks* holding references to it.
If tasks that need the same archive would execute sequentially, though,
the archive would be downloaded, deleted, downloaded again.

This here is a fix for the issue by first marking all repository
archives for later use and only once all marks have been turned into
references and those references have been closed is the archive deleted.


mrnugget:
Going with https://github.com/sourcegraph/src-cli/pull/450 after discussing this with @eseliger :)

PR #450, opened 2021-02-01
Alternative approach to keeping archives cached

mrnugget:
This is a second, alterantive approach to https://github.com/sourcegraph/src-cli/pull/449 for keeping archives cached across different executions.

It models `RepoZip` as a [thunk](https://en.wikipedia.org/wiki/Thunk) in a way: `RepoFetcher.Checkout()` returns a `RepoZip` that's passed around with the `Task`. Once all `Tasks` have called `Fetch()` and `Close()` only then the archive is cleaned up.

### DIFF
--- a/internal/campaigns/repo_fetcher.go
+++ b/internal/campaigns/repo_fetcher.go
@@ -18,10 +18,10 @@ import (
 // RepoFetcher abstracts the process of retrieving an archive for the given
 // repository.
 type RepoFetcher interface {
-	// Fetch must retrieve the given repository and return it as a RepoZip.
-	// This will generally imply that the file should be written to a temporary
-	// location on the filesystem.
-	Fetch(context.Context, *graphql.Repository) (RepoZip, error)
+	// Checkout returns a RepoZip for the given repository that's possibly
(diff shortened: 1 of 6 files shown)


internal/campaigns/repo_fetcher.go
```diff
@@ -47,92 +76,82 @@ type RepoZip interface {
...
+	path          string
+	repo          *graphql.Repository
 
-	mu         sync.Mutex
+	client api.Client
+
+	// references is the number of *active* tasks that currently use the archive.
 	references int
+	// marks is the number of tasks that *will* make use of the archive.
+	marks int
```
eseliger, line 95: Not a blocker, but `marks` felt very unintuitive to me, maybe something like `checkouts` or so would be clearer?

eseliger (approved):
LGTM! Thanks for taking this on 👍

internal/campaigns/repo_fetcher.go
mrnugget: Yes! Sorry, that is a leftover.

PR #452, opened 2021-02-02
campaigns: wrap the troubleshooting suggestion

LawnGnome:
I like the troubleshooting suggestion added in #451, but I'd prefer to wrap the output, since it's quite wide right now:

image


mrnugget:
Good stuff!

Issue #453, opened 2021-02-02
src serve-git does not detect repos if CWD is a symlink


mattboehm:
I have `~/repos` symlinked to `~/Dropbox/repos`. If I `cd` into `~/repos` and run `src serve-git`, the server starts and appears to be hosting all of the files, but going to `/v1/list-repos` shows that no repos were detected. If I run the same command from the actual directory, it works and detects all of my repositories.

chrispine:
Thanks for the report, we will look into this.

mrnugget:
@keegancsmith assigning to you for now, since you built `serve-git` and might have a hunch as to what the problem is. Otherwise, we can plan this in the next campaigns sprint.

PR #454, opened 2021-02-03
servegit: eval symlinks on root

keegancsmith:
This is a useful use case which allows changing your root by updating a
symlink. Additionally it is convenient. We have to explicitly resolve
the symlink since filepath.Walk does not support resolving symlinks. We
resolve the symlink each time we crawl so the user can update the value
of the symlink at runtime. Note: we still ignore symlinks with the repos
directory. Supporting that would be more complex due to symlink cycles.

Fixes https://github.com/sourcegraph/src-cli/issues/453


LawnGnome (approved):
LGTM. :+1:

PR #456, opened 2021-02-03
output: hide the cursor when updating

LawnGnome:
I actually like the locking change here on `Output` as well, regardless of the merits of the cursor handling change.

This seems to improve things a bit on Windows and some, but not all, Linux terminal emulators.


mrnugget (approved):
TIL!

eseliger (commented):
🤯

PR #468, opened 2021-02-16
Fix additional files being deleted prematurely

eseliger:
When multiple containers use the same additional files, they could sometimes be deleted before the last container was done using them. Docker bind mount errors were happening then.
This fixes it by taking the workspace path into account for the local file path, making it truly unique between workspace runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,7 +17,7 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
-- Workspaces could sometimes fail set up with docker bind mount errors.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- Workspaces could sometimes fail set up with docker bind mount errors.
```
mrnugget, line 20:
```suggestion
- Workspaces could sometimes fail with docker bind mount errors, due to a race condition of multiple workspaces accessing the same auxilliary files.
```

mrnugget (approved):

PR #470, opened 2021-02-16
Ask user for confirmation before deleting all users

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/17859 by asking the user for confirmation before deleting all users on an instance.


eseliger (approved):
5128

LawnGnome (approved):
Oh jeez. :see_no_evil:

PR #472, opened 2021-02-18
Use Go 1.16 in CI and goreleaser builds

mrnugget:
Newest version of goreleaser seems to require Go 1.16: https://twitter.com/caarlos0/status/1362016990748151812

And since src-cli is rather small and works with Go 1.16, why not?


eseliger (approved):
seems okay to me, as long as it works :D hopefully I don't need to have 2 go versions locally now

LawnGnome (approved):
I don't _think_ that goreleaser change will affect us in the short term, but I also don't see a reason not to do this.

PR #474, opened 2021-02-22
Deprecate old ExecutionCache file formats

mrnugget:
This fixes sourcegraph/sourcegraph#17306 by deleting support for old cache files.

The ticket is slightly wrong, since the previous code did already delete
these files, so we don't have to deprecate and then delete support. We
already transitioned to the new cache file format in 3.24 and 3.25 now.

The next release, 3.26, will only read the new cache format.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Changed
 
+- Deprecated cache file formats are not read by `src campaign [apply|preview]` anymore.
+
 ### Fixed
 
 ### Removed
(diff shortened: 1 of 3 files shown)


eseliger (approved):

LawnGnome:
LGTM, but we did just release this in 3.25.1. I assume we're OK with that?

eseliger:
🤔 🤦‍♂️ 

I think we would be, but it was unintentional... If we do such things, I think we need to have a better process, even if it's just holding back the PR. If someone thinks we aren't, I'll revert and issue a second release.

LawnGnome:
Yeah, I'm not particularly concerned, but we probably shouldn't have merged this if it's 3.26 material, since we don't currently have a process to hold changes back. (I'm not really inclined to create such a process; this is a pretty rare case.)

mrnugget:
Not a huge problem, since it's about caches anyway. The two month deprecation path was incredibly conservative.

Issue #463, opened 2021-02-11
Can't log in to SourceGraph with self-signed certificate


JohnAZoidberg:
I set Sourcegraph up with a self-signed certificate, as described here: https://docs.sourcegraph.com/admin/ssl_https_self_signed_cert_nginx

Now I try to log in with the `src` command but it won't accept it and I can't figure out a way to ignore it. Adding `insecure` to `~/.curlrc` doesn't work.

```
daniel@daniel-virtual-machine:~$ src login https://15.1.1.1:7443
❌ Problem: Error communicating with https://15.1.1.1:7443: Post "https://15.1.1.1:7443/.api/graphql": x509: certificate signed by unknown authority

🛠  To fix: Create an access token at https://15.1.1.1:7443/user/settings/tokens, then set the following environment variables:

   SRC_ENDPOINT=https://15.1.1.1:7443
   SRC_ACCESS_TOKEN=(the access token you just created)

   To verify that it's working, run this command again.

   (If you need to supply custom HTTP request headers, see information about SRC_HEADER_* env vars at https://github.com/sourcegraph/src-cli/blob/main/AUTH_PROXY.md.)
```

Should be easily doable like #239 but I don't feel like compiling Go programs at the moment.

JohnAZoidberg:
Of course I can add the root CA as a system-wide trusted CA but I'd prefer to rather have the `src` command ignore the failure or accept a path to my CA.

mrnugget:
Thanks for reporting @JohnAZoidberg! We'll take a look.

jallenwork:
+1 on this, my teammates using macOS for development can't login to our self signed instance with the `src` utility. We have our certs available in the macOS keychain, but it seems like the `src` tool isn't respecting that?

JohnAZoidberg:
> Of course I can add the root CA as a system-wide trusted CA but I'd prefer to rather have the `src` command ignore the failure or accept a path to my CA.

On my development machine I need to add the cert to my browser to visit the website and make the plugin work. Or I can add it to the system certs.

But I'm opening this issue because I wanted to upload my generated LSIF to sourcegraph from the build server, which doesn't need the cert for any other purpose than this single action.

PR #475, opened 2021-02-22
Add flag to insecurely skip certificate validation

eseliger:
This adds an optional flag to disable certificate validation. Given that, I think it should fix https://github.com/sourcegraph/src-cli/issues/463. I don't know how I feel about this, but it seems to work ok:

image

### DIFF
--- a/internal/api/flags.go
+++ b/internal/api/flags.go
@@ -5,26 +5,29 @@ import "flag"
 // Flags encapsulates the standard flags that should be added to all commands
 // that issue API requests.
 type Flags struct {
-	dump    *bool
-	getCurl *bool
-	trace   *bool
+	dump               *bool
+	getCurl            *bool
+	trace              *bool
+	insecureSkipVerify *bool
 }
 
 // NewFlags instantiates a new Flags structure and attaches flags to the given
 // flag set.
 func NewFlags(flagSet *flag.FlagSet) *Flags {
 	return &Flags{
-		dump:    flagSet.Bool("dump-requests", false, "Log GraphQL requests and responses to stdout"),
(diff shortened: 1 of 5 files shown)


internal/api/flags.go
```diff
@@ -5,26 +5,29 @@ import "flag"
...
 // flag set.
 func NewFlags(flagSet *flag.FlagSet) *Flags {
 	return &Flags{
-		dump:    flagSet.Bool("dump-requests", false, "Log GraphQL requests and responses to stdout"),
-		getCurl: flagSet.Bool("get-curl", false, "Print the curl command for executing this query and exit (WARNING: includes printing your access token!)"),
-		trace:   flagSet.Bool("trace", false, "Log the trace ID for requests. See https://docs.sourcegraph.com/admin/observability/tracing"),
+		dump:               flagSet.Bool("dump-requests", false, "Log GraphQL requests and responses to stdout"),
+		getCurl:            flagSet.Bool("get-curl", false, "Print the curl command for executing this query and exit (WARNING: includes printing your access token!)"),
+		trace:              flagSet.Bool("trace", false, "Log the trace ID for requests. See https://docs.sourcegraph.com/admin/observability/tracing"),
+		insecureSkipVerify: flagSet.Bool("insecure-skip-verify", false, "Skip validating x.509 certificates against trusted chains"),
```
LawnGnome, line 21: Although X.509 is more technically correct, I think we should put TLS in there instead:

```suggestion
		insecureSkipVerify: flagSet.Bool("insecure-skip-verify", false, "Skip validation of TLS certificates against trusted chains"),
```

LawnGnome (approved):
LGTM. I particularly appreciate putting `insecure` in the flag name!

internal/api/flags.go
eseliger: yeah, I was inspired too much by the error message I was trying to overcome, I guess 😁

Issue #478, opened 2021-02-22
campaigns: volume workspace still have issues with non-root execution


LawnGnome:
Found this while testing #477: using volume workspace mode with a non-root user within the step container can result in weird error messages due to potential umask issues on the temporary `run.sh` script that's mounted in, and the default Git configuration within the helper image only applying to the root user.

PR #479, opened 2021-02-23
campaigns: better handle non-root images in volume mode

LawnGnome:
#478 has the details on what's being fixed here. I also realised that we don't use `curl` any more within the volume workspace itself, so we can reduce the size of the helper image.


mrnugget (approved):
Nice catch!

eseliger (approved):

PR #483, opened 2021-03-01
depend on codeintelutils from sourcegraph/sourcegraph

shrouxm:


go.mod
```diff
@@ -21,9 +21,9 @@ require (
 	github.com/pkg/browser v0.0.0-20180916011732-0a3d74bf9ce4
 	github.com/pkg/errors v0.9.1
 	github.com/sourcegraph/campaignutils v0.0.0-20210209103213-27e58c3ce8d3
-	github.com/sourcegraph/codeintelutils v0.0.0-20210118231003-6698e102a8a1
 	github.com/sourcegraph/go-diff v0.6.1
 	github.com/sourcegraph/jsonx v0.0.0-20200629203448-1a936bd500cf
+	github.com/sourcegraph/sourcegraph/enterprise/lib v0.0.0-20210301223818-d8082f761977
```
Strum355, line 26: Do we want this to always point to HEAD via a replace directive pointing to the local path? I feel that would lower friction a lot whereby codeintelutils changes but sg/sg still uses the old remote version

Strum355 (approved):

PR #485, opened 2021-03-04
search: add support for streaming search

stefanhengl:
Stacked on top of #484 

This adds the flag `-stream` to the `search` sub-command. Searches
submitted with `-stream` are routed to the endpoint `search/stream`
instead of the GraphQL api.

In addition, users can set `-display <int>` to set the display limit.

Screenshot 2021-03-04 at 13 43 53
We will add support for JSON in a follow-up PR.

### DIFF
--- a/cmd/src/search.go
+++ b/cmd/src/search.go
@@ -54,11 +54,27 @@ Other tips:
 		explainJSONFlag = flagSet.Bool("explain-json", false, "Explain the JSON output schema and exit.")
 		apiFlags        = api.NewFlags(flagSet)
 		lessFlag        = flagSet.Bool("less", true, "Pipe output to 'less -R' (only if stdout is terminal, and not json flag)")
+		streamFlag      = flagSet.Bool("stream", false, "Consume results as stream.")
+
+		// Streaming.
+		_ = flagSet.Int("display", -1, "Limit the number of results shown. Only supported for streaming.")
 	)
 
 	handler := func(args []string) error {
 		if err := flagSet.Parse(args); err != nil {
 			return err
+		}
+
+		if *streamFlag {
+			// Remove -stream from args.
+			argsWOStream := make([]string, 0, len(args)-1)
+			for _, a := range args {
+				if a == "-stream" {
+					continue
+				}
+				argsWOStream = append(argsWOStream, a)
+			}
+			return streamHandler(argsWOStream)
 		}
 
 		if *explainJSONFlag {
--- a/cmd/src/search.go
+++ b/cmd/src/search.go
@@ -17,6 +17,7 @@ import (
 
 	isatty "github.com/mattn/go-isatty"
 	"github.com/sourcegraph/src-cli/internal/api"
+	"github.com/sourcegraph/src-cli/internal/streaming"
 	"jaytaylor.com/html2text"
 )
 
@@ -50,14 +51,12 @@ Other tips:
 
 	flagSet := flag.NewFlagSet("search", flag.ExitOnError)
 	var (
-		jsonFlag        = flagSet.Bool("json", false, "Whether or not to output results as JSON")
+		jsonFlag        = flagSet.Bool("json", false, "Whether or not to output results as JSON.")
 		explainJSONFlag = flagSet.Bool("explain-json", false, "Explain the JSON output schema and exit.")
 		apiFlags        = api.NewFlags(flagSet)
-		lessFlag        = flagSet.Bool("less", true, "Pipe output to 'less -R' (only if stdout is terminal, and not json flag)")
-		streamFlag      = flagSet.Bool("stream", false, "Consume results as stream.")
-
-		// Streaming.
-		_ = flagSet.Int("display", -1, "Limit the number of results shown. Only supported for streaming.")
+		lessFlag        = flagSet.Bool("less", true, "Pipe output to 'less -R' (only if stdout is terminal, and not json flag).")
+		streamFlag      = flagSet.Bool("stream", false, "Consume results as stream. Streaming search only supports a subset of flags and parameters: trace, insecure-skip-verify, display.")
+		display         = flagSet.Int("display", -1, "Limit the number of results that are displayed. Only supported together with stream flag. Statistics continue to report all results.")
 	)
 
 	handler := func(args []string) error {
@@ -66,15 +65,12 @@ Other tips:
 		}
 
 		if *streamFlag {
-			// Remove -stream from args.
-			argsWOStream := make([]string, 0, len(args)-1)
-			for _, a := range args {
-				if a == "-stream" {
-					continue
-				}
-				argsWOStream = append(argsWOStream, a)
+			opts := streaming.Opts{
+				Display: *display,
+				Trace:   apiFlags.Trace(),
 			}
-			return streamHandler(argsWOStream)
+			client := cfg.apiClient(apiFlags, flagSet.Output())
+			return streamSearch(flagSet.Arg(0), opts, client, os.Stdout)
 		}
 
 		if *explainJSONFlag {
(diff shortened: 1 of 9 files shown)


mrnugget:
I'll take a deeper look, but please add a changelog for these additions :)

internal/api/flags.go
```diff
@@ -31,3 +38,10 @@ func defaultFlags() *Flags {
 		insecureSkipVerify: &d,
 	}
 }
+
+func StreamingFlags(flagSet *flag.FlagSet) *Flags {
+	return &Flags{
+		trace:              flagSet.Bool("trace", false, "Log the trace ID for requests. See https://docs.sourcegraph.com/admin/observability/tracing"),
+		insecureSkipVerify: flagSet.Bool("insecure-skip-verify", false, "Skip validation of TLS certificates against trusted chains"),
+	}
```
mrnugget, line 46: Why the duplication here? The `NewFlags` already returns a `*Flags` with both of these fields.

Edit: And wouldn't parsing fail if someone sets the `get-curl` flag, for example?

cmd/src/search.go
```diff
@@ -54,13 +54,29 @@ Other tips:
...
+			argsWOStream := make([]string, 0, len(args)-1)
+			for _, a := range args {
+				if a == "-stream" {
+					continue
+				}
+				argsWOStream = append(argsWOStream, a)
+			}
+			return streamHandler(argsWOStream)
+		}
+
```
mrnugget, line 79: If you define `-stream` as a noop on the other flagset that you then use in `streamHandler` you wouldn't have to do this here, right?

cmd/src/search_stream.go
```diff
@@ -0,0 +1,413 @@
...
+				ProposedQueries: proposedQueries,
+			})
+			if err != nil {
+				fmt.Printf("ExecuteTemplate: %s\n", err)
+				return
+			}
+		},
+		OnMatches: func(matches []streaming.EventMatch) {
+			for _, match := range matches {
+				if file, ok := match.(*streaming.EventFileMatch); ok {
```
mrnugget, line 101: How about a `switch match := match.(type)` instead of `if` and `continue`?

cmd/src/search_stream.go
```diff
@@ -0,0 +1,413 @@
...
+					err = t.ExecuteTemplate(os.Stdout, "symbol", struct {
+						SourcegraphEndpoint string
+						*streaming.EventSymbolMatch
+					}{
+						SourcegraphEndpoint: cfg.Endpoint,
+						EventSymbolMatch:    symbol,
+					},
+					)
+					if err != nil {
+						fmt.Printf("ExecuteTemplate: %s\n", err)
```
mrnugget, line 157: Why not print all of these error messages stderr? (You can use `flagSet.Output()` as below)

cmd/src/search_stream.go
```diff
@@ -0,0 +1,413 @@
...
+
+	// Send request.
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("error sending request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	// Process response.
+	err = streaming.Decoder{
```
mrnugget, line 65: I think I'd personally pull some of this out to make it more readable. All of these callbacks require only the template (and stderr, see below), so you could use closures to make it flatter:

```go
func newDecoder(t *template.Template) streaming.Decoder {
	return streaming.Decoder{
		OnProgress: onProgress(t),
		// [...]
	}
}

func onProgress(t *template.Template) func(*streaming.Progress) {
	return func(progress *streaming.Progress) {
		// We only show the final progress.
		if !progress.Done {
			return
		}
		err = t.ExecuteTemplate(os.Stdout, "progress", progress)
		if err != nil {
			fmt.Printf("ExecuteTemplate: %s\n", err)
		}
		return
	}
}
```

cmd/src/search_stream.go
```diff
@@ -0,0 +1,413 @@
...
+				}
+
+				if commit, ok := match.(*streaming.EventCommitMatch); ok {
+					err = t.ExecuteTemplate(os.Stdout, "commit", struct {
+						SourcegraphEndpoint string
+						*streaming.EventCommitMatch
+					}{
+						SourcegraphEndpoint: cfg.Endpoint,
+						EventCommitMatch:    commit,
+					})
```
mrnugget, line 139: I think you could clean up this render code a bit if you introduce something like a `Context` struct that has all of the fields defined but on which you only set the current struct.

mrnugget (approved):
Approved but left a few comments, because I think the code can be made cleaner and easier to understand/maintain.

Also: no tests? 😬

cmd/src/search.go
stefanhengl: yes, but then we mess with the auto-documentation. Suddenly the "streaming search" has a "-stream" flag which is odd. I traded odd code against odd documentation.

internal/api/flags.go
stefanhengl: > Edit: And wouldn't parsing fail if someone sets the get-curl flag, for example?
 
parsing only fails if someone calls `search` with `-stream` AND `-get-curl`, which is intended.  `-curl` and `-dump` make a lot less sense for streaming than for the GraphQL API. I assume the use case for `get-curl` is to pipe the output to something like `jq`? The streaming endpoint does not return valid JSON, but a stream of events that contain JSON.

stefanhengl:
> Also: no tests? 😬

The PRs are stacked. You reviewed before I had the chance to push the 3rd PR with tests :-).

cmd/src/search_stream.go
stefanhengl: Yeah, I also went back and forth about this one. I think the current approach is pretty straightforward although maybe a bit verbose. If we use a common context struct then the templates get more difficult to read. I will see if I can find something better.

cmd/src/search.go
mrnugget: Why not create a separate `stream-search` command?

stefanhengl: It is based on a discussion we had around exhaustive search which is documented in [RFC 288](https://docs.google.com/document/d/1dk309wEXA34b-LF66SBOzcNBJ6jZbWXT_eeBL9TZ2Ro/edit). At some point we want phase out -stream and replace search with streaming search. That is part of the reason why I wrote it as handler.

internal/api/flags.go
keegancsmith: we could make get curl work though? My original streaming client was curl :D But yeah, until we have stabalized the API properly I don't want people to start writing there own clients.

stefanhengl:
@keegancsmith PTAL. Tests are now integrated in this PR.

cmd/src/search.go
keegancsmith: This is kind of weird though:

```shellsession
$ ./src search -stream -get-curl foo
flag provided but not defined: -get-curl
Usage of streaming search:
  -display int
    	Limit the number of results that are displayed. Note that the statistics continue to report all results. (default -1)
  -insecure-skip-verify
    	Skip validation of TLS certificates against trusted chains
  -trace
    	Log the trace ID for requests. See https://docs.sourcegraph.com/admin/observability/tracing
```

Given we intend on replacing search with stream search in the future, can't we make the stream search stuff not use its own flagset. I would think an approach like

```go
streaming.Search(streaming.Opts{
  Query: flagSet.Args(),
  Trace: *trace,
  Display: *display,
  InsecureSkipVerify: *insecureSkipVerify,
})
```

would make sense. Or even you pass in the http client you use that hides the trace/insecure stuff.

That way you can also move the large amounts of business logic out of `./cmd/src` into `./internal/streaming`

keegancsmith (approved):

cmd/src/search.go
stefanhengl: I actually like this behavior, but since both reviewers commented on this I may be on the track here. I refactored it to be a simple function call instead of a handler and moved parts of the logic to internal/streaming.

PR #486, opened 2021-03-04
search: add tests for search_streaming.go

stefanhengl:
Stacked on top of #485 

Note: This adds a new dependency on [autogold](https://github.com/hexops/autogold) to src-cli


cmd/src/testdata/TestCommitTemplate.golden
```diff
@@ -0,0 +1,6 @@
+`[38;5;239m([0m[38;5;237mhttps://sourcegraph.com[0m[38;5;239m)
+[0m[0m[38;5;68msourcegraph/sourcegraph-atom > Stephen Gutekanst : all: release v1.0.5[0m[38;5;2m (1 match)[0m
+[38;5;239m--------------------------------------------------------------------------------
+[0m[38;5;239m[0m  oo [38;5;0m[48;5;11mbar[0m
+
+`
```
keegancsmith, line 6: Using a gostring looks a bit ugly here. Can we rather use `autogold.Raw`?

go.sum
```diff
@@ -1,6 +1,5 @@
 github.com/Masterminds/semver v1.5.0 h1:H65muMkzWKEuNDnfl9d70GUjFniHKHRbFPGBuZ3QEww=
```
keegancsmith, line 1: I'm surprised at the number of transitive dependencies autogold is adding. The use case you have here is literally just a few lines of go.

cmd/src/search_stream_test.go
```diff
@@ -0,0 +1,198 @@
+package main
```
keegancsmith, line 1: A different way to test this would be to start a streaming server and return your mocked events, then you can just diff against the total output of src-cli search -stream. That should cover a lot more code paths and be a test that is less likely to change as you change internal details.

keegancsmith (approved):

go.sum
stefanhengl: Yeah, I was surprised, too. I think the use case for autogold is perfect though. It is really snapshotting the UI.

cmd/src/search_stream_test.go
stefanhengl: Sure. Let's remove autogold then. Doing both, spinning up a stream server and using autogold is an overkill.

stefanhengl:
closing this one in favor of #485. Initially I wanted to use autogold for the tests, however based on the comments here I reimplemented the tests without adding new dependencies while at the same time increasing the coverage.

PR #489, opened 2021-03-09
s/campaign/batch change/

LawnGnome:
This is _fairly_ mechanical, although I actually did most of it by hand. The only bit that's really interesting is that I needed to abstract two GraphQL mutations to maintain cross-version compatibility: I did so by introducing a new `Operations` interface in `internal/batchers/graphql` that delegates to an appropriate implementation based on our existing "feature flag" quirk detection, and takes that level of detail out of `internal/batchers/Service`.

We're also renaming the volume workspace Docker image: this is fine, since previous tagged versions will continue to just work. The release automation will be mildly spicy when we do the next `src` release, as there's only so much one can bench test without actually invoking GitHub Actions; it may be worth cutting a release as soon as we have the documentation up to date while this is (relatively) fresh in my mind, even though it will lift the veil on the great renaming slightly early. (Since everything is aliased, nothing should break in practice.)

This is paired with https://github.com/sourcegraph/sourcegraph/pull/18963, which includes the schema changes required to make this work. Those schema changes are _not_ breaking, but we should nevertheless keep both repos in sync for (I hope) obvious reasons.

Closes sourcegraph/sourcegraph#18940.


.github/workflows/docker.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         with:
```
eseliger, line 28: Is this auto-publishing?

eseliger (approved):
Nice approach with the two backends, LGTM 🌟

cmd/src/batch.go
```diff
@@ -0,0 +1,52 @@
...
+	usage := `'src batch' manages batch changes on a Sourcegraph instance.
+
+Usage:
+
+	src batch command [command options]
+
+The commands are:
+
+	apply                 applies a batch spec to create or update a batch
+						  change
```
mrnugget, line 20:
```suggestion
	apply                 applies a batch spec to create or update a batch change
```

cmd/src/batch.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// Register the command.
+	commands = append(commands, &command{
+		flagSet: flagSet,
+		aliases: []string{
+			"batchchange",
+			"batch-change",
+			"batchchanges",
+			"batch-changes",
+			"batches",
```
mrnugget, line 45: Better safe than sorry!

mrnugget (approved):

.github/workflows/docker.yml
LawnGnome: Every release, yes.

cmd/src/batch.go
LawnGnome: It has to be wrapped because it goes past 80 characters, but I messed up the whitespace preceding the word, so I'll fix that.

cmd/src/batch.go
LawnGnome: I was starting to wonder if I should start adding translations into other major languages. :laughing:

Issue #487, opened 2021-03-05
Add `campaigns.name` to changeset templating


malomarrec:
In the context of creating issues along with campaigns, I'd like to be able to get the campaigns name from within a run.

chrispine:
`batch.name`?

malomarrec:
So I'm writing an [example](https://github.com/sourcegraph/campaign-examples/tree/master/github-issues) of how to create GitHub issues, along with changes, in a declarative way.

There is a container that takes care of checking for existing issues, syncing them, and creating new ones.

Two things would be helpful to add to templating:
- `batch.name` : passing the name of the campaigns to the step
- `batch.run_type` : passing whether this is a "preview" or an "apply" to the step

That way I can make it so that my script (1) links to the campaign name, and reuse that as a namespace for tickets (2) create tickets only if "batch.run_type == 'apply'" and skip previews.

malomarrec:
The only think I'd miss after that is a 
`runAfterApplyCompletes` step (or `applyComplete` hook), to trigger a job that will link issues with PRs

mrnugget:
`batch.run_type` doesn't work because while it's executing and while the `changesetTemplate` is rendered it's still undecided what will happen: the fully rendered batch spec and the changeset specs are uploaded in any case. `apply` or `preview` only affects what happens after and _not_ the content of the specs themselves.

malomarrec:
I see. Then the only way to access this would be through events hooks, I guess.

PR #491, opened 2021-03-10
Add batch_change.{name,description} as templating variables

mrnugget:
This fixes https://github.com/sourcegraph/src-cli/issues/487 by adding `batch_change.name` and `batch_change.description` to the templating vars.

I didn't want to use `batch` because it's ambigious, but I'm also not 100% sure about `batch_change` since _technically_ the batch change doesn't exist yet. So maybe it should be `batch_spec`? But that's also not 100% clear cut.

**For code reviewers**: note that I introduced a little helper struct. That might seem overblown, but while writing the code I realized that I'm about to pass the whole `BatchSpec` around if I don't add this helper thing. And if I were to pass the `*BatchSpec` to every `*Task` and every `runSteps` then we could get rid of all the intermediate steps, but _then_ everything would need to dig deep into the `*BatchSpec`. If you think that's fine I can change it. I just thought that for now it's easier and safer to simply pass two duplicated strings around.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Added
 
+- Two new [templating](https://docs.sourcegraph.com/campaigns/references/batch_spec_templating) variables have been added: `batch_change.name` and `batch_change.description`.
+
 ### Changed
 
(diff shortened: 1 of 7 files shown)


malomarrec:
`batch_change` seems clear to me, because the intent is to create a batch change, even though it doesn't really exist yet.

When I create a batch change, and set a name, I am using a name from the batch change namespace. So it's clear to me what batch_change means here

mrnugget:
I'll follow up with docs to `sourcegraph/sourcegraph`

PR #494, opened 2021-03-15
proof-of-concept search select for batch changes

rvantonder:
Disclaimer: _this is completely untested / unvalidated, but I'm putting up a proof-of-concept rather than writing a slack thread about it_

Batch changes will benefit from using the new `select:` parameter for filtering repos/files. The main reasons are:

- `select:` automatically deduplicates repos and files (no need to send everything over the wire for it to get deduplicated here)
- For `select:repo` specifically, we added an optimization that will stop searching repos if one result is found (we don't continue searching for `count:99999` matches). This "should" be faster.
- `select:` has a more natural behavior for describing searches for batch changes compared to `type:` ( `select:` only affects result types, `type:` may interpret inputs in various ways).

You'll have to try this out, maybe like a beta? I'm not sure how this part is tested, but it should be a worthwhile update. Please let me know if you do try it out and something seems wrong.

Feel free to close this PR whenever, it's meant for visibility and discussion.

### DIFF
--- a/internal/batches/service.go
+++ b/internal/batches/service.go
@@ -658,12 +658,6 @@ query ChangesetRepos(
                 ... on Repository {
                     ...repositoryFields
                 }
-                ... on FileMatch {
-                    file { path }
-                    repository {
-                        ...repositoryFields
-                    }
-                }
             }
         }
     }
@@ -679,26 +673,16 @@ func (svc *Service) resolveRepositorySearch(ctx context.Context, query string) (
 		}
 	}
 	if ok, err := svc.client.NewRequest(repositorySearchQuery, map[string]interface{}{
-		"query":       setDefaultQueryCount(query),
+		"query":       setDefaultQueryCount(query) + " select:repo",
 		"queryCommit": false,
 		"rev":         "",
 	}).Do(ctx, &result); err != nil || !ok {
 		return nil, err
 	}
 
-	ids := map[string]*graphql.Repository{}
 	var repos []*graphql.Repository
 	for _, r := range result.Search.Results.Results {
-		existing, ok := ids[r.ID]
-		if !ok {
-			repo := r.Repository
-			repos = append(repos, &repo)
-			ids[r.ID] = &repo
-		} else {
-			for file := range r.FileMatches {
-				existing.FileMatches[file] = true
-			}
-		}
+		repos = append(repos, &r.Repository)
 	}
 	return repos, nil
 }
@@ -758,7 +742,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 		a.WriteString("query DirectoriesContainingFile() {\n")
 
 		for _, repo := range batch {
-			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$ type:path count:99999`, regexp.QuoteMeta(fileName), regexp.QuoteMeta(repo.Name))
+			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$ select:file count:99999`, regexp.QuoteMeta(fileName), regexp.QuoteMeta(repo.Name))
 
 			a.WriteString(fmt.Sprintf(searchQueryTmpl, queryIDByRepo[repo], query))
 		}


internal/batches/service.go
```diff
@@ -679,26 +673,16 @@ func (svc *Service) resolveRepositorySearch(ctx context.Context, query string) (
 		}
 	}
 	if ok, err := svc.client.NewRequest(repositorySearchQuery, map[string]interface{}{
-		"query":       setDefaultQueryCount(query),
+		"query":       setDefaultQueryCount(query) + " select:repo",
```
rvantonder, line 676: Here it should be as simple as adding `select:repo` (yuck for doing it like this in the string but OK). Seems like file matches are only used for deduping, in which case, all that related code can be removed.

internal/batches/service.go
```diff
@@ -758,7 +742,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 		a.WriteString("query DirectoriesContainingFile() {\n")
 
 		for _, repo := range batch {
-			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$ type:path count:99999`, regexp.QuoteMeta(fileName), regexp.QuoteMeta(repo.Name))
+			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$ select:file count:99999`, regexp.QuoteMeta(fileName), regexp.QuoteMeta(repo.Name))
```
rvantonder, line 745: This change should be functionally equivalent. `type:` is problematic because it applies to both an input parameter (e.g., matches files corresponding to the pattern `file:`, or if `file:` were not specified, some pattern) and then also  returns result type `file`. `select:` strictly applies to return type, which is the desired behavior here, so we should prefer that. In time `type:` will be deprecated in favor of more predictable parameters like `select`.

rvantonder:
(Don't actually review, just putting it on your radar)

rvantonder:
> === RUN   TestResolveRepositorySearch
    service_test.go:53: wrong number of repos. want=3, have=5
--- FAIL: TestResolveRepositorySearch (0.00s)
=== RUN   TestResolveRepositories_Unsupported
=== RUN   TestResolveRepositories_Unsupported/allowUnsupported:true
    service_test.go:154: wrong number of repos. want=4, have=1
=== RUN   TestResolveRepositories_Unsupported/allowUnsupported:false
    service_test.go:170: wrong number of repos. want=4, have=0
--- FAIL: TestResolveRepositories_Unsupported (0.00s)
    --- FAIL: TestResolveRepositories_Unsupported/allowUnsupported:true (0.00s)
    --- FAIL: TestResolveRepositories_Unsupported/allowUnsupported:false (0.00s)

Interesting 😄 . Taking a quick look at this test.

rvantonder:
Ah, looks like these tests fail because it's effectively testing or relying that the dedupe logic works, which I removed.

eseliger:
Thanks @rvantonder! I will pick this up hopefully today :) Very nice that we can finally do that!

mrnugget:
@eseliger You worked on this, right? Can you give a short update on what the current state is? (Asking because the "review requested" notification for this PR shows up for me every day 😬 )

eseliger:
Ah yes, so the current status is this doesn't work as-is, because we need the file results per repo, too. They are used in templating.
It is still a good change and we should do it, but it requires some more work to get done.
In numbers: A spec that just queried lang:java, k8s.sgdev.org would previously crash occasionally computing the result and consume loads of memory, also it took 65s to resolve, with the select:repo filter, it only takes about 2 seconds. 
So to get this to work: We should switch to using the `select:repo` filter, (with a feature gate, because older sourcegraph versions don't support this keyword), and we then need to add lazy loading of file results per-repo, if needed. The first part I worked on, the second we still need to handle. I think it's 1-2d of work to get this over the finish line.

rvantonder:
Chipping in to say, after talking to @eseliger, yeah, it is a good idea to decouple:

- fetch only repos (e.g., using `select:repo`)
- fetching file/contents for templating (using whatever other mechanism)

This primarily because you can optimize for the first case if a spec doesn't use templating. There may be other tradeoffs too. Inspecting the spec to decide what to do is probably a good framework in general to consider tradeoffs.

mrnugget:
Thanks for the overview!

It sounds to me like we'd have a performance regression then as long as we don't check whether templating is used, right?

Because right now we have a single request to fetch the repositories and the file matches. If we decouple that we need at least 2 requests, right?

rvantonder:
> If we decouple that we need at least 2 requests, right?

I used "decoupling" loosely, didn't mean for it to imply decomposing the current state into two requests when templating is in effect. Instead, basically have:

- if templating used -> do current thing (fallback, default)
- if templating absent -> use `select:repo` (opportunity to optimize, introduce a new code path or GQL query).

rvantonder:
This PR doesn't do that, so it's not a good example to work off of any longer--when I wrote it I didn't take these concerns into account. Thinking through it with @eseliger cleared up some of that. At the same time, talking to him about this shows that there are real benefits to doing it (i.e. "worth it"):

>  A spec that just queried lang:java, k8s.sgdev.org would previously crash occasionally computing the result and consume loads of memory, also it took 65s to resolve, with the select:repo filter, it only takes about 2 seconds.

mrnugget:
> Instead, basically have:
> 
> * if templating used -> do current thing (fallback, default)
> * if templating absent -> use `select:repo` (opportunity to optimize, introduce a new code path or GQL query).

Ah, that makes sense! Thanks.

My immediate reaction to "check whether templating is used" was "oh, this is going to be nasty", but it might just be as easy as checking whether both templating delimiters are present in a given string.

rvantonder:
I would just parse it as a template according to whatever syntax we currently support and and a traversal to inspect if there are template nodes. I'm not super familiar with the templating structures but it looks like "anything other than [`NodeText`](https://golang.org/pkg/text/template/parse/#NodeType)" would be a strong enough condition to imply "templating is in effect".

mrnugget:
> I'm not super familiar with the templating structures but it looks like "anything other than [`NodeText`](https://golang.org/pkg/text/template/parse/#NodeType)" would be a strong enough condition to imply "templating is in effect".

Yep, that seems to work:

```go
func usesTemplating(in string) bool {
	temp := template.Must(template.New("cool-template").Delims("${{", "}}").Parse(in))
	for _, n := range temp.Tree.Root.Nodes {
		if n.Type() != parse.NodeText {
			return true
		}
	}
	return false
}

func TestUsesTemplating(t *testing.T) {
	tests := []struct {
		name  string
		input string
		want  bool
	}{
		{name: "no templating", input: "foobar", want: false},
		{name: "only templating", input: "${{ .CoolVar }}", want: true},
		{name: "text and templating", input: "foobar and ${{ .CoolVar }}", want: true},
	}

	for _, tc := range tests {
		t.Run(tc.name, func(t *testing.T) {
			have := usesTemplating(tc.input)
			if have != tc.want {
				t.Errorf("wrong result. want=%t, have=%t", tc.want, have)
			}
		})
	}
}

```

eseliger:
I pushed a branch here with the backcompat changes: https://github.com/sourcegraph/src-cli/pull/499/files
The only thing missing would be your logic to determine templating is used and then fetching the results once the changeset in that repo is executed.
I don't think it's a performance regression, since we can now make things asynchronous:
We don't need to wait for all results to be computed on a 500k repos instance that returns a few million results for `lang:java`, but we can start right away once we know which _repos_ qualify, and then start the first executions, fetch the file results for just that repo and start processing. I think this will be much faster

mrnugget:
Closing this to get some peace in my GitHub notifications and because we have a PoC in another PR that we need to work on.

PR #497, opened 2021-03-17
Swapping latest vs. instance

emchap:
A swap as promised! Figured we should just always route people to the compatible version unless they work here.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -15,9 +15,18 @@
 
 ## Installation
 
-Binary downloads are available on the [releases tab](https://github.com/sourcegraph/src-cli/releases), and through Sourcegraph.com. _If the latest version does not work for you,_ consider using the version compatible with your Sourcegraph instance instead.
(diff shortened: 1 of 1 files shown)


LawnGnome:
Thanks, @emchap!

Just for future people who come across this: the rationale behind closing this is that we're comfortable with suggesting to users that they use the latest version regardless of their Sourcegraph instance: in general, we support that case as much as we support running the exact recommended version, as `src` is designed for backward compatibility.

mrnugget:
Should we then also update the [Quickstart for Batch Changes](https://docs.sourcegraph.com/batch_changes/quickstart#install-the-sourcegraph-cli)? Or is this something different here?

emchap:
@mrnugget We should bring them in alignment with each other, yeah. (That functionality is why I use src-cli, though I didn't realize it was documented there!) Would it make sense to link to the readme so we only have to document in one spot?

mrnugget:
> Would it make sense to link to the readme so we only have to document in one spot?

Normally I'd say yes, but I do like that everything's contained in the quickfix guide and you can just copy/paste as you go along.

PR #498, opened 2021-03-23
More user-friendly error message if batch change name is invalid

mrnugget:
This fixes https://github.com/sourcegraph/sourcegraph/issues/16955.

Preview:

user-friendly-message

### DIFF
--- a/internal/batches/batch_spec.go
+++ b/internal/batches/batch_spec.go
@@ -2,6 +2,7 @@ package batches
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/gobwas/glob"
 	"github.com/hashicorp/go-multierror"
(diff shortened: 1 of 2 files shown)


eseliger (approved):
neat idea 😁

internal/batches/batch_spec.go
```diff
@@ -107,6 +108,21 @@ type Group struct {
 func ParseBatchSpec(data []byte, features featureFlags) (*BatchSpec, error) {
 	var spec BatchSpec
 	if err := yaml.UnmarshalValidate(schema.BatchSpecJSON, data, &spec); err != nil {
+		if multiErr, ok := err.(*multierror.Error); ok {
+			var newMultiError *multierror.Error
+
+			for _, e := range multiErr.Errors {
+				// In case of `name` we try to make the error message more user-friendly.
+				if strings.Contains(e.Error(), "name: Does not match pattern") {
```
chrispine, line 116: If the error message changes sometime in the future, this will stop working. Would it make sense to add a test to catch that kind of regression?

chrispine (approved):
LGTM, except maybe we should have a test, since I could see this regressing pretty easily and silently.

LawnGnome (approved):
:label:

internal/batches/batch_spec.go
mrnugget: Good point! I was unsure about to test this (or whether I should) but a tiny test that makes sure this behaviour stays sounds good to me.

PR #500, opened 2021-03-25
Add gitHead to extension manifest

efritz:


cmd/src/extensions_publish.go
```diff
@@ -87,12 +94,32 @@ Notes:
...
+			if err != nil {
+				return err
+			}
+		} else {
+			command := exec.Command("git", "rev-parse", "head")
+			command.Dir = manifestDir
+
+			out, err := command.CombinedOutput()
+			if err != nil {
+				return fmt.Errorf("failed to determine git head: %q", err)
```
felixfbecker, line 119: Maybe we should ignore this error (not add the field in that case) so that it's possible to publish an extension without having a git repo (in a simple folder) or before making the initial commit

felixfbecker (approved):

cmd/src/extensions_publish.go
efritz: If we rely on this for troubleshooting I'd rather make sure we have it or else require the user to explicitly ignore it on publish (they can supply a `-git-head 'canned string'` on the command line to do this).

Thoughts?

mrnugget (approved):
Code LGTM, no opinions on whether to make this a required thing for extensions or not.

cmd/src/extensions_publish.go
eseliger: Should we add your suggestion in this comment to the error message then?

eseliger (approved):

cmd/src/extensions_publish.go
felixfbecker: Here are some arguments for ignoring the error:

- It's how `npm publish` behaves
- It would be a breaking change
- We don't require the `repository` field in package.json either afaik, so it's a bit weird to require the commit, but not the repository
- I'm worried that it increases the barrier to create extensions
- If the goal is to have this information primarily for code intel and our own extensions, we can pretty much guarantee that they always will be in a repo and have that info
  - just not for customer's private extensions (for which the commit is not useful to us anyway)
- We could still print a warning

I don't care strongly about this though so feel free to pick either behavior and merge that.

PR #507, opened 2021-04-05
Fix broken link for batch changes docs

KeisukeYamashita:
## What

Fix broken links for batch changes.

## Why

Better document maintenability.


chrispine (approved):
Thanks!

PR #508, opened 2021-04-06
Use select:repo to speed up repository resolution

eseliger:
This PR adopts select:repo to speed up initial resolution of matching repositories in batch changes, if the Sourcegraph instance supports it. To fetch file matches per repository, we each out to the server again to get the file results, per repo. This is significantly faster than waiting for all results for all repos to be retrieved.

Co-authored-by: Rijnard van Tonder <rvantonder@gmail.com>

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -272,7 +272,7 @@ func batchExecute(ctx context.Context, out *output.Output, svc *batches.Service,
 		Parallelism: flags.parallelism,
 		FetchFileMatches: func(repo *graphql.Repository) (map[string]bool, error) {
 			fileResults := make(map[string]bool)
-			for _, onQuery := range repo.IncludedInSearchQueries {
+			for onQuery := range repo.IncludedInSearchQueries {
 				r, err := svc.FetchFileResultsForRepo(ctx, repo, onQuery)
 				if err != nil {
 					return nil, err
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -270,6 +270,19 @@ func batchExecute(ctx context.Context, out *output.Output, svc *batches.Service,
 		Timeout:     flags.timeout,
 		TempDir:     flags.tempDir,
 		Parallelism: flags.parallelism,
+		FetchFileMatches: func(repo *graphql.Repository) (map[string]bool, error) {
+			fileResults := make(map[string]bool)
+			for _, onQuery := range repo.IncludedInSearchQueries {
+				r, err := svc.FetchFileResultsForRepo(ctx, repo, onQuery)
+				if err != nil {
+					return nil, err
+				}
+				for k := range r {
+					fileResults[k] = true
+				}
+			}
+			return fileResults, nil
+		},
 	}
 
 	p := newBatchProgressPrinter(out, *verbose, flags.parallelism)
--- a/internal/batches/executor.go
+++ b/internal/batches/executor.go
@@ -362,9 +362,15 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		x.updateTaskStatus(task, func(status *TaskStatus) {
 			status.CurrentlyExecuting = "Resolving file matches..."
 		})
-		task.Repository.FileMatches, err = x.FetchFileMatches(task.Repository)
+		req, err := requiresSearchResultPaths(task)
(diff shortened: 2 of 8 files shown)


cmd/src/batch_common.go
```diff
@@ -270,6 +270,19 @@ func batchExecute(ctx context.Context, out *output.Output, svc *batches.Service,
 		Timeout:     flags.timeout,
 		TempDir:     flags.tempDir,
 		Parallelism: flags.parallelism,
+		FetchFileMatches: func(repo *graphql.Repository) (map[string]bool, error) {
```
mrnugget, line 273: Can we make this a method on `svc` and pass that in instead, like `FetchFileMatches: svc.FetchFileMatches`?

(We've talked about this on Zoom, but I don't like how here all the layers bleed into each other: the command-layer passing things from the service into the executor, but it's also not _too bad_ as to be a blocker, IMHO, so just leaving this as a note)

internal/batches/executor.go
```diff
@@ -358,6 +358,19 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		log.Close()
 	}()
 
+	if x.features.selectSearchOperator {
```
mrnugget, line 361: Doesn't this also need the check for whether the `steps` actually use templating?

internal/batches/executor.go
```diff
@@ -358,6 +358,19 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
...
+		x.updateTaskStatus(task, func(status *TaskStatus) {
+			status.CurrentlyExecuting = "Resolving file matches..."
+		})
+		task.Repository.FileMatches, err = x.FetchFileMatches(task.Repository)
```
mrnugget, line 365: I assume `task.Repository` is a clone for each task, right? Otherwise we'd have a race condition

internal/batches/service.go
```diff
@@ -678,25 +686,47 @@ func (svc *Service) resolveRepositorySearch(ctx context.Context, query string) (
...
-		} else {
-			for file := range r.FileMatches {
-				existing.FileMatches[file] = true
+	if svc.features.selectSearchOperator {
+		repos = make([]*graphql.Repository, 0, len(result.Search.Results.Results))
+		for _, r := range result.Search.Results.Results {
+			if r.Repository.IncludedInSearchQueries == nil {
+				r.Repository.IncludedInSearchQueries = make(map[string]bool)
+			}
+			r.Repository.IncludedInSearchQueries[queryWithDefaultCount] = true
```
mrnugget, line 713: But if you use `queryWithDefaultCount` here, doesn't that mean that not all file matches will be fetched later on? Before this PR we'd fetch all the FileMatches we could get, but with this change we'd use the default count per repository, right?

eseliger: The "default count" is `9999999` 😬  it's the default for batch changes, not sourcegraph search default

mrnugget: Ahhh! `queryWithCount` might be a better name then (I now see though that the name comes from `setDefaultQueryCount`)

LawnGnome (commented):
I did look at this, but Thorsten basically got in with everything I would have asked or said. (And, you know, more.)

Generally LGTM, pending the answers to those points!

chrispine:
This one has been sitting here for a while. What's the status?

mrnugget:
@eseliger can correct me if I'm wrong, but when we last talked I think the outcome was that this is slightly more complicated than we thought it would be, because we'd need to traverse the rest of the batch spec to check whether `repository.search_result_paths` is accessed. With the tooling we now have in place for partial evaluation of batch specs that's definitely easier to do now than back then.

But the PR still needs more work than simply rebasing.

This _could_ be an opportunity to extract this logic into something like a `ResolveRepositories(batchSpec string) ([]*graphql.Repository, error)` function that can then also be used by SSBC.

eseliger:
I think you're right. This needs some more work, but is _definitely_ worth it, it improves startup performance significantly and reduces load on the server too. 

That _could_ work but it's not _that_ simple, because we need to implement two `searcher` interfaces, one that uses the graphql API and one that uses the internal search APIs, I guess?

eseliger:
Will be done as part of https://github.com/sourcegraph/sourcegraph/pull/24222

PR #509, opened 2021-04-08
Filter out repositories with .batchignore at root

mrnugget:
This implements https://github.com/sourcegraph/sourcegraph/issues/18330 by querying the locations of `.batchignore` files in each repository yielded by the `on` attribute in a batch spec. If locations were found, the repository is ignored.

This can be overwritten by using the `-force-override-ignore` flag.

Example: given my instance has the following repositories:

- `github.com/sourcegraph-testing/zap`
- `github.com/sourcegraph-testing/titan`
- `github.com/sourcegraph-testing/tidb`
- `github.com/sourcegraph-testing/etcd`
- `github.com/sourcegraph-testing/batch-changes-testing-ignore`

And I use the following batch spec:

```yaml
on:
  - repositoriesMatchingQuery: repohasfile:README.md repo:sourcegraph-testing

steps:
  - run: echo "a horse says 'hello'" >> README.md
    container: alpine:3
```

with this change the `batch-changes-testing-ignore` repository will be ignored:

- no archive will be downloaded
- no steps executed

A message is printed that says it's ignored.

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -31,6 +31,7 @@ var (
 
 type batchApplyFlags struct {
 	allowUnsupported bool
+	allowIgnored     bool
 	api              *api.Flags
 	apply            bool
 	cacheDir         string
@@ -54,6 +55,10 @@ func newBatchApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *batchA
 	flagSet.BoolVar(
 		&caf.allowUnsupported, "allow-unsupported", false,
 		"Allow unsupported code hosts.",
+	)
+	flagSet.BoolVar(
+		&caf.allowIgnored, "no-ignore", false,
+		"Do not ignore repositories that have a .batchignore file.",
 	)
 	flagSet.BoolVar(
 		&caf.apply, "apply", false,
@@ -234,6 +239,14 @@ func batchExecute(ctx context.Context, out *output.Output, svc *batches.Service,
 			batchCompletePending(pending, "Resolved repositories")
 
 			block := out.Block(output.Line(" ", output.StyleWarning, "Some repositories are hosted on unsupported code hosts and will be skipped. Use the -allow-unsupported flag to avoid skipping them."))
+			for repo := range repoSet {
+				block.Write(repo.Name)
+			}
+			block.Close()
+		} else if repoSet, ok := err.(batches.IgnoredRepoSet); ok {
+			batchCompletePending(pending, "Resolved repositories")
+
+			block := out.Block(output.Line(" ", output.StyleWarning, "Some repositories contain .batchignore files and will be skipped. Use the -no-ignore flag to avoid skipping them."))
 			for repo := range repoSet {
 				block.Write(repo.Name)
 			}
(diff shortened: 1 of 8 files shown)


mrnugget:
Closing this for now, since I discussed this with @malomarrec on a Zoom call right now. Outcome: we don't want to exclude those repositories from the search results, because we want to print which repositories were ignored. See https://github.com/sourcegraph/sourcegraph/issues/18330#issuecomment-815752987

mrnugget:
Turns out that the functionality we built for monorepo support — a method that returns the locations of a file in a list of repositories — can be used here: instead of adding a `-repohasfile:.batchignore` filter to the search query I changed the code to search the locations of `.batchignore` files in all the yielded repositories.

https://user-images.githubusercontent.com/1185253/114032062-3fe0dd00-987c-11eb-8a4f-1fef4d264ca3.mp4

cmd/src/batch_common.go
```diff
@@ -55,6 +56,10 @@ func newBatchApplyFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *batchA
 		&caf.allowUnsupported, "allow-unsupported", false,
 		"Allow unsupported code hosts.",
 	)
+	flagSet.BoolVar(
+		&caf.allowIgnored, "no-ignore", false,
```
LawnGnome, line 60: We might want to make the name of the flag a bit more forceful (pun somewhat intended), just to reinforce that it probably shouldn't be used normally. `-force-override-ignore`, or something like that?

cmd/src/batch_common.go
```diff
@@ -238,6 +243,14 @@ func batchExecute(ctx context.Context, out *output.Output, svc *batches.Service,
 				block.Write(repo.Name)
 			}
 			block.Close()
+		} else if repoSet, ok := err.(batches.IgnoredRepoSet); ok {
+			batchCompletePending(pending, "Resolved repositories")
+
+			block := out.Block(output.Line(" ", output.StyleWarning, "Some repositories contain .batchignore files and will be skipped. Use the -no-ignore flag to avoid skipping them."))
```
LawnGnome, line 249: Maybe a slight wording tweak, since this implies that you just get an opaque set of repositories, but then we print them:

```suggestion
			block := out.Block(output.Line(" ", output.StyleWarning, "The repositories listed below contain .batchignore files and will be skipped. Use the -no-ignore flag to avoid skipping them."))
```

LawnGnome (approved):
This is surprisingly clean! A real testament to your design work when implementing the file matching in the first place, I think.

This would need documentation and a changelog, of course, but otherwise I think this is good to go.

eseliger (approved):

internal/batches/graphql/repository.go
```diff
@@ -42,6 +42,8 @@ type Repository struct {
 	URL                string
 	ExternalRepository struct{ ServiceType string }
 
+	BatchIgnoreLocations []string
```
mrnugget, line 45: Note to myself: remove this

cmd/src/batch_common.go
mrnugget: Done!

PR #510, opened 2021-04-12
Restore balance (by adding missing parenthesis in help string)

mrnugget:
I was going to push this to `main`, but then who would see my funny commit message?


eseliger (approved):
Looks good)

mrnugget:
> Looks good)

This made my heartrate jump.

PR #511, opened 2021-04-14
Extract code from batches package into smaller sub-packages

mrnugget:
This is the first step towards cleaning up this code base a bit. 

It extracts code into smaller sub-packages:

- `executor`
- `mock`
- `git`
- `log`
- `service`
- `workspace`

And it renames some of the types so they make more sense in this new structure: `batches.ExecutorOpts` to `executor.Opts` for example.

This doesn't magically fix problems, but it makes it easier to see where dependencies and strong coupling exists.

And it's only the first step: there's still a few TODOs that I added. Next I want to clear up the interplay between `service` and `executor` and the `main` package.

**I opened this PR to make reviewing easier: it doesn't change behaviour but only renames and moves things**


mrnugget:
Pushed another commit that creates the `workspace` package, which I think fits into the rest of these changes.

eseliger (approved):
Nice, thanks for doing this! Can't wait to see what this develops into.. 😬

LawnGnome (approved):
This feels sensible to me. Let's get it in and see how it feels as we use it!

PR #516, opened 2021-04-22
Remove `output` and import from `batch-change-utils/output`

mrnugget:
This is the partner PR to https://github.com/sourcegraph/batch-change-utils/pull/11.

This removes `output` from `src-cli` and imports it from `batch-change-utils` instead.


eseliger (approved):
nice 🎊

mrnugget:
Closing because we said that we'll move it to `sourcegraph/sourcegraph`. New PR coming.

PR #517, opened 2021-04-22
Clean up shared code between `batch apply` and `preview`

mrnugget:
This is a follow-up to #511 and #512.

It cleans up the code that's shared between `batch apply` and `previe`
by trying to make the boundaries between the "TUI +
execution"-glue-code, the `service` pkg, the `executor` pkg and the
helpers clearer.

A lot of this is subjective, of course (what does "clearer" even mean,
man), but I think less duplication is Good and less cross-talk between a
trio of components too.

### DIFF
--- a/internal/batches/executor/executor_test.go
+++ b/internal/batches/executor/executor_test.go
@@ -440,7 +440,6 @@ output4=integration-test-batch-change`,
 					}
 
 					task.Steps = tc.steps
-					// task.Archive = repoFetcher.Checkout(task.Repository, task.Path)
 					executor.AddTask(task)
 				}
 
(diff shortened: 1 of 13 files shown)


internal/batches/executor/executor_test.go
```diff
@@ -438,7 +440,7 @@ output4=integration-test-batch-change`,
 					}
 
 					task.Steps = tc.steps
-					task.Archive = repoFetcher.Checkout(task.Repository, task.Path)
+					// task.Archive = repoFetcher.Checkout(task.Repository, task.Path)
```
eseliger, line 443: Not sure if this is up for review(it's a draft but review was requested) but this seems wrong?

mrnugget: Woops. Shouldn't open PRs when I'm that tired.

And yes, this is wrong, I'll remove it, and yes, it was supposed to be ready for review :)

eseliger (approved):
🧹

LawnGnome:
I like it!

tenor

PR #519, opened 2021-04-23
Use `sourcegraph/sourcegraph/lib/output` instead of `internal/output`

mrnugget:
This is the counterpart to https://github.com/sourcegraph/sourcegraph/pull/20348 and removes `./internal/output`, replacing it with `sourcegraph/sourcegraph/lib/output`.


eseliger (approved):

LawnGnome (approved):
:+1:

PR #520, opened 2021-04-26
Conditional execution of steps with `if:`

mrnugget:
This extends the schema of the batch spec `steps` to add a new field:

- `if: <string that can use templating to evaluate to "true" or "false">`

(the corresponding schema changes `sourcegraph/sourcegraph` are here: https://github.com/sourcegraph/sourcegraph/pull/20399)

`if:` is evaluated in a best-effort partial-evaluation manner before the execution of the `steps`. If we can statically evaluate it, we use the result of that evaluation to build a per-repository/per-workspace list of tasks, which increases cache utilisation.

See the comments in `partial_eval.go` for more details.

It also extends the `StepContext` in the templates to have the `steps.{added,modified,deleted}_files` and `steps.path` fields.

### Example batch spec

```yaml
name: conditional-exec
description: Testing conditional execution of steps

on:
  - repository: github.com/sourcegraph-testing/mkcert
  - repository: github.com/sourcegraph-testing/zap
  - repository: github.com/sourcegraph/automation-testing

steps:
  #
  # Step 1.
  # No `if:`, runs in every repository/workspace
  #
  - run: echo "This is ${{ repository.name }}" >> message.txt
    container: alpine:3
  #
  # Step 2.
  # `if:` checks for repository name. Only runs in github.com/sourcegraph/automation-testing
  #
  - run: echo "hello viewers! is the background noise from the landscapers too loud? leave a message after the beep" >> message.txt
    if: "${{ eq repository.name \"github.com/sourcegraph/automation-testing\" }}"
    container: alpine:3
  #
  # Step 3.
  # `if:` checks repository name. Only runs in repos
  #
  - run: echo "repo name contains sourcegrap-testing" >> message.txt
    # The new builtin function `matches` makes it easy to check for repository names:
    if: "${{ matches repository.name \"*sourcegraph-testing*\" }}"
    container: alpine:3
  #
  # Step 4.
  # Checks for go.mod existance and saves to outputs
  #
  - run:  if [[ -f "go.mod" ]]; then echo "true"; else echo "false"; fi
    container: alpine:3
    outputs:
      goModExists:
        value: ${{ step.stdout }}
  #
  # Step 5.
  # `if:` checks whether the value just saved to outputs is true
  #
  - run: go fmt ./...
    container: golang
    # `if:` supports templating and has access to the full step context, including `previous_step.stdout`, `repository.name`, etc.
    if: ${{ outputs.goModExists }}

  #
  # Step 6.
  # `if:` checks whether the steps are executed in a workspace.
  #
  - run: echo "hello workspace" >> workspace.txt
    container: golang
    if: ${{ eq steps.path "sub/directory/in/repo" }}

changesetTemplate:
  title: Testing conditional execution of steps
  body:  Testing conditional execution of steps
  branch: thorsten/conditional-exec
  published: false
  commit:
    message: Conditional exec of steps
```

### Demo video

Click [here](http://s3.thorstenball.com/conditional_execution.mp4)


internal/batches/executor/templating.go
```diff
@@ -6,24 +6,57 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/gobwas/glob"
 	"github.com/pkg/errors"
 	"github.com/sourcegraph/src-cli/internal/batches/git"
 	"github.com/sourcegraph/src-cli/internal/batches/graphql"
 )
 
+var builtins = template.FuncMap{
```
chrispine, line 15: nice

internal/batches/service/service.go
```diff
@@ -226,7 +241,30 @@ func (svc *Service) BuildTasks(ctx context.Context, repos []*graphql.Repository,
 
 	var tasks []*executor.Task
 
-	attr := &executor.BatchChangeAttributes{Name: spec.Name, Description: spec.Description}
+	// TODO(mrnugget): Factor this out
```
chrispine, line 244: I'm assuming you are still intending TODO this before merging? Or did you already factor it out and forgot this comment? (I don't know what the "this" refers to.)

chrispine (approved):
Just one question about the TODO comment. Thanks!

internal/batches/service/service.go
mrnugget: Yeah. I want to clean this up and possibly do some architectural changes. The PR wasn't meant as "ready for review" or "ready to merge" yet, just as a showcase for the functionality.

mrnugget:
@sourcegraph/batchers this has been updated to also allow `bool|string` for the `if:` attribute.

And I extended how far we can statically evaluate the condition by implementing support for Go's `not` and `ne` template functions. That means common things like `if: ${{ ne repository.name "github.com/foo" }}` or `if: ${{ not (matches repository.name "github.com/horst*") }}` can now be used to work with the cache.

internal/batches/batch_spec_test.go
```diff
@@ -87,4 +90,83 @@ changesetTemplate:
...
+  commit:
+    message: Append Hello World to all README.md files
+  published: false
+`
+
+		for _, tt := range []struct {
+			raw  string
+			want string
+		}{
+			{raw: `"true"`, want: "true"},
```
eseliger, line 153: (Don't add this, I just want to point out how weird yaml is)
```suggestion
			{raw: `"true"`, want: "true"},
			{raw: `yes`, want: "true"},
```

internal/batches/executor/partial_eval.go
```diff
@@ -0,0 +1,388 @@
+package executor
```
eseliger, line 1: pure magic happening here 🧙

eseliger (approved):
Very neat 🌟 Thanks a lot for picking up the suggestion of making it a single field and sorting out the caching with a partial evaluator :)

internal/batches/batch_spec.go
```diff
@@ -178,6 +180,15 @@ func ParseBatchSpec(data []byte, features FeatureFlags) (*BatchSpec, error) {
 		errs = multierror.Append(errs, errors.New("batch spec includes workspaces, which is not supported in this Sourcegraph version"))
 	}
 
+	for i, step := range spec.Steps {
+		if step.If != "" && !features.AllowConditionalExec {
```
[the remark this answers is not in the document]
mrnugget: done!

internal/batches/batch_spec_test.go
mrnugget: Hehe, that actually doesn't work, because we're effectively skipping the YAML parsing for this attribute: if you put in  `yes` the `interface{}` cannot be cast to a `bool`, but is a string.

internal/batches/executor/partial_eval.go
mrnugget: sicp_wizard

internal/batches/batch_spec_test.go
eseliger: I think the yaml parser converts it to bool already 🤔

mrnugget:
```go
func TestYamlFoobar(t *testing.T) {
	type person struct {
		Cool interface{} `json:"Cool" yaml:"cool"`
	}

	personSchema := `{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "title": "Person",
  "description": "Person",
  "type": "object",
  "properties": {
    "cool": {
	  "oneOf": [{ "type": "boolean" }, { "type": "string" }]
    }
  }
}`

	data := []byte(`cool: yes`)
	var p person
	if err := yaml.UnmarshalValidate(personSchema, data, &p); err != nil {
		t.Fatal(err)
	}

	switch cool := p.Cool.(type) {
	case bool:
		t.Errorf("is a bool: %t", cool)
	case string:
		t.Errorf("is a string: %s", cool)
	default:
		t.Errorf("is something else: %+v (%T)", cool, cool)
	}
}

```

prints:

```
--- FAIL: TestYamlFoobar (0.00s)
    batch_spec_test.go:203: is a string: yes
FAIL
FAIL    github.com/sourcegraph/src-cli/internal/batches 0.010s
FAIL
```


Not trying to prove you wrong, but just investigating where it falls apart. I think it's because of our custom yamlv3 unmarshalling etc.

eseliger: Strange 🤔 Well, doesn't matter too much 🤷

internal/batches/debug.go
```diff
@@ -1,6 +1,8 @@
 package batches
 
-import "github.com/sourcegraph/sourcegraph/lib/output"
+import (
+	"github.com/sourcegraph/sourcegraph/lib/output"
+)
```
LawnGnome, line 5: I know where you changed code... :innocent:

internal/batches/executor/partial_eval.go
```diff
@@ -0,0 +1,388 @@
...
+	}
+
+	// TODO: Support finalVal
+	var (
+		// finalVal is the value of the previous Cmd in a pipe (i.e. `${{ 3 + 3 | eq 6 }}`)
+		// It needs to be the final (fixed) argument of a call if it's set.
+		finalVal reflect.Value
+
+		ok bool
+	)
```
LawnGnome, line 125: Maybe these could be named return arguments instead?

internal/batches/executor/partial_eval.go
```diff
@@ -0,0 +1,388 @@
...
+// much as possible.
+//
+// If, after evaluation, more than text is left (i.e. because the template
+// requires information that's only available later) the function returns with
+// the first return value being false, because the template is not "static".
+// first return value is true.
+//
+// If only text is left we check whether that text equals "true". The result of
+// that check is the second return value.
+func isStaticBool(input string, ctx *StepContext) (bool, bool, error) {
```
LawnGnome, line 25: I'm not a big fan of named return arguments, but this might be a case where they'd be useful for documentation purposes, at least?

internal/batches/executor/partial_eval.go
```diff
@@ -0,0 +1,388 @@
...
+		return isStatic, false, nil
+	}
+
+	return true, isTrueOutput(t.Tree.Root), nil
+}
+
+// parseAndPartialEval parses input as a text/template and then attempts to
+// partially evaluate the parts of the template it can evaluate ahead of time
+// (meaning: before we've executed any batch spec steps and have a full
+// StepContet available).
```
LawnGnome, line 48:
```suggestion
// StepContext available).
```

internal/batches/executor/templating.go
```diff
@@ -96,24 +138,17 @@ func (stepCtx *StepContext) ToFuncMap() template.FuncMap {
...
-		"replace": strings.ReplaceAll,
-		"join_if": func(sep string, elems ...string) string {
-			var nonBlank []string
-			for _, e := range elems {
-				if e != "" {
-					nonBlank = append(nonBlank, e)
-				}
-			}
-			return strings.Join(nonBlank, sep)
-		},
```
LawnGnome, line 110: This is a good simplification. :+1:

internal/batches/service/service.go
```diff
@@ -185,96 +184,75 @@ func (svc *Service) EnsureImage(ctx context.Context, name string) (docker.Image,
...
+}
+
+// findWorkspaces matches the given repos to the workspace configs and
+// searches, via the Sourcegraph instance, the locations of the workspaces in
+// each repository.
+// The repositories that were matched by a workspace config are returned in
+// workspaces. root contains the repositories that didn't match a config.
+// If the user didn't specify any workspaces, the repositories are returned as
+// root repositories.
+func (svc *Service) findWorkspaces(
```
LawnGnome, line 207: Is this used anywhere now?

LawnGnome (approved):
This is really cool!

We might want to tighten up what's accepted here slightly: playing around with `if` conditions, I can see that we report errors in templates pretty nicely, but there's some weirdness with non-template strings. If I define a step with, say, `if: dsjfldl`, then that seems to silently gets treated as `true` right now. Maybe we should error in that case?

mrnugget:
> We might want to tighten up what's accepted here slightly: playing around with `if` conditions, I can see that we report errors in templates pretty nicely, but there's some weirdness with non-template strings. If I define a step with, say, `if: dsjfldl`, then that seems to silently gets treated as `true` right now. Maybe we should error in that case?

It's not treated as `true`. If you put in `if: fadsfasdf` then it gets detected as "static" and the boolean value of that condition is `false`. *But* it turns out that it looks like we execute steps if `len(steps) == 0`. I'm going to build a check in for that, but I'm probably going to do that in another PR, so thanks for uncovering :)

internal/batches/executor/partial_eval.go
mrnugget: Good point, changed!

internal/batches/executor/partial_eval.go
mrnugget: Ah! That's good. Fixed.

internal/batches/service/service.go
mrnugget: Nice catch! More - lines for the diff gods!

mrnugget:
> I'm going to build a check in for that, but I'm probably going to do that in another PR, so thanks for uncovering :)

This was far easier than I assumed it would be. Pushed a commit for this.

PR #533, opened 2021-05-11
Set default log to src-cli.debug.log if build tag debug is used

mrnugget:
I don't care about the specifics - filename, build tag, etc. - but I
*need* this. And the reason why I need it is that it's hard to log
something when our TUI is active. Not because of how the TUI is
implemented, but because the whole concept of an updating TUI doesn't
gel with printf-debugging.

And this is perfect for me: I can `tail -f` the log file and then use
`log.Printf` whereever I want.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -78,3 +78,17 @@ To remove it and then force the upstream image on Docker Hub to be used again:
 ```sh
 docker rmi sourcegraph/src-batch-change-volume-workspace
 ```
+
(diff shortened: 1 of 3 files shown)


eseliger (approved):
Should we add this to the readme?

LawnGnome (approved):
I agree with @eseliger about documenting this, but otherwise, this is perfect, and I just wish I'd thought of it several months ago!

mrnugget:
Updated `DEVELOPMENT.md` to include info on this.

PR #535, opened 2021-05-12
Make checking of cache explicit in TUI

mrnugget:
This is based on @eseliger's idea, which was confirmed independently by @courier-new as something worthwhile.

In short: before this change we wouldn't tell users whether we have cached results _and_ we'd only check the check once we start executing a task. The problem with that is that if you have 500 tasks to execute, 480 of which are cached, the UI would tell you "executing 4/500 tasks" and then only speed up once it gets to the cached values.

This PR here pulls the cache-checking out of the task execution and makes it a separate step in the batch spec execution and in the UI:

image

---

**Note to reviewers:** there is also a lot of other clean-up/refactoring/move-around stuff in here that I either (a) had to do to make this easier or (b) wanted to do after digging more into this and realizing that we could probably decouple the executor from this whole process of caching/building-changeset-specs.

I'm not done yet, but I decided to open this PR before things get unreviewable.

Take a look at 3e8475dd7951e3430d35afd76e873b047415517c to see the gist of this change. The other commits mostly move things in separate files, rename stuff, clean up dependency chains.

I'm going to create a new branch off of this one for more cleanups/refactorings.

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -298,7 +298,7 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) error {
 	// EXECUTION OF TASKS
 
 	svc.InitCache(opts.flags.cacheDir)
-	svc.InitExecutor(ctx, executor.Opts{
+	svc.InitExecutor(ctx, executor.NewExecutorOpts{
 		CacheDir:      opts.flags.cacheDir,
 		CleanArchives: opts.flags.cleanArchives,
 		Creator:       workspaceCreator,
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -299,7 +299,6 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) error {
 
 	svc.InitExecutor(ctx, executor.Opts{
 		CacheDir:      opts.flags.cacheDir,
-		ClearCache:    opts.flags.clearCache,
 		CleanArchives: opts.flags.cleanArchives,
 		Creator:       workspaceCreator,
 		Parallelism:   opts.flags.parallelism,
@@ -309,7 +308,7 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) error {
 	})
 
 	pending = batchCreatePending(opts.out, "Checking cache for changeset specs")
-	uncachedTasks, cachedSpecs, err := svc.CheckCache(ctx, tasks)
+	uncachedTasks, cachedSpecs, err := svc.CheckCache(ctx, tasks, opts.flags.clearCache)
 	if err != nil {
 		return err
 	}
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -297,6 +297,7 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) error {
 
 	// EXECUTION OF TASKS
 
+	svc.InitCache(opts.flags.cacheDir)
 	svc.InitExecutor(ctx, executor.Opts{
 		CacheDir:      opts.flags.cacheDir,
 		CleanArchives: opts.flags.cleanArchives,
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -313,10 +313,19 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) error {
 	if err != nil {
 		return err
 	}
-	if len(uncachedTasks) > 0 {
-		batchCompletePending(pending, fmt.Sprintf("Found %d cached changeset specs. %d tasks need to be executed", len(cachedSpecs), len(uncachedTasks)))
+	var specsFoundMessage string
+	if len(cachedSpecs) == 1 {
+		specsFoundMessage = "Found 1 cached changeset spec"
 	} else {
-		batchCompletePending(pending, fmt.Sprintf("Found %d cached changeset specs. No tasks need to be executed", len(cachedSpecs)))
+		specsFoundMessage = fmt.Sprintf("Found %d cached changeset specs", len(cachedSpecs))
+	}
+	switch len(uncachedTasks) {
+	case 0:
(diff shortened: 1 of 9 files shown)


internal/batches/executor/check_cache.go
```diff
@@ -0,0 +1,45 @@
+package executor
+
+import (
+	"context"
+
+	"github.com/pkg/errors"
+	"github.com/sourcegraph/src-cli/internal/batches"
+)
+
+func CheckCache(ctx context.Context, cache ExecutionCache, clearCache bool, features batches.FeatureFlags, task *Task) (specs []*batches.ChangesetSpec, found bool, err error) {
```
mrnugget, line 10: This is Not Good :tm: and I want to refactor this into another entity/thing.

internal/batches/executor/changeset_specs.go
```diff
@@ -0,0 +1,215 @@
...
+import (
+	"fmt"
+	"strings"
+
+	"github.com/pkg/errors"
+	"github.com/sourcegraph/go-diff/diff"
+	"github.com/sourcegraph/src-cli/internal/batches"
+)
+
+func createChangesetSpecs(task *Task, result executionResult, features batches.FeatureFlags) ([]*batches.ChangesetSpec, error) {
```
mrnugget, line 12: Nothing change here. Everything in this file is moved out of `executor.go`. It's all related to building/rendering changeset specs.

internal/batches/executor/executor.go
```diff
@@ -423,11 +258,17 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		return
 	}
 
+	// Build the changeset specs.
```
mrnugget, line 261: This has been moved down so that everything related to the spec-building is close together.

internal/batches/executor/executor.go
```diff
@@ -305,59 +193,6 @@ func (x *executor) do(ctx context.Context, task *Task) (err error) {
 		status.StartedAt = time.Now()
 	})
 
-	// Check if the task is cached.
-	cacheKey := task.cacheKey()
-	if x.clearCache {
```
mrnugget, line 310: This removed section is now in the `CheckCache` function

internal/batches/executor/task.go
```diff
@@ -0,0 +1,40 @@
+package executor
```
mrnugget, line 1: Nothing new in this file. `Task` has been moved here out of `executor.go`

internal/batches/executor/task_status.go
```diff
@@ -0,0 +1,82 @@
+package executor
```
mrnugget, line 1: Nothing new here. `TaskStatus` has been moved here out of `executor.go`

internal/batches/service/service.go
```diff
@@ -28,6 +28,10 @@ type Service struct {
 	client           api.Client
 	features         batches.FeatureFlags
 	imageCache       *docker.ImageCache
+
+	// TODO(mrnugget): I don't like this state here, ugh.
+	exec  executor.Executor
+	cache executor.ExecutionCache
```
mrnugget, line 34: If we have a separate entity that can do the whole check-cache+build-specs stuff, then we don't need this anymore.

cmd/src/batch_common.go
```diff
@@ -295,19 +295,32 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) error {
...
 		TempDir:       opts.flags.tempDir,
+	})
+
+	pending = batchCreatePending(opts.out, "Checking cache for changeset specs")
+	uncachedTasks, cachedSpecs, err := svc.CheckCache(ctx, tasks, opts.flags.clearCache)
+	if err != nil {
+		return err
+	}
+	if len(uncachedTasks) > 0 {
+		batchCompletePending(pending, fmt.Sprintf("Found %d cached changeset specs. %d tasks need to be executed", len(cachedSpecs), len(uncachedTasks)))
```
LawnGnome, line 317: I think we might have hit the point where we need a pluralisation helper, since we have things like "Found 1 cached changeset specs" coming out of this. (It doesn't look like we have one in the `src-cli` tree right now.)

To _completely_ nitpick, it might read better if this is a semicolon instead of a full stop between the clauses, since our other outputs aren't full sentences.

internal/batches/executor/check_cache.go
LawnGnome: What are you thinking here?

(I wrote this before I got to your musings on the fields currently in the `Service` struct, but if there's more than just that, I'm interested!)

LawnGnome (approved):
This looks reasonable to me. I guess the more process question in my mind here is whether you consider this releasable in its current state, given you want to clean this up further: if not, should we merge this to `main`, or have a (very) short lived feature branch for this and your further cleanups that we can review piecemeal, then merge back to `main`?

internal/batches/executor/check_cache.go
mrnugget: I'm not sure yet. I'm playing around with some ideas in another branch. The big problem is the `TaskStatus`es and how they're woven into the executor etc.

cmd/src/batch_common.go
mrnugget: Done! And yes, pluralisation helper would be nice.

mrnugget: (Also: semicolon nitpicks are amazing. I always want to use semicolons but never dare to, because I'm not sure whether it's correct to use)

mrnugget:
> I guess the more process question in my mind here is whether you consider this releasable in its current state, given you want to clean this up further: if not, should we merge this to `main`, or have a (very) short lived feature branch for this and your further cleanups that we can review piecemeal, then merge back to `main`?

Yeah, I get the hestitation. And I agree, this is not ideal. But I've thought about this a bunch (before opening the PR and after) and I think in this case I'd rather merge this to `main` now since it _is_ releasable: all tests pass & I tested this manually a bunch (with manually deleted/added cache files etc.) And since the changes here are large (in LoC numbers) I anticipate a lot of rebase problems if we keep this in a feature branch.

PR #536, opened 2021-05-12
Introduce executor.Coordinator and split up responsibilities between Coordinator/Service/executor

mrnugget:
On the quest to making caching more flexible (see https://github.com/sourcegraph/sourcegraph/issues/20850) I dug into the `executor`, `TaskStatus`, `Service` and how all the things fit together. This is the result of that digging. Hopefully it improves the clarity in the codebase by pulling apart some of the things that have been mixed together over time: caching, execution of steps, updating of task statuses.

It's also a follow-up to https://github.com/sourcegraph/src-cli/pull/535 in which I left a bunch of TODOs lying around. These are now gone.

And it also contains @eseliger's fix from #539.

---

Now, what's in here?

I introduced another entity in the `executor` package: the `Coordinator`. Its job is to coordinate the execution of a batch spec: kick off the `executor` that does the actual execution of `steps`, check the cache before and save the results to the cache after, go through the `importChangesets` statements.

This separation lets us write clearer tests: the executor tests are now only about the execution of steps and the coordinator tests now test the creation and caching of changeset specs, which the executor knows nothing about.

(Sidenote: a better name for the `executor` package would now be `execution` because it would give us: `execution.NewCoordinator` as identifier)

Out of necessity I also had to change how the updating of `TaskStatus`es works. I have some more ideas for the future (how about updates to a task are published on a channel instead of sharing the whole slice of all `TaskStatus` behind a mutex?).


internal/batches/executor/changeset_specs_test.go
```diff
@@ -0,0 +1,313 @@
+package executor
```
mrnugget, line 1: Nothing changed in this file. I only moved the test here from `executor_test.go`

internal/batches/executor/coordinator.go
```diff
@@ -0,0 +1,269 @@
...
+	// Build the changeset specs.
+	specs, err := createChangesetSpecs(taskResult.task, taskResult.result, c.opts.AutoAuthorDetails)
+	if err != nil {
+		return specs, err
+	}
+
+	// Update the status of Task
+	status.Update(taskResult.task, func(status *TaskStatus) {
+		status.ChangesetSpecs = specs
+		// TODO(mrnugget): Ideally we'd get rid of this dependency on the task
```
mrnugget, line 164: The reason we still have this dependency is that we log information about the diff in the TUI.

eseliger (approved):
Nice work! 🧹 I like how this code base is taking some nice shape :)

internal/batches/executor/coordinator.go
```diff
@@ -0,0 +1,269 @@
...
+		// status handler here.
+	})
+
+	return specs, nil
+}
+
+type executionProgressPrinter func([]*TaskStatus)
+
+// Execute executes the given Tasks and the importChangeset statements in the
+// given spec. It regularily calls the executionProgressPrinter with the
```
LawnGnome, line 174:
```suggestion
// given spec. It regularly calls the executionProgressPrinter with the
```

internal/batches/executor/task_status.go
```diff
@@ -8,18 +8,62 @@ import (
...
+// CopyStatuses creates a copy of all TaskStatuses and calls the provided
+// callback with list.
+func (tsc *TaskStatusCollection) CopyStatuses(callback func([]*TaskStatus)) {
+	tsc.statusesMu.RLock()
+	defer tsc.statusesMu.RUnlock()
+
+	var s []*TaskStatus
+	for _, status := range tsc.statuses {
+		s = append(s, status)
+	}
```
LawnGnome, line 55: Do we need to shallow copy the statuses, since the read lock is still going to be held until after `callback(s)` returns anyway?

LawnGnome (approved):
I think this is a really powerful new concept! Great job! :+1:

Are there any background documents on docs.sourcegraph.com that we need to update/write as a result of this?

internal/batches/executor/task_status.go
mrnugget: This is just copy and pasted from the old method, so I didn't chnage it. I'm not 100% sure we still need it. Will investigate in a separate branch (because I'm scared of this part of the codebase)

mrnugget: Ah, the reason we do the copy of the pointers is because `tsc.statuses` is a `map[*Task]*TaskStatus`, not a slice.

PR #538, opened 2021-05-16
batches: make published field optional

LawnGnome:
The first part of sourcegraph/sourcegraph#18277.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,16 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Added
 
+- Starting with Sourcegraph 3.30.0, the `published` field is optional in batch specs. If omitted, the publication state will be controlled through the Batch Changes UI. [#538](https://github.com/sourcegraph/src-cli/pull/538)
+
+### Changed
+
+### Fixed
+
+## 3.29.1
+
+### Added
+
 - LSIF uploads now respect the `-insecure-skip-verify` flag to insecurely (surprise!) skip TLS certificate validation when communicating with Sourcegraph. [#559](https://github.com/sourcegraph/src-cli/pull/559)
 
 ### Changed
--- a/internal/batches/executor/changeset_specs.go
+++ b/internal/batches/executor/changeset_specs.go
@@ -9,7 +9,9 @@ import (
 	"github.com/sourcegraph/src-cli/internal/batches"
 )
 
-func createChangesetSpecs(task *Task, result executionResult, autoAuthorDetails bool) ([]*batches.ChangesetSpec, error) {
+var errOptionalPublishedUnsupported = errors.New(`This Sourcegraph version requires the "published" field to be specified in the batch spec; upgrade to version 3.30.0 or later to be able to omit the published field and control publication from the UI.`)
+
+func createChangesetSpecs(task *Task, result executionResult, features batches.FeatureFlags) ([]*batches.ChangesetSpec, error) {
 	repo := task.Repository.Name
 
 	tmplCtx := &ChangesetTemplateContext{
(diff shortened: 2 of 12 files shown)


internal/batches/executor/changeset_specs.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/sourcegraph/src-cli/internal/batches"
 )
 
+var errOptionalPublishedUnsupported = errors.New("This Sourcegraph version requires the published field to be specified in the batch spec; upgrade to be able to omit the published field and control publication from the UI.")
```
mrnugget, line 12:
```suggestion
var errOptionalPublishedUnsupported = errors.New(`This Sourcegraph version requires the "published" field to be specified in the batch spec; upgrade to be able to omit the published field and control publication from the UI.`)
```

mrnugget (approved):

CHANGELOG.md
```diff
@@ -14,6 +14,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 - Starting with Sourcegraph 3.28.0 batch spec `steps` can contain an `if: <template string>` attribute that determines whether the given step will be executed. [#520](https://github.com/sourcegraph/src-cli/pull/520)
+- Starting with Sourcegraph 3.29.0, the `published` field is optional in batch specs. If omitted, the publication state will be controlled through the Batch Changes UI. [#538](https://github.com/sourcegraph/src-cli/pull/538)
```
eseliger, line 17: Could it be confusing if a version is mentioned in the changelog that doesn't yet exist?

eseliger (approved):

internal/batches/features.go
```diff
@@ -15,6 +15,7 @@ type FeatureFlags struct {
 	AllowWorkspaces          bool
 	BatchChanges             bool
 	AllowConditionalExec     bool
+	AllowOptionalPublished   bool
 }
 
 func (ff *FeatureFlags) SetFromVersion(version string) error {
```
courier-new, line 21: Are all of our feature flags dictated by Sourcegraph version?

LawnGnome: Yep. Feature flag is arguably a misnomer here (my fault); these are a way for us to know what the Sourcegraph backend that `src` is talking to supports, so we can adjust `src`'s behaviour accordingly.

courier-new: Gotcha, thanks!

CHANGELOG.md
LawnGnome: Maybe? I don't mind pulling the changelog entry and reinserting it later, but I also don't know how much our users are going to be trying things out based on the `src` changelog instead of the docs.

eseliger: True. It's probably fine.

internal/batches/executor/changeset_specs.go
```diff
@@ -9,6 +9,8 @@ import (
 	"github.com/sourcegraph/src-cli/internal/batches"
 )
 
+var errOptionalPublishedUnsupported = errors.New(`This Sourcegraph version requires the "published" field to be specified in the batch spec; upgrade to be able to omit the published field and control publication from the UI.`)
```
chrispine, line 12: Should we say that they need to upgrade to at least 3.29.0, or "upgrade to the latest", or leave it as it is? (I don't have a strong feeling one way or the other, but wanted to bring it up.)

chrispine (approved):
LGTM

internal/batches/executor/changeset_specs.go
LawnGnome: I updated this to include 3.30.0 as the version. (If I'm wrong, we can change that tomorrow.)

LawnGnome:
I've rebased this on main, which required undoing a change @mrnugget had made to remove the full feature flag struct from the co-ordinator. Alas. Otherwise, the rebase was fairly mechanical.

In spite of having some rebase pain from leaving this open the first time, I've chosen not to learn from my mistakes, and will hold off on merging this for two reasons, though:

1. I'd like #556 to be reviewed and merged first, and
2. I don't want to release this as part of the expected 3.29.0 release in the next 48 hours.

I expect to hit merge on this on Thursday or Friday.

eseliger (approved):

PR #539, opened 2021-05-17
Fix skip errors flag

eseliger:
We didn't return the specs in case of an error, so we would ignore the error, but still have no changeset specs uploaded, which resulted in the generated batch spec in the preview being empty.


courier-new (approved):
Nice find! Thanks for adding a regression test for this, too.

PR #541, opened 2021-05-18
codeintel: Update codeintel/upload package

efritz:
This updates the codeintelutils package that handles the logic of uploading LSIF indexes to a Sourcegraph instance. This is a big overhaul that affects UX. Screenies below.

- [Multipart upload example](https://user-images.githubusercontent.com/103087/119034939-a0cbfc80-b974-11eb-9564-f43e53bd5491.mp4)
- [Failure message example](https://user-images.githubusercontent.com/103087/119034942-a1649300-b974-11eb-91bb-ed6c18290125.mp4)

Pre-merge checklist: 
- [x] Merge https://github.com/sourcegraph/sourcegraph/pull/21159
- [x] Update go.mod


Strum355 (approved):
Looks awesome :tada: for CI purposes, is this disableable?

efritz:
@Strum355 Sure is. You can pass `-json` to get machine-readable output, `-no-progress` for no animation, and the `lib/output` library apparently detects TTY.

cmd/src/lsif_upload.go
```diff
@@ -331,69 +202,36 @@ type errorWithHint struct {
...
+	if err == upload.ErrUnauthorized {
+		err = errorWithHint{err: err, hint: errUnauthorizedHint}
 	}
+	out.WriteLine(output.Linef(output.EmojiFailure, output.StyleWarning, "error: %s", err))
 
-	if l.verbosity >= lsifUploadVerbosityTraceShowResponseBody {
-		fmt.Printf("< Response Body: %s\n", body)
+	if !lsifUploadFlags.ignoreUploadFailures {
+		// Report but don't return the error
+		os.Exit(0)
```
eseliger, line 228: I think it is a bit weird if the command somewhere deep down calls os.Exit itself and doesn't let the command executor decide what should happen when a command returns an error, wdyt?

eseliger (approved):

cmd/src/lsif_upload.go
efritz: We could early-return nil here instead - would that make sense? Do you have the same feelings on the os.Exit(1)?

eseliger: Yeah, sorry I meant all calls to `os.Exit` outside of the command handler. Not a blocker, but I'm wondering if we could just pass up the error, so in the future we might be able to have things like an error reporter in the command executor or something like that.

efritz: Yeah that makes sense, will convert to a non-terminal flow here.

PR #543, opened 2021-05-21
Clean up the `src version` output

stimut:
The output of `src version` uses inconsistent capitalization of
"version", and the recommended version can sometimes be a downgrade
compared to what is actually installed. The changes in this PR are to:

- standardize on lower-case `v` in "version"
- add `or later` to output of recommended version

Fixes sourcegraph/sourcegraph#16850

### DIFF
--- a/cmd/src/version.go
+++ b/cmd/src/version.go
@@ -33,11 +33,11 @@ Examples:
 			return err
 		}
 		if recommendedVersion == "" {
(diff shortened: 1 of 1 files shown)


mrnugget (approved):
Thank you, @stimut!

eseliger (approved):
Thanks!

PR #546, opened 2021-05-26
refactor: Pass *Task directly to runSteps

mrnugget:
This is a tiny follow-up to #540 and removes the TODO that Adam confirmed in his review.

It simplifies things by passing `*Task` directly to `runSteps`.

### DIFF
--- a/internal/batches/executor/executor.go
+++ b/internal/batches/executor/executor.go
@@ -176,22 +176,15 @@ func (x *executor) do(ctx context.Context, task *Task, status taskStatusHandler)
 
 	// Actually execute the steps.
 	opts := &executionOpts{
-		archive:               task.Archive,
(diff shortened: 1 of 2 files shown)


courier-new (approved):
That's so much nicer, great call! 😄

eseliger (approved):

LawnGnome (approved):
If there was ever an argument for loose coupling of task and the executor, it went out the window when we copied damn near every field on the task into the options anyway. Good thinking!

mrnugget:
> If there was ever an argument for loose coupling of task and the executor, it went out the window when we copied damn near every field on the task into the options anyway. Good thinking!

Yeah, I also think of the executor now more as "the thing that executes tasks" vs. "the thing that executes steps", now with the `Coordinator` and the `TaskBuilder` also doing part of the stuff that the executor previously did.

PR #547, opened 2021-05-26
servegit: decompress input to git-upload-pack

keegancsmith:
Ensure that if the Content-Encoding of a servegit request is gzip then we read it as a compressed input. git-upload-pack does not accept gzipped input and fails. This is only sometimes gzipped and I'm not sure of the conditions so we can't rely on it being always or never encoded.

This is a port of https://github.com/sourcegraph/sourcegraph/pull/21067

Co-authored-by: @efritz

### DIFF
--- a/internal/servegit/service.go
+++ b/internal/servegit/service.go
@@ -1,6 +1,7 @@
 package servegit
 
 import (
+	"compress/gzip"
 	"fmt"
 	"log"
 	"net/http"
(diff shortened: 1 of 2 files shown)


efritz:
Is there any other places that we need to update? (a la https://github.com/sourcegraph/sourcegraph/pull/21067#pullrequestreview-663133145)

efritz (approved):
Is there some common core here that's extractable/worthy of a util function in `github.com/sourcegraph/sourcegraph/lib`?

LawnGnome (approved):

keegancsmith:
> Is there some common core here that's extractable/worthy of a util function in `github.com/sourcegraph/sourcegraph/lib`?

Yes this could make sense to share. We have a lot more instrumentation in gitserver version (which was a copy of this version originally I think :) ). I've added to my personal backlog to explore next month.

> Is there any other places that we need to update? (a la [sourcegraph/sourcegraph#21067 (review)](https://github.com/sourcegraph/sourcegraph/pull/21067#pullrequestreview-663133145))

I believe this is everything. A quick search for something specific to this indicates that. https://sourcegraph.com/search?q=context:%40keegan+x-git-upload-pack-advertisement&patternType=regexp

PR #550, opened 2021-05-28
Fix broken UI by setting FinishedAt again

mrnugget:
This was a left-in "debug removal" that I added in https://github.com/sourcegraph/src-cli/commit/348c49a24dc18e6da5a8a231fea52d840ddcc0a3#diff-e7d4e9760b2767b61760170f3620004882b9fa493251c3619f15d68534891e75L142

It broke the UI. I'll release a new version right after merging this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Fixed
 
 - The step-wise caching for `src batch [apply|preview]` introduced in 3.28.1 could break if a cached diff contained quoted. This fixes the application by disabling any unquoting/expansion.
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- A regression was introduced in 3.28.1 that broke the UI for `src batch [apply|preview]` and lead to the execution of steps looking like it got stuck in the first repository.
```
eseliger, line 20:
```suggestion
- A regression was introduced in 3.28.1 that broke the UI for `src batch [apply|preview]` and led to the execution of steps looking like it got stuck in the first repository.
```

eseliger (approved):

PR #551, opened 2021-05-31
Clear step caches when -clear-cache is specified

mrnugget:
Forgot to implement this but ran into it myself when debugging a batch spec.

Will do a patch release after this is in.


eseliger (approved):
👌

PR #552, opened 2021-06-01
docker: fix publish workflow

### LINKED ISSUE #548: Updating the Docker Hub description on release fails
Example: https://github.com/sourcegraph/src-cli/runs/2680286684

LawnGnome:
Since a credential rotation, the GitHub Action responsible for pushing the [src-batch-change-volume-workspace][1] has been failing. This action works as expected when used via Docker outside of the GitHub Action infrastructure; I suspect there's an issue with the handling of a particular character in the action-specific entry script.

That said, there's not _much_ reason to pull in a third party action here; the push script already has the required credentials, and the Docker Hub API, while undocumented, is straightforward. I've added the required urllib-foo to make this happen. (I'd normally pull in Python's excellent requests library for this kind of thing, but that means we'd have to start installing dependencies in the GitHub Action that runs this script, and that feels like more effort than I'm really willing to go to here.)

Fixes #548.

[1]: https://hub.docker.com/r/sourcegraph/src-batch-change-volume-workspace

### DIFF
--- a/.github/workflows/goreleaser-check.yml
+++ b/.github/workflows/goreleaser-check.yml
@@ -1,4 +1,4 @@
-name: GoReleaser
+name: GoReleaser check
 
 on:
   - push
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -28,15 +28,7 @@ jobs:
         with:
           platforms: arm64
 
-      - run: ./docker/batch-change-volume-workspace/push.py -d ./docker/batch-change-volume-workspace/Dockerfile -i sourcegraph/src-batch-change-volume-workspace -p linux/amd64,linux/arm64,linux/386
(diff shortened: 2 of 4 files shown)


.github/workflows/goreleaser-check.yml
```diff
@@ -1,4 +1,4 @@
-name: GoReleaser
+name: GoReleaser check
```
LawnGnome, line 1: Technically, this is unrelated, but it's been annoying me for several releases now.

eseliger (approved):
This is the most python I've read in two years I guess 😅 not sure how much my review is worth but it _looks_ legit to me. Thanks for looking into this!

.github/workflows/goreleaser-check.yml
courier-new: Helpful!!! 🙌

courier-new (approved):
Looks great! Curious how you were able to test this? 😮

LawnGnome:
> Looks great! Curious how you were able to test this? open_mouth

I ran the script locally, but pushed it to an image in my personal account rather than Sourcegraph. Once this is merged, I'll likely cut a release just to iron out whatever issues we might have with the Action part, which is fairly difficult to test.

PR #556, opened 2021-06-15
batches/executor: fix TestCreateChangesetSpecs

LawnGnome:
The `specWith()` and `taskWith()` helpers inadvertently mutate `defaultChangesetSpec` and `defaultTask` fields, rather than isolating their changes to just the returned spec and task, respectively. Practically speaking, this means that all tests have the same spec and task based on the last element in the `tests` slice, which means that all the test cases end up testing the same, identical thing.

Instead, we must deep copy the spec and task before invoking the mutator function. I've chosen to pull in a third party package (specifically, [copystructure](https://github.com/mitchellh/copystructure)) for this: naïvely using JSON to marshal and unmarshal doesn't work because most of the interesting task fields are excluded when marshalled to JSON.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,6 +12,7 @@ require (
 	github.com/json-iterator/go v1.1.11
 	github.com/kballard/go-shellquote v0.0.0-20180428030007-95032a82bc51
 	github.com/mattn/go-isatty v0.0.12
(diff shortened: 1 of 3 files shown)


chrispine (approved):
LGTM

PR #557, opened 2021-06-22
servegit: use service from sourcegraph/lib

keegancsmith:
This version was duplicated into sourcegraph and then improved. It now
has been moved into the shared library for us to use.

For context see https://github.com/sourcegraph/src-cli/pull/547#pullrequestreview-669502177


LawnGnome (approved):
LGTM, and I appreciate the other dependency updates as well. :+1:

go.mod
```diff
@@ -6,28 +6,29 @@ require (
...
-	golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4 // indirect
-	gopkg.in/yaml.v2 v2.3.0 // indirect
+	golang.org/x/net v0.0.0-20210614182718-04defd469f4e
 	jaytaylor.com/html2text v0.0.0-20200412013138-3577fbdbcff7
 )
 
 replace github.com/gosuri/uilive v0.0.4 => github.com/mrnugget/uilive v0.0.4-fix-escape
+
+// See: https://github.com/ghodss/yaml/pull/65
+replace github.com/ghodss/yaml => github.com/sourcegraph/yaml v1.0.1-0.20200714132230-56936252f152
```
keegancsmith, line 34: I'm so confused where this came from? I didn't do this. Maybe `go mod` magic did it via what is in sourcegraph/lib? Just FYI if you see some funky regression related to yaml.

keegancsmith: argh I have a really old base revision. my bad. I think I ddi this while fixing something with a `git checkout origin/main`.

keegancsmith:
@LawnGnome FYI I dropped support for go1.15 on CI. I requested re-review just to confirm that is fine :)

eseliger (approved):

PR #559, opened 2021-06-25
Support -insecure-skip-verify flag for LSIF upload

Strum355:
As we previously used a different API client as was provided in the src-cli codebase, we did not take into account the `-skip-insecure-verify` flag. This affected a customer as they used self-signed certs.

This PR passes the src-cli configured API client that honours the flag to the LSIF upload function that resides in sg/sg/lib. The only part of the API client that is consumed is the direct pass-through method, [Do(...)](https://sourcegraph.com/github.com/sourcegraph/src-cli@main/-/blob/internal/api/api.go?L147)

Related to https://github.com/sourcegraph/sourcegraph/pull/22399 and will require bumping go.mod here when it is merged.

cc @efritz for :eyes: after PTO

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"flag"
 	"fmt"
+	"io"
 	"net/url"
 	"os"
 	"strings"
@@ -66,6 +67,7 @@ func handleLSIFUpload(args []string) error {
 	}
 
 	client := api.NewClient(api.ClientOpts{
+		Out:   io.Discard,
 		Flags: lsifUploadFlags.apiFlags,
 	})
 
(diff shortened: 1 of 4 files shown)


Strum355:
**Note** tests are failing due to `go.mod` pointing at the old sg/sg/lib API while the related PR is unmerged

cmd/src/lsif_upload.go
```diff
@@ -64,7 +65,11 @@ func handleLSIFUpload(args []string) error {
 		return handleLSIFUploadError(nil, err)
 	}
 
-	uploadID, err := upload.UploadIndex(lsifUploadFlags.file, lsifUploadOptions(out))
+	client := api.NewClient(api.ClientOpts{
+		Flags: lsifUploadFlags.apiFlags,
+	})
+
```
coury-clark, line 71: It might be a good idea to log a warning if the client is initialized without TLS validation.

coury-clark (approved):

cmd/src/lsif_upload.go
Strum355: Thats a good idea, I'll follow up with a separate PR for that, as those changes would be more wide-reaching across teams 🙂

PR #561, opened 2021-06-28
docker: remove description updates when publishing a release

LawnGnome:
We used to update the Docker Hub description for
https://hub.docker.com/r/sourcegraph/src-batch-change-volume-workspace
on release, but doing so reliably has proven to be difficult, as Docker
Hub doesn't provide a reliable API to do this. When it fails, the entire
release shows up as red, which is obviously not ideal!

Since this description rarely changes, and the pushes themselves are
reliable, this removes the description update and makes it a manual
process.


eseliger (approved):
That must've been fun 🙂

LawnGnome:
> That must've been fun slightly_smiling_face

I tried to fix it a few weeks back, and it just won't go. I can take a hint.

PR #562, opened 2021-07-02
For internal use only: add a -text-only flag to `src batch [apply|preview]`

mrnugget:
This adds a `-text-only` flag to `src batch [apply|preview]` that is **for internal use only**.

It's a highly-experimental feature that should not be used if you're not sure about what you're doing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 - Starting with Sourcegraph 3.30.0, the `published` field is optional in batch specs. If omitted, the publication state will be controlled through the Batch Changes UI. [#538](https://github.com/sourcegraph/src-cli/pull/538)
+- For internal use only: `-text-only` flag added to `src batch [apply|preview]`. [#562](https://github.com/sourcegraph/src-cli/pull/562)
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


LawnGnome (commented):
One semi-related thought: right now we have a fairly half-arsed attempt at a non-fancy mode in `output` to handle the case where the output isn't going directly to a terminal. I'm thinking we might want to remove that and use this as the default once we have this fully implemented.

mrnugget:
> One semi-related thought: right now we have a fairly half-arsed attempt at a non-fancy mode in `output` to handle the case where the output isn't going directly to a terminal. I'm thinking we might want to remove that and use this as the default once we have this fully implemented.

Ah, yes. I forgot about that. Question is though: is the text output for human or machine consumption? Because what's in here is definitely for machines.

LawnGnome:
> Question is though: is the text output for human or machine consumption?

More human, but I'm not super convinced it's actually very useful for that.

manfredwedderin (approved):

mrnugget:
@sourcegraph/batchers As discussed yesterday: I want to get this in. I have some ideas on how to reduce the code duplication, but since that would require touching the _old_ code I'd rather add a bit of duplication for now to get something running fast than to touch and possibly break the old code. For that I want to take my time.

eseliger (approved):

PR #565, opened 2021-07-08
Reduce duplication in text-only UI and TUI

mrnugget:
This introduces the concept of a `batchExecUI`: the UI that displays the
status of a batch spec execution.

Both `batchExecTUI` and `batchExecJSONLinesUI` implement that interface
and can thus be used interchangeably.

The result is that we can get rid of the duplicated code paths.

Note: this is not as nice as it could be, because I want to be careful
and only change what is necessary in this PR to get rid of the
duplication (vs. rethinking how we batch spec execution reports
progress, etc.) So please review from that point of view.


cmd/src/batch_exec_tui.go
```diff
@@ -0,0 +1,361 @@
...
+	defer block.Close()
+
+	block.Write("To view the batch change, go to:")
+	block.Writef("%s", batchChangeURL)
+}
+
+func (ui *batchExecTUI) ExecutionError(err error) {
+	printExecutionError(ui.out, err)
+}
+
```
mrnugget, line 222: Everything below this line was just moved here from `batch_common.go`

eseliger (approved):

mrnugget:
FYI @courier-new` @LawnGnome Since there's so time pressure to get this in ASAP I'm waiting another day to merge this in case you also want to add something :)

cmd/src/batch_exec_json_lines_ui.go
```diff
@@ -0,0 +1,222 @@
...
+	logOperationStart("RESOLVING_REPOSITORIES", "")
+}
+
+func (ui *batchExecJSONLinesUI) ResolvingRepositoriesDone(repos []*graphql.Repository, unsupported batches.UnsupportedRepoSet, ignored batches.IgnoredRepoSet) {
+	if unsupported != nil && len(unsupported) != 0 {
+		logOperationSuccess("RESOLVING_REPOSITORIES", fmt.Sprintf("%d unsupported repositories", len(unsupported)))
+	} else if ignored != nil && len(ignored) != 0 {
+		logOperationSuccess("RESOLVING_REPOSITORIES", fmt.Sprintf("%d ignored repositories", len(ignored)))
+	} else {
+		logOperationSuccess("RESOLVING_REPOSITORIES", fmt.Sprintf("Resolved %d repositories", len(repos)))
```
LawnGnome, line 67: Since this will presumably end up in the UI, these should all be pluralised. (Let's introduce or import a helper, since this comment is going to come up a bunch.)

cmd/src/batch_exec_json_lines_ui.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+}
+
+func (ui *batchExecJSONLinesUI) DeterminingWorkspaces() {
+	logOperationStart("DETERMINING_WORKSPACES", "")
+}
+
+func (ui *batchExecJSONLinesUI) DeterminingWorkspacesSuccess(num int) {
+	logOperationSuccess("DETERMINING_WORKSPACES", fmt.Sprintf("Found %d workspaces with steps to execute", num))
```
LawnGnome, line 76: This needs a pluralise helper.

cmd/src/batch_exec_tui.go
```diff
@@ -0,0 +1,361 @@
...
+		block.Close()
+	}
+}
+
+func (ui *batchExecTUI) DeterminingWorkspaces() {
+	ui.pending = batchCreatePending(ui.out, "Determining workspaces")
+}
+
+func (ui *batchExecTUI) DeterminingWorkspacesSuccess(num int) {
+	batchCompletePending(ui.pending, fmt.Sprintf("Found %d workspaces with steps to execute", num))
```
LawnGnome, line 112: This needs a pluralise helper.

cmd/src/batch_exec_tui.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func (ui *batchExecTUI) NoChangesetSpecs() {
+	ui.out.WriteLine(output.Linef(output.EmojiWarning, output.StyleWarning, `No changeset specs created`))
+}
+
+func (ui *batchExecTUI) UploadingChangesetSpecs(num int) {
+	var label string
+	if num == 1 {
+		label = "Sending changeset spec"
```
LawnGnome, line 165:
```suggestion
		label = "Sending 1 changeset spec"
```

cmd/src/batch_exec_tui.go
```diff
@@ -0,0 +1,361 @@
...
+				// verbose mode, but let the original error bubble up rather
+				// than this one.
+				out.Verbosef("Unexpected error parsing the GraphQL error: %v", cerr)
+			} else if code == "ErrCampaignsUnlicensed" || code == "ErrBatchChangesUnlicensed" {
+				// OK, let's print a better message, then return an
+				// exitCodeError to suppress the normal automatic error block.
+				// Note that we have hand wrapped the output at 80 (printable)
+				// characters: having automatic wrapping some day would be nice,
+				// but this should be sufficient for now.
+				block := out.Block(output.Line("🪙", output.StyleWarning, "Batch Changes is a paid feature of Sourcegraph. All users can create sample"))
```
LawnGnome, line 248: I'm still proud of this emoji choice. :muscle:

LawnGnome (approved):
This is great! It's much closer to modelling what's really going on here, where the output is essentially a side effect of the state machine that's driving batch spec execution.

I have a bunch of really nitpicky comments where we're not pluralising correctly, but otherwise, let's do it!

courier-new (approved):
I didn't give this a super thorough review as I'm still wrapping my head around what this change means or does but nothing jumped out to me as bad so 👍 🙂

PR #567, opened 2021-07-14
Fix empty diff being returned when single *cached* step had to be executed

mrnugget:
This PR fixes a bug in the step-wise caching logic that @eseliger and I ran into last week.

### Problem

When no cached result for the *complete* task was found, but a cached result for a single step and that step was the only left to execute then an empty diff was returned.

### How to reproduce

1. Execute a batch spec with the following steps

    ```yaml
    steps:
      - run: echo "this is step 1" >> README.txt
        container: alpine:3
      - run: echo "this is step 2" >> README.md
        container: alpine:3
    ```
    the complete results and the results for each step are now cached.
2. Update the batch spec and re-execute:
    ```yaml
    steps:
      - run: echo "this is step 1" >> README.txt
        container: alpine:3
    ```

This will produce an empty diff because the `Coordinator` did not find a cached result for the complete task/batch spec, but for the first step.

`runSteps` didn't handle this case though, when only a single step had to be executed but that was also cached.

### The fix

Take a look at my comments in the diffs here in GitHub to see what the fix is. Spoiler: it's just a condition and an early exit to handle the case of "cached step result == the only step to execute".

In order to get to this fix though I wrote a lot of tests, found one other bug (`Files` was not cached), and cleaned up the 300 line long `runSteps` function.

`runSteps` is now much easier to understand and doesn't contain _all_ levels of abstraction. The outer `runSteps` concerns itself with which steps need to be executed and how to handle their results and `executeSingleStep` does the lower-level work of rendering templates, starting containers, setting up the logger, etc.

### DIFF
--- a/internal/batches/executor/run_steps.go
+++ b/internal/batches/executor/run_steps.go
@@ -78,13 +78,15 @@ func runSteps(ctx context.Context, opts *executionOpts) (result executionResult,
 	}
 	defer workspace.Close(ctx)
 
-	execResult := executionResult{
-		Outputs: make(map[string]interface{}),
-		Path:    opts.task.Path,
-	}
-	results := make([]StepResult, len(opts.task.Steps))
+	var (
+		execResult = executionResult{
+			Outputs: make(map[string]interface{}),
+			Path:    opts.task.Path,
+		}
+		previousStepResult StepResult
+		startStep          int
+	)
 
-	var startStep int
 	if opts.task.CachedResultFound {
 		// Set the Outputs to the cached outputs
 		execResult.Outputs = opts.task.CachedResult.Outputs
@@ -100,8 +102,6 @@ func runSteps(ctx context.Context, opts *executionOpts) (result executionResult,
 		default:
 			opts.reportProgress(fmt.Sprintf("Skipping steps 1 to %d. Found cached results.", startStep))
 		}
-	} else {
-		startStep = 0
 	}
 
 	for i := startStep; i < len(opts.task.Steps); i++ {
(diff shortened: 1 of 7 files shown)


internal/batches/executor/run_steps.go
```diff
@@ -352,13 +197,138 @@ func runSteps(ctx context.Context, opts *executionOpts) (result executionResult,
...
 	execResult.Diff = string(diffOut)
-	if len(results) > 0 && results[len(results)-1].files != nil {
-		execResult.ChangedFiles = results[len(results)-1].files
-	}
+	execResult.ChangedFiles = previousStepResult.Files
 
 	return execResult, stepResults, err
 }
 
+const workDir = "/work"
```
mrnugget, line 205: Everything below this line is "just" extracted code from `runSteps`. There is no new logic, it's the same code, but in separate functions.

internal/batches/executor/run_steps.go
```diff
@@ -78,27 +78,44 @@ func runSteps(ctx context.Context, opts *executionOpts) (result executionResult,
...
+				return execResult, nil, errors.Wrap(err, "parsing cached step diff")
+			}
+
+			execResult.Diff = string(opts.task.CachedResult.Diff)
+			execResult.ChangedFiles = &changes
+			stepResults = append(stepResults, opts.task.CachedResult)
+
+			return execResult, stepResults, nil
+		}
+
```
mrnugget, line 112: This is the fix for the problem this PR aims to address. The rest of the changes are either tests or were made to make this fix easy to make (like the removal of the whole `results` indexing, introducing `previousStepResult`, etc.)

eseliger (approved):
Love that cleanup. LGTM!

courier-new (approved):
Looks good as far as I can tell! Nice cleanup, indeed!

PR #574, opened 2021-08-03
Fix typo in message printed during batch runs

sankalp-khare:
While doing a `src batch` run I noticed there was a typo in the output. Traced that to this line of code. Please accept the fix.

Thanks!~


LawnGnome:
Whoops! Thanks for catching that! :rocket:

PR #575, opened 2021-08-06
batches: disable network access by default

martin-sucha:
Most of the time the container should not need network access as all
the tools to change the files should be available in the Docker image
already. Preventing network access by default reduces chance that
some code in the Docker image would send the source to the internet.

### DIFF
--- a/internal/batches/batch_spec.go
+++ b/internal/batches/batch_spec.go
@@ -93,6 +93,7 @@ type Step struct {
 	Env       env.Environment   `json:"env,omitempty" yaml:"env"`
 	Files     map[string]string `json:"files,omitempty" yaml:"files,omitempty"`
 	Outputs   Outputs           `json:"outputs,omitempty" yaml:"outputs,omitempty"`
+	AllowNetwork bool           `json:"allow_network,omitempty" yaml:"allow_network,omitempty"`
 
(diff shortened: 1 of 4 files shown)


martin-sucha:
This is theoretically a breaking change, but I think it's worth it because of the security benefit as most of the time no network access should be needed.

When I try this code I get error from server, probably also needs to have schema updated on the server side.

LawnGnome:
Thanks for the PR, @martin-sucha! This definitely feels like a useful piece of hardening work.

Our next team meeting is on Tuesday, and I've added this PR to our agenda to discuss: I think the backward compatibility point you raised is a valid concern for us, so we'll talk it over and decide what we want to do from there.

cc: @malomarrec

mrnugget:
Hey @martin-sucha! Sorry for taking so long to get back to you on this.

We've talked about this idea here again and we all agreed that it doesn't make sense as a default right now. A lot of usecases for batch changes use network connections (e.g. install dependencies) and would break with the introduction of this.

martin-sucha:
@mrnugget Thanks for the update! What do you think about adding the `allow_network` option, but with default=true?

mrnugget:
That would be much better, as long as adding the flag itself doesn't change the default behaviour in any way.

PR #577, opened 2021-08-12
batches: fix feature detection in `src batch validate`

### LINKED ISSUE #576: batches: feature detection fails in src batch validate
I'm not sure of the root cause yet, but any spec feature that relies on Sourcegraph feature detection currently fails. For example, with a batch spec using workspaces:

```
$ src batch validate -f ../scratch/new-campaigns/workspaces.batch.yaml 
parsing batch spec: 1 error occurred:
	* batch spec includes workspaces, which is not supported in this Sourcegraph version
```

But actually executing the spec is fine:

```
$ src batch preview -f ../scratch/new-campaigns/workspaces.batch.yaml 
✅ Parsing batch spec
✅ Resolving namespace
✅ Preparing container images  ███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████
✅ Set workspace type
✅ Resolved 1 repositories
✅ Found 1 workspaces with steps to execute
✅ Found 0 cached changeset specs; 1 task needs to be executed
✅ Executing... (1/1, 0 errored)  ████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████
✅ Sending changeset spec  ███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████
✅ Creating batch spec on Sourcegraph

✅ To preview or apply the batch spec, go to:
```

LawnGnome:
When creating the `Service` instance, `src batch validate` failed to request or initialise the Sourcegraph feature flags based on the Sourcegraph version in use. As a result, only baseline features were enabled when parsing and validating batch specs.

Fixes #576.

### DIFF
--- a/cmd/src/batch_validate.go
+++ b/cmd/src/batch_validate.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"context"
 	"flag"
 	"fmt"
 
 	"github.com/sourcegraph/sourcegraph/lib/output"
+	"github.com/sourcegraph/src-cli/internal/api"
 	"github.com/sourcegraph/src-cli/internal/batches/service"
 	"github.com/sourcegraph/src-cli/internal/batches/ui"
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
@@ -25,9 +27,12 @@ Examples:
 `
 
 	flagSet := flag.NewFlagSet("validate", flag.ExitOnError)
+	apiFlags := api.NewFlags(flagSet)
 	fileFlag := flagSet.String("f", "", "The batch spec file to read.")
 
 	handler := func(args []string) error {
+		ctx := context.Background()
+
 		if err := flagSet.Parse(args); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


cmd/src/batch_validate.go
```diff
@@ -36,7 +41,13 @@ Examples:
...
 		}
 
-		svc := service.New(&service.Opts{})
+		svc := service.New(&service.Opts{
+			Client: cfg.apiClient(apiFlags, flagSet.Output()),
+		})
+
+		if err := svc.DetermineFeatureFlags(ctx); err != nil {
+			return err
+		}
```
eseliger, line 50: Just a side-note, no action required: It'd be nice if this is not something that can be forgotten 😬 

```go
package service

func NewWithFeatureFlags(client *api.Client) (Service, error) {
	svc := service.New(&service.Opts{
		Client: client,
	})
	err := svc.DetermineFeatureFlags(ctx)
	return svc, err
}
```

eseliger (approved):
Great catch!

cmd/src/batch_validate.go
LawnGnome: Yeah, I was on the fence on this one too. I think I'd rather keep this PR as trivial as it is for now, but we might want to rethink in general how we design and couple the service at some point, since it's not as simple as it once was.

PR #578, opened 2021-08-17
Remove published: false from template for batch new

eseliger:
Now that this property isn't required anymore, we can get rid of that part. Honors the feature flag for optional publish values.

Closes https://github.com/sourcegraph/sourcegraph/issues/22180

### DIFF
--- a/cmd/src/batch_new.go
+++ b/cmd/src/batch_new.go
@@ -1,16 +1,13 @@
 package main
 
 import (
+	"context"
 	"flag"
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
LGTM.

courier-new (commented):
Great refactor here as well! Thanks so much Erik! 😄

eseliger:
Involuntarily 😆  The feature flags are only available in the service layer :)

PR #584, opened 2021-08-23
test: don't rely on root being a git repo

stefanhengl:
The test `TestInferRepo` unnecesarily depends on the root being a git
repository. This change updates `TestInferRepo` to create a fresh git
repo in a temporary dir.

Why?

I want to add src-cli to the nixos package repository. During the build
stage, nixos downloads a tar of the repo. The tar, however, is not a git
repo and thus `TestInferRepo` fails.


efritz (approved):
Thanks for the _why_!

PR #587, opened 2021-08-23
Use correct ref when finding directories in repos

mrnugget:
@eseliger found this potential bug. It's extracted from the other PR with the tests fixed.

### DIFF
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -648,7 +648,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 		a.WriteString("query DirectoriesContainingFile {\n")
 
 		for _, repo := range batch {
-			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$ type:path count:99999`, regexp.QuoteMeta(fileName), regexp.QuoteMeta(repo.Name))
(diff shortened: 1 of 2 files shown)


internal/batches/service/service.go
```diff
@@ -660,7 +660,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 		a.WriteString("query DirectoriesContainingFile() {\n")
 
 		for _, repo := range batch {
-			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$ type:path count:99999`, regexp.QuoteMeta(fileName), regexp.QuoteMeta(repo.Name))
+			query := fmt.Sprintf(`file:(^|/)%s$ repo:^%s$@%s type:path count:99999`, regexp.QuoteMeta(fileName), repo.BaseRef(), regexp.QuoteMeta(repo.Name))
```
eseliger, line 663: I'm a bit scared of spaces or whatever breaking this, but what can we do 🤷‍♂️

eseliger (approved):
Just to make sure: did you give this a try yet? I didn't actually try it out yet on that branch where it came from.

internal/batches/service/service.go
mrnugget: Spaces in branch names?

eseliger: no idea what people can get into git.. :D

mrnugget:
> Just to make sure: did you give this a try yet?

Yep, tried it locally.

PR #588, opened 2021-08-23
Refactor workspaces

eseliger:


internal/batches/errors.go
```diff
@@ -77,3 +77,9 @@ func (e IgnoredRepoSet) Append(repo *graphql.Repository) {
 func (e IgnoredRepoSet) HasIgnored() bool {
 	return len(e) > 0
 }
+
+type ValidationError struct{ Reason string }
+
+func (e ValidationError) Error() string {
+	return fmt.Sprintf("validation failed: %s", e.Reason)
+}
```
eseliger, line 85: I think I want to have this to distinguish validation things from actual errors, so we can in the future have better error reporting. Can remove it, though.

internal/batches/service/task_builder.go
```diff
@@ -81,17 +80,24 @@ func (tb *taskBuilder) buildTask(r *graphql.Repository, path string, onlyWorkspa
...
 
-	var tasks []*executor.Task
-	for repo, ws := range workspaces {
-		for _, path := range ws.paths {
-			t, ok, err := tb.buildTask(repo, path, ws.onlyFetchWorkspace)
+	tasks := []*executor.Task{}
+	for _, ws := range workspaces {
+		repo, ok := repoByID[ws.RepoID]
+		if !ok {
+			return nil, errors.New("invalid state")
```
eseliger, line 93: it is, right?

internal/batches/service/workspaces.go
```diff
@@ -0,0 +1,106 @@
...
+// workspaces. root contains the repositories that didn't match a config.
+// If the user didn't specify any workspaces, the repositories are returned as
+// root repositories.
+func findWorkspaces(
+	ctx context.Context,
+	spec *batcheslib.BatchSpec,
+	finder directoryFinder,
+	repos []*graphql.Repository,
+) ([]RepoWorkspaces, error) {
+	// Pre-compile all globs.
```
eseliger, line 32: This function and type can relatively easy be pulled over to the backend now. Next step :)

internal/batches/service/workspaces.go
```diff
@@ -0,0 +1,106 @@
+package service
```
mrnugget, line 1: I think we need to move some tests too

internal/batches/service/workspaces.go
```diff
@@ -0,0 +1,106 @@
...
+	RepoID             string
+	Paths              []string
+	OnlyFetchWorkspace bool
+}
+
+// findWorkspaces matches the given repos to the workspace configs and
+// searches, via the Sourcegraph instance, the locations of the workspaces in
+// each repository.
+// The repositories that were matched by a workspace config are returned in
+// workspaces. root contains the repositories that didn't match a config.
```
mrnugget, line 23: Comment is outdated

mrnugget (approved):

internal/batches/service/workspaces.go
eseliger: Done!

mrnugget (approved):

PR #593, opened 2021-09-01
Make template package independent of src-cli internals

eseliger:
- Get rid of dependency on graphql.Repository
- Move git package into sg/sg/lib


internal/batches/service/build_tasks.go
```diff
@@ -37,21 +38,23 @@ func buildTasks(ctx context.Context, spec *batcheslib.BatchSpec, repos []*graphq
...
-	var taskSteps []batcheslib.Step
+	taskSteps := []batcheslib.Step{}
 	for _, step := range spec.Steps {
+		// If no if condition is given, just go ahead and add the step to the list.
 		if step.IfCondition() == "" {
 			taskSteps = append(taskSteps, step)
 			continue
 		}
 
+		stepCtx := &template.StepContext{
```
mrnugget, line 54: Do you need a fresh `template.StepContext` in every iteration?

internal/batches/template/templating.go
```diff
@@ -85,6 +85,31 @@ type BatchChangeAttributes struct {
...
+}
+
+type TemplatingRepository struct {
+	ID            string
+	Name          string
+	DefaultBranch TemplatingBranch
+	FileMatches   map[string]bool
+}
+
+func (r TemplatingRepository) SearchResultPaths() (list fileMatchPathList) {
```
mrnugget, line 100: Do we need the `Templating` in the name? `template.TemplatingRepository` is a bit redundant.

mrnugget (approved):

PR #597, opened 2021-09-01
text-only output: log stdout/stderr of steps

mrnugget:
This is part of https://github.com/sourcegraph/sourcegraph/issues/23496 and adds the logging of STDOUT/STDERR of steps to the `-text-only` output. Please see the comment I left for more info on how the interface could be improved/extended.

Also: turns out I don't need the `process.Logger` that was extracted in https://github.com/sourcegraph/sourcegraph/pull/24545 and instead what's needed is the `IntervalLogger` here that flushes the captured output with a delay (so we don't create a new huge JSON message for every line)

### DIFF
--- a/internal/batches/executor/executor.go
+++ b/internal/batches/executor/executor.go
@@ -1,6 +1,7 @@
 package executor
 
 import (
+	"bytes"
 	"context"
 	"fmt"
 	"os/exec"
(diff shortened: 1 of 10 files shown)


internal/batches/executor/executor.go
```diff
@@ -176,6 +176,8 @@ func (x *executor) do(ctx context.Context, task *Task, ui TaskExecutionUI) (err
 		reportProgress: func(currentlyExecuting string) {
 			ui.TaskCurrentlyExecuting(task, currentlyExecuting)
 		},
+		newUiStdoutWriter: ui.StepStdoutWriter,
+		newUiStderrWriter: ui.StepStderrWriter,
```
mrnugget, line 180: The idea is that in the near future we don't pass 3+ things to `runSteps`, but instead just a single thing that implements something like `StepsExecutionUI`, which has methods for different events: `DownloadingArchive()`, `StartingStep()`, `FinishedStep()`, etc.

But that seems a bit out of scope of this PR and we still don't know 100% what exactly we need from the JSON output in the future, since we might be able to get around some logging once we have more information about which Task gets executed on the server-side.

eseliger (approved):

PR #598, opened 2021-09-02
Refactor zipfetcher and consistently use rev, not baseref

eseliger:
Before, sometimes we used Rev, sometimes BaseRef, which could in the worst case lead to inconsistencies. Search results might not match the fetched zip archive, or detected workspaces might differ from what's on disk in the end.


internal/batches/graphql/repository.go
```diff
@@ -60,7 +58,7 @@ func (r *Repository) BaseRef() string {
 		return ensurePrefix(r.Branch.Name)
 	}
 
-	return r.DefaultBranch.Name
+	return ensurePrefix(r.DefaultBranch.Name)
```
eseliger, line 61: This was inconsistent.

internal/batches/repozip/fetcher.go
```diff
@@ -13,63 +13,105 @@ import (
...
 	// Users need to call `Fetch()` on the RepoZip before using it and
 	// `Close()` once // they're done using it.
-	Checkout(repo *graphql.Repository, path string) RepoZip
+	Checkout(repo RepoRevision, path string) Archive
+}
+
+// Archive implementations represent a downloaded repository archive.
+type Archive interface {
+	// Fetch downloads the archive if it's not on disk yet.
+	Fetch(context.Context) error
```
mrnugget, line 37: If we're renaming things we might want to rethink `Fetcher`, since the `Archive` itself has a `Fetch` method.

Ideas I just had while looking at this (meaning: they're not the best we could come up with, most likely): `Fetcher` could be `ArchiveRegistry` and `Archive.Fetch` could be `Ensure()` or something.

internal/batches/util/repo.go
```diff
@@ -1,16 +1,40 @@
...
+func NewTemplatingRepo(repoName string, fileMatches map[string]bool) template.Repository {
 	return template.Repository{
-		Name:        r.Name,
-		FileMatches: r.FileMatches,
+		Name:        repoName,
+		FileMatches: fileMatches,
+	}
+}
+
+func SlugForPathInRepo(repoName, commit, path string) string {
```
mrnugget, line 20: Random thought: instead of a `util` package, could we have a `repository` package? `batches/repository.SlugForPath` is a good name.

mrnugget (approved):
Didn't look too closely, but idea looks good to me! I like `repozip` as a package.

internal/batches/repozip/fetcher.go
eseliger: Done 👍

PR #605, opened 2021-09-03
Move fallback from `.` to `""` in workspace path into workspace finder

eseliger:
I don't see where this could happen. We get the path from the search API and that returns "" for the root. Can we remove this special handling?

### DIFF
--- a/internal/batches/service/build_tasks.go
+++ b/internal/batches/service/build_tasks.go
@@ -14,32 +14,21 @@ func buildTasks(ctx context.Context, spec *batcheslib.BatchSpec, workspaces []Re
 	tasks := make([]*executor.Task, 0, len(workspaces))
 
 	for _, ws := range workspaces {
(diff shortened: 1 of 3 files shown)


eseliger:
TODO: Find out if this is actually not needed anymore.

mrnugget:
We run the paths through `path.Dir` when finding workspaces, which returns `"."` if it's in the root:

https://play.golang.org/p/RmqBbk81BZm

eseliger:
Oh! Will adjust this PR then

mrnugget:
Ideally we'd change that where we call `path.Dir`, so we have all the logic in a single place and can have "`""` means root" as an invariant everywhere.

eseliger:
Yes, that's what I wanted to do

eseliger:
Updated. PTAL!

mrnugget (approved):

PR #606, opened 2021-09-03
executor UI: extract StepsExecutionUI from TaskExecutionUI and implement it in TUI and JSON

mrnugget:
This is a follow-up to https://github.com/sourcegraph/src-cli/pull/597 and splits up the existing interface into two.

That allows us to remove all formatting from `runSteps` and `executeSingleStep` and also provide more information for the JSON interface (e.g. when a step is finished).

**Note**: the `OPERATION` and `STATUS` fields in the `logEvents` don't make a lot of sense. Happy to change everything there.

### DIFF
--- a/internal/batches/executor/coordinator.go
+++ b/internal/batches/executor/coordinator.go
@@ -2,13 +2,13 @@ package executor
 
 import (
 	"context"
-	"io"
 	"reflect"
 	"strconv"
 	"time"
 
 	"github.com/cockroachdb/errors"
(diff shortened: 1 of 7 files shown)


mrnugget:
@sourcegraph/batchers Merging this after @eseliger discussed this a bit. We both think that we need to change the events and what data is logged when anyway, so we won't sweat that now until we know more about what the workflow in the web UI will be and which data we need to display.

PR #608, opened 2021-09-03
For internal use only: add `src batches exec` command

mrnugget:
This adds the for-internal-use-only `src batch exec` command that will be used by server-side batch changes to execute a batch spec in specific workspaces on different machines, concurrently.

**Note:** the vast majority of the added code you see here is copy&pasted from `batch_common.go`. I really _don't_ want to add more conditionals to the huge function in there though, so I created a new one. Yes, there's duplication, but it's so much easier to understand and to maintain. And we'll likely make a lot of changes to this new function here anyway, so effort to generalize and DRY this code will probably be wasted.

### Example

Given the following input file, `input.json`

```json
{
  "rawSpec": "\nname: text-only-output-4\ndescription: Just a test campaign spec to test text only output\n\non:\n  - repository: github.com/sourcegraph/automation-testing\n  # - repository: github.com/sourcegraph-testing/mkcert\n  # - repository: github.com/sourcegraph-testing/tidb\n  # - repository: github.com/sourcegraph-testing/titan\n  # - repository: github.com/sourcegraph-testing/etcd\n  # - repository: github.com/sourcegraph-testing/tiny-go-test-repository\n\nsteps:\n  - run: |\n      echo \"this is step 1\" >> README.txt\n      echo \"this is step 1\"\n      echo \"this is step 1\" 1>&2\n    container: alpine:3\n  - run: |\n      echo \"this is step 2\" >> README.md\n      echo \"this is step 2\"\n      echo \"this is step 2\" 1>&2\n    container: alpine:3\n  - run: |\n      echo \"this is step 3\" >> README.md\n      echo \"this is step 3\"\n      echo \"this is step 3\" 1>&2\n    container: alpine:3\n    outputs:\n      myOutput:\n        value: \"my-output.txt\"\n  - run: |\n      echo \"this is step 4\" >> README.md\n      cat README.md\n      echo \"previous_step.modified_files=${{ previous_step.modified_files }}\" >> README.md\n      ls -lrt\n    container: alpine:3\n  - run: echo \"this is step 5\" >> ${{ outputs.myOutput }}\n    container: alpine:3\n\nchangesetTemplate:\n  title: Testing step caching\n  body:  Testing step caching\n  branch: thorsten/text-only-output-2\n  published: true\n  commit:\n    message: Caching testing\n",
  "workspaces": [
    {
      "repository": {
        "id": "UmVwb3NpdG9yeTo3",
        "name": "github.com/sourcegraph/automation-testing"
      },
      "branch": {
        "abbrevName": "master",
        "target": {
          "oid": "d559e66a56e6615759f60f543475fd69e347f592"
        }
      },
      "path": "",
      "onlyFetchWorkspace": false,
      "steps": [
        {
          "run": "echo \"this is step 1\" >> README.txt\necho \"this is step 1\"\necho \"this is step 1\" 1>&2",
          "container": "alpine:3",
          "env": {}
        }
      ],
      "searchResultPaths": []
    }
  ]
}
```

we can use `src batch exec` to execute the `workspaces.steps` (not the `steps` in the included raw batch spec!!) in the given `workspaces`:

```
src batch exec -f input.json
```

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -15,7 +15,6 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/go-multierror"
-
 	batcheslib "github.com/sourcegraph/sourcegraph/lib/batches"
 
 	"github.com/sourcegraph/src-cli/internal/api"
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/go-multierror"
+
 	batcheslib "github.com/sourcegraph/sourcegraph/lib/batches"
 
 	"github.com/sourcegraph/src-cli/internal/api"
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -49,27 +49,39 @@ type batchExecuteFlags struct {
 	textOnly bool
 }
 
-func newBatchExecuteFlags(flagSet *flag.FlagSet, cacheDir, tempDir string) *batchExecuteFlags {
+func newBatchExecuteFlags(flagSet *flag.FlagSet, workspaceExecution bool, cacheDir, tempDir string) *batchExecuteFlags {
 	caf := &batchExecuteFlags{
 		api: api.NewFlags(flagSet),
 	}
 
-	flagSet.BoolVar(
-		&caf.textOnly, "text-only", false,
-		"INTERNAL USE ONLY. EXPERIMENTAL. Switches off the TUI to only print JSON lines.",
-	)
-	flagSet.BoolVar(
-		&caf.allowUnsupported, "allow-unsupported", false,
-		"Allow unsupported code hosts.",
-	)
-	flagSet.BoolVar(
-		&caf.allowIgnored, "force-override-ignore", false,
-		"Do not ignore repositories that have a .batchignore file.",
-	)
-	flagSet.BoolVar(
-		&caf.apply, "apply", false,
-		"Ignored.",
-	)
+	if !workspaceExecution {
+		flagSet.BoolVar(
+			&caf.textOnly, "text-only", false,
+			"INTERNAL USE ONLY. EXPERIMENTAL. Switches off the TUI to only print JSON lines.",
+		)
+		flagSet.BoolVar(
+			&caf.allowUnsupported, "allow-unsupported", false,
+			"Allow unsupported code hosts.",
+		)
+		flagSet.BoolVar(
+			&caf.allowIgnored, "force-override-ignore", false,
+			"Do not ignore repositories that have a .batchignore file.",
+		)
+		flagSet.BoolVar(
+			&caf.apply, "apply", false,
+			"Ignored.",
+		)
+		flagSet.BoolVar(
+			&caf.keepLogs, "keep-logs", false,
+			"Retain logs after executing steps.",
+		)
+		flagSet.StringVar(
+			&caf.namespace, "namespace", "",
+			"The user or organization namespace to place the batch change within. Default is the currently authenticated user.",
+		)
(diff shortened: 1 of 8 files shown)


cmd/src/batch_common.go
```diff
@@ -48,27 +49,39 @@ type batchExecuteFlags struct {
...
+			"Do not ignore repositories that have a .batchignore file.",
+		)
+		flagSet.BoolVar(
+			&caf.apply, "apply", false,
+			"Ignored.",
+		)
+		flagSet.BoolVar(
+			&caf.keepLogs, "keep-logs", false,
+			"Retain logs after executing steps.",
+		)
```
eseliger, line 77: technically this flag could be of use, if we ever allowed something like "ssh into container when failed" or something like that.

internal/batches/ui/tui.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/go-multierror"
 	"github.com/neelance/parallel"
+
 	"github.com/sourcegraph/sourcegraph/lib/output"
```
eseliger, line 12: nitpick: this seems wrong 🤔

cmd/src/batch_exec.go
```diff
@@ -0,0 +1,252 @@
...
+						OID: w.Branch.Target.OID,
+					},
+				},
+				ExternalRepository: struct{ ServiceType string }{ServiceType: "github"},
+				DefaultBranch: &graphql.Branch{
+					Name: w.Branch.Name,
+					Target: graphql.Target{
+						OID: w.Branch.Target.OID,
+					},
+				},
```
eseliger, line 241: Not even sure if this is required. Ideally, this field is never used directly but accessed through `Rev()` and `BaseRef()`

cmd/src/batch_exec.go
```diff
@@ -0,0 +1,252 @@
...
+		if merr, ok := err.(*multierror.Error); ok {
+			opts.ui.ParsingBatchSpecFailure(merr)
+			return cmderrors.ExitCode(2, nil)
+		} else {
+			// This shouldn't happen; let's just punt and let the normal
+			// rendering occur.
+			return err
+		}
+	}
+	opts.ui.ParsingBatchSpecSuccess()
```
eseliger, line 124: idea: We could share these with batch_common, if we made these separate functions. Might also read nicer in the main loop and reduces duplication:

```
parsingBatchSpec(ui, batchSpec)
preparingContainerImages(ui, svc, batchSpec)
..
```

eseliger (approved):
🚀

cmd/src/batch_common.go
mrnugget: But we do log more via the JSON log lines than we ever did in these log files. So I don't think they'd be useful or am I missing a usecase?

cmd/src/batch_exec.go
mrnugget: Removed that and the `ExternalRepository`.

cmd/src/batch_common.go
eseliger: Oh is this only stdout/stderr? Then nvm

internal/batches/ui/tui.go
mrnugget: Fixed it... This is so confusing to me.

cmd/src/batch_common.go
mrnugget: That and a bunch of other things, but not more useful than the JSON output: https://sourcegraph.com/search?q=context:%40sourcegraph/all+file:executor+Logf%28+repo:src-cli&patternType=literal

eseliger: great :)

cmd/src/batch_exec.go
mrnugget: Okay, we can try that, but let's do that in another PR, since that's tricky stuff with some of the error handling etc.

PR #611, opened 2021-09-06
Start/Wait on exec.Cmd when using pipes

mrnugget:
This might've been the cause for the flaky test we saw (example failure [here](https://github.com/sourcegraph/src-cli/runs/3507335874)).

The docs say that `cmd.Run()` in combination with `StdoutPipe`/`StderrPipe` is wrong:

> `Wait` will close the pipe after seeing the command exit, so most callers
> need not close the pipe themselves. It is thus incorrect to call `Wait`
> before all reads from the pipe have completed. For the same reason, it
> is incorrect to call `Run` when using `StdoutPipe`. See the example for
> idiomatic usage.

See: https://pkg.go.dev/os/exec#example-Cmd.StdoutPipe

### DIFF
--- a/internal/batches/executor/run_steps.go
+++ b/internal/batches/executor/run_steps.go
@@ -313,8 +313,6 @@ func executeSingleStep(
 		cmd.Dir = *dir
 	}
 
-	var stdoutBuffer, stderrBuffer bytes.Buffer
-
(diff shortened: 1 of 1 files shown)


eseliger (approved):
Great catch! Even if it doesn't fix the flake, this is a good improvement!

PR #612, opened 2021-09-06
Streamline and improve JSON UI

eseliger:
Next step I will move the step output definition into the shared lib, so we can use it server-side for parsing.

This requires a server-side change which I am going to publish when we release SRC CLI so we have a version match in the executors, too.

### DIFF
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -152,9 +152,9 @@ func (svc *Service) CreateChangesetSpec(ctx context.Context, spec *batcheslib.Ch
 //
 // Progress information is reported back to the given progress function: perc
 // will be a value between 0.0 and 1.0, inclusive.
-func (svc *Service) EnsureDockerImages(ctx context.Context, spec *batcheslib.BatchSpec, progress func(perc float64)) (map[string]docker.Image, error) {
-	total := len(spec.Steps) + 1
-	progress(0)
+func (svc *Service) EnsureDockerImages(ctx context.Context, spec *batcheslib.BatchSpec, progress func(done, total int)) (map[string]docker.Image, error) {
+	total := len(spec.Steps)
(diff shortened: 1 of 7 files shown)


internal/batches/service/service.go
```diff
@@ -152,9 +152,9 @@ func (svc *Service) CreateChangesetSpec(ctx context.Context, spec *batcheslib.Ch
 //
 // Progress information is reported back to the given progress function: perc
 // will be a value between 0.0 and 1.0, inclusive.
-func (svc *Service) EnsureDockerImages(ctx context.Context, spec *batcheslib.BatchSpec, progress func(perc float64)) (map[string]docker.Image, error) {
-	total := len(spec.Steps) + 1
```
eseliger, line 156: This was off by 1 always 🤔

internal/batches/ui/json_lines.go
```diff
@@ -17,142 +21,167 @@ import (
...
+		logOperationSuccess(LogEventOperationResolvingRepositories, map[string]interface{}{"ignored": len(ignored)})
 	} else {
-		switch len(repos) {
-		case 0:
-			logOperationSuccess("RESOLVING_REPOSITORIES", "No repositories resolved")
-		case 1:
-			logOperationSuccess("RESOLVING_REPOSITORIES", "Resolved 1 repository")
-		default:
-			logOperationSuccess("RESOLVING_REPOSITORIES", fmt.Sprintf("Resolved %d repositories", len(repos)))
-		}
```
eseliger, line 78: This should be handled by the UI, not src cli (for SSBC).

mrnugget: Agree, that was just lazy copy&paste from me.

mrnugget (approved):

PR #625, opened 2021-10-06
internal/batches: include the url field in repositories

LawnGnome:
Removing this in #598 broke the `src batch repos` command, which used the `url` field when rendering the link to the repo in Sourcegraph.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- `src batch repos` failed with a template error in src-cli 3.31.1 and 3.32.0. This has been fixed. [#625](https://github.com/sourcegraph/src-cli/pull/625)
+
 ### Removed
 
 ## 3.31.1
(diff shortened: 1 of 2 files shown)


eseliger:
@mrnugget should we include this in the exec json files as well then? I removed it before (for one, because I thought it was unused) because I feared that at some point we use this field on the struct because "it's there", but we didn't pass it down in the SSBC path.

mrnugget:
Which files? Do you mean the JSON log line output?

eseliger:
The json input file for the `batch exec` command.

mrnugget:
Hmm, the `url` returned in GraphQL is basically just the name of the repository: 

https://github.com/sourcegraph/sourcegraph/blob/f0e6c5298df55ff77316c49b834d2805ce735236/cmd/frontend/graphqlbackend/repository.go#L252-L254

https://github.com/sourcegraph/sourcegraph/blob/f0e6c5298df55ff77316c49b834d2805ce735236/internal/search/result/repo.go#L43-L49

I wonder whether it wouldn't be confusing if we tried to reconstruct that method just for the sake of passing it down to `src batch exec`, where it'll be unused.

Issue #634, opened 2021-10-20
Support serving bare repositories


andir:
On my Git host I've all the repositories in "bare" format. They have been created using `git init --bare $name` and thus do not have a `.git` folder within them. The repositories don't need a full checkout nor a `.git` folder when being accessed via SSH. Currently serving those repositories doesn't work as the check for `.git` folder within them fails.

On the filesystem the layout looks like this:
```
/data
/data/foo.git
/data/bar.git
/data/project1/zes.git
/data/project2/des.git
...
```

Ideally each of the repositories would appear with their name with the `.git` suffix stripped (`foo`, `bar`, `project1/zes`, ...).

Is this something you'd like in the upstream code base? I wouldn't mind trying to tackle the issue as the original code seems to have enough test coverage for me to be confident in not breaking stuff on the way.

keegancsmith:
It is difficult to detect a bare git repository generally which is why we rely on the `.git` convention. You also run into problems on sub repos.

However, saying that we may have something which works enough of the time and is performant: check for a `HEAD` file and `objects` dir. I feel like this is an acceptable option.

burmudar:
Closing this issue as part of repository maintenance. We are no longer using GitHub issues to track src-cli requests or reports. For future questions or support requests, please contact Sourcegraph Support (mailto:support@sourcegraph.com).

PR #639, opened 2021-10-23
Support serving bare repositories

andir:
This is in response to my own issue (#634) where I was asking for support for serving bare repositories from a directory.

This PR has a couple of changes. Some of them can be seen as a refactoring of sorts. The comments in the original implementation were a bit misleading as they appeared to be the other way around (compared to what the tests wanted). I took the liberty to rewrite them a bit. All of the original tests still pass. I've added one case where a bare repository is created (by suffixing the name of a repo with `.bare`) which seems to work. I am not happy about the repository name suffix hack that I had to add but it also doesn't seem like it is worth the effort to refactor the entire test machinery just because of this.

I am open for feedback on this.

cc @mrnugget & @keegancsmith who where very responsive to my issue

### DIFF
--- a/internal/servegit/serve.go
+++ b/internal/servegit/serve.go
@@ -147,6 +147,24 @@ func isBareRepo(path string) bool {
 	return string(out) != "false\n"
 }
 
+// Check if git thinks the given path is a proper git checkout
+func isGitRepo(path string) bool {
+	if fi, err := os.Stat(path); err != nil || !fi.IsDir() {
+		return false
+	}
+
+	// Executing git rev-parse --git-dir in the root of a worktree returns .git
+	c := exec.Command("git", "rev-parse", "--git-dir")
+	c.Dir = path
+	out, err := c.CombinedOutput()
+
+	if err != nil {
+		return false
+	}
+
+	return string(out) == ".git\n"
+}
+
 // Repos returns a slice of all the git repositories it finds.
 func (s *Serve) Repos() ([]Repo, error) {
 	var repos []Repo
@@ -177,8 +195,7 @@ func (s *Serve) Repos() ([]Repo, error) {
 		//
 		// A directory which also is a repository (have .git folder inside it)
 		// will contain nil error. If it does, proceed to configure.
-		gitdir := filepath.Join(path, ".git")
-		if fi, err := os.Stat(gitdir); err != nil || !fi.IsDir() {
+		if !isGitRepo(path) {
 			s.Debug.Printf("not a repository root: %s", path)
 			return nil
 		}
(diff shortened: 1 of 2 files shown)


internal/servegit/serve.go
```diff
@@ -130,6 +130,49 @@ func (s *Serve) handler() http.Handler {
...
+	}
+
+	return string(out) != "false\n"
+}
+
+// Check if git thinks the given path is a proper git checkout
+func isGitRepo(path string) bool {
+	if !isDir(path) {
+		return false
+	}
```
mrnugget, line 162: I wanted to suggest to pull `isDir(path)` out of these two functions, but then I saw these 3 lines:

https://github.com/sourcegraph/src-cli/pull/639/files#diff-2d1e7e4eecf58c3cd47322de771f3e05f7eba484526fe2e85d35289c2cb8fcbfR192-R194

So, I don't think the `isDir` check (here and in the other function) is even necessary, right?

andir: it is a quick, cheaper way to abort early. I could imagine that with lots of repositories (or folders) this could consume some time and this would at least ensure that we aren't forking/exec'ing very often.

mrnugget: Yes, but:

1. We do `if !isDir(path) { return false }` _twice_ in both of these functions. We can pull it out and do it once.
2. We _already_ check whether the path is a directory by checking `fi.IsDir` in the lines above.

andir: I removed the duplicate checks now.

andir:
Rebased and addressed the latest review comments.

internal/servegit/serve_test.go
```diff
@@ -32,7 +32,7 @@ func TestReposHandler(t *testing.T) {
 		repos: []string{"project1", "project2"},
 	}, {
 		name:  "nested",
-		repos: []string{"project1", "project1/subproject", "project2", "dir/project3"},
+		repos: []string{"project1", "project1/subproject", "project2", "dir/project3", "dir/project4.bare"},
```
keegancsmith, line 35: This doesn't seem like a hack at all, succinct. So I actually like it :) FYI I think there is a more traditional way to indicate a repo is base, and that is if the name has the suffix `.git`. I believe git used to check the suffix to decide, but it has a more robust check these days based on actual presence of git files.

keegancsmith (approved):
LGTM. Sorry about the slow review.

internal/servegit/serve_test.go
andir: Oh yeah, The repos actually end up with `.git` in my deployment. I just felt like being explicit there.

PR #645, opened 2021-11-01
api: add User-Agent header

LawnGnome:
This PR unconditionally changes the User-Agent to `src-cli`, rather than the default Go user-agent, and adds the OS and architecture to src for support and product purposes by default. This can be disabled via an environment variable and/or flag, and is documented accordingly.

Closes sourcegraph/sourcegraph#15769.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -127,6 +127,18 @@ mv /usr/local/bin/src /usr/local/bin/src-cli
 
 You can then invoke it via `src-cli`.
 
+## Telemetry
+
+`src` includes the operating system and architecture in the `User-Agent` header sent to Sourcegraph. For example, running `src` version 3.21.10 on an x86-64 Linux host will result in this header:
+
+```
+src-cli/3.21.10 linux amd64
+```
+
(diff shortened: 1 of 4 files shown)


README.markdown
```diff
@@ -127,6 +127,18 @@ mv /usr/local/bin/src /usr/local/bin/src-cli
...
 
+## Telemetry
+
+`src` includes the operating system and architecture in the `User-Agent` header sent to Sourcegraph. For example, running `src` version 3.21.10 on an x86-64 Linux host will result in this header:
+
+```
+src-cli/3.21.10 linux amd64
+```
+
+To disable this and _only_ send the version, you can set `-user-agent-telemetry=false` for a single command, or set the `SRC_DISABLE_USER_AGENT_TELEMETRY` environment variable to any non-blank string.
```
eseliger, line 138: Currently the logic doesn't even send the version. I think this should match what's in code(and I personally think we always want to know the version)

eseliger (approved):

README.markdown
LawnGnome: Good catch! Yeah, I don't think there's any problem sending the version even if "telemetry" is disabled.

eseliger (approved):

mrnugget (approved):

PR #647, opened 2021-11-02
batches: pull images in parallel

LawnGnome:
Closes https://github.com/sourcegraph/sourcegraph/issues/17577.


internal/batches/service/service.go
```diff
@@ -29,26 +30,34 @@ type Service struct {
...
+		parallelism = 1
+	}
+
 	return &Service{
 		allowUnsupported: opts.AllowUnsupported,
 		allowIgnored:     opts.AllowIgnored,
 		allowFiles:       opts.AllowFiles,
 		client:           opts.Client,
 		imageCache:       docker.NewImageCache(),
+		parallelism:      parallelism,
```
eseliger, line 60: this sorta implies to me that service always needs to know this value, but it's only required for `EnsureDockerImages` - any low effort change we can make to only pass it into there?

internal/batches/service/service.go
```diff
@@ -157,20 +166,74 @@ func (svc *Service) CreateChangesetSpec(ctx context.Context, spec *batcheslib.Ch
 // Progress information is reported back to the given progress function: perc
 // will be a value between 0.0 and 1.0, inclusive.
 func (svc *Service) EnsureDockerImages(ctx context.Context, spec *batcheslib.BatchSpec, progress func(done, total int)) (map[string]docker.Image, error) {
-	total := len(spec.Steps)
+	// Figure out the image names used in the batch spec.
+	names := map[string]struct{}{}
+	for i := range spec.Steps {
+		names[spec.Steps[i].Container] = struct{}{}
+	}
```
eseliger, line 173: just that is already a good improvement hehe

eseliger (approved):
Nice improvement :)

internal/batches/service/service.go
mrnugget: I agree with this. The `Service` and `service.Opts` structs already have a lot of fields and nobody knows how they all relate to each other.

Can't we make `parallelism int` an argument of `EnsureDockerImages`?

internal/batches/service/service.go
```diff
@@ -157,20 +166,74 @@ func (svc *Service) CreateChangesetSpec(ctx context.Context, spec *batcheslib.Ch
...
 	images := make(map[string]docker.Image)
-	for i := range spec.Steps {
-		img, err := svc.EnsureImage(ctx, spec.Steps[i].Container)
-		if err != nil {
-			return nil, err
+	for image := range complete {
+		if image.err != nil {
+			// If EnsureImage errored, then we'll early return here and let the
+			// worker context clean things up.
+			return nil, image.err
```
mrnugget, line 231: Does that work? The goroutines don't check for `ctx.Done()` and could be stucking waiting to put something in `complete` (which is unbuffered) because there's no reader on the other side. And the goroutine that does `wg.Wait()` would be stuck on that.

Example: the `return` here fires but another goroutine just returned from `EnsureImage` above and now tries to put something in `complete`. It will get stuck.

I think you need something like

```go
select {
case <-ctx.Done():
  return
case  complete <- image{name:  name, image: img,err:   err }:
  // all good
}
```

in your goroutine above.

(And possibly another check for `ctx.Done()` before `EnsureImage` so we don't download images if one failed)

LawnGnome: Fixed in 180078d (and tested in 188a328).

LawnGnome:
Going to re-request a review on this, mostly for the tests, and to a lesser extent because I had to merge the implementation with SSBC work that has happened since the original version of this PR.

This probably doesn't need a tonne of love, but if someone could give it a quick once over, I'd appreciate it!

eseliger (approved):

internal/batches/service/service.go
```diff
@@ -154,23 +155,98 @@ func (svc *Service) CreateChangesetSpec(ctx context.Context, spec *batcheslib.Ch
...
+	inputs := make(chan string, total)
+
+	// Set up a worker context that we can use to terminate the workers if an
+	// error occurs.
+	workerCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	// Spawn worker goroutines to call EnsureImage on each image name.
+	if parallelism < 1 {
+		parallelism = 1
```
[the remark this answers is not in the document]
LawnGnome: I thought about it, but I don't think it's any more expensive than actually running the steps, so if someone wants to hammer their network and filesystem pulling and extracting images I figure that's up to them.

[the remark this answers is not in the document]
eseliger: I mean we would just not have anything to pull anyways so it doesn't matter mich, but we spawn a few more goroutines than needed and immediately terminate them. Not worth much effort though 😀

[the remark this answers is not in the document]
LawnGnome: Oh, I misread your comment! I thought you meant a global total, not the total number of images.

Yes, we probably should do that. I'll see about a follow up.

PR #648, opened 2021-11-03
Use extracted batcheslib.BuildChangesetSpecs

mrnugget:
Counterpart to: https://github.com/sourcegraph/sourcegraph/pull/26997


internal/batches/graphql/repository.go
```diff
@@ -66,3 +66,11 @@ func (r *Repository) Rev() string {
 
 	return r.DefaultBranch.Target.OID
 }
+
+func (r *Repository) FileMatchesList() []string {
+	matches := make([]string, 0, len(r.FileMatches))
+	for path := range r.FileMatches {
+		matches = append(matches, path)
+	}
+	return matches
```
eseliger, line 75: I think we should sort these before returning, for caching

eseliger (approved):

internal/batches/graphql/repository.go
mrnugget: Good call. That also wasn't done before: https://github.com/sourcegraph/src-cli/blob/240d3b6f6fdb6eafdf0aa0775e74da57198136a7/internal/batches/util/repo.go#L13-L22

PR #651, opened 2021-11-04
Refactor caching and coordinator

eseliger:
Commits contain change descriptions.


cmd/src/batch_common.go
```diff
@@ -292,29 +292,45 @@ func executeBatchSpec(ctx context.Context, ui ui.ExecUI, opts executeBatchSpecOp
...
+	// Add external changeset specs.
+	importedSpecs, importErr := svc.CreateImportChangesetSpecs(ctx, batchSpec)
+	err = nil
+	if execErr != nil {
+		err = multierror.Append(err, execErr)
+	}
+	if importErr != nil {
+		err = multierror.Append(err, importErr)
+	}
 	if err != nil && !opts.flags.skipErrors {
```
mrnugget, line 334: Does this still work or would you need to do a `err.ErrorOrNil()` after declaring err to be a multierror?

There's also some error-unpacking going on the TUI to pretty-print errors. Does this work with mulitierror?

internal/batches/executor/coordinator_test.go
```diff
@@ -475,7 +421,9 @@ func TestCoordinator_Execute_StepCaching(t *testing.T) {
 	task.CachedResultFound = false
 
 	// Now we execute the spec with -clear-cache:
```
mrnugget, line 423: Comment is outdated

mrnugget (approved):
Nice! Only thing to check is the error handling, because that can be tricky

cmd/src/batch_common.go
mrnugget: Fixed this!

PR #653, opened 2021-11-10
Simplify caching logic in 'src batch exec'

mrnugget:
This is part of https://github.com/sourcegraph/sourcegraph/issues/26929
and simplifies the caching-related logic in `src batch exec` to only
what's needed: checking for per-step cache results.

Since server-side batch changes use server-side caching for changeset
specs we don't need to check for changeset specs.


eseliger (approved):
I guess technically we would want to remove the cache result if any has a full result stored, because the JSON logs UI would otherwise be broken. but doesn't matter so much I think, in firecracker the disk is wiped every time anyways.

mrnugget:
Restored the old logic in case something is not wiped in firecracker/locally and we need to debug

eseliger (approved):

PR #658, opened 2021-11-23
ssbc: Allow cache to read step results from current working directory

eseliger:


internal/batches/executor/execution_cache.go
```diff
@@ -157,16 +157,16 @@ type JSONCacheWriter interface {
 	WriteAfterStepResult(key string, value execution.AfterStepResult)
 }
 
-type JSONLinesCache struct {
+type SSBCCache struct {
```
mrnugget, line 160: Plz no acronyms!

```suggestion
type ServerSideCache struct {
```

mrnugget (approved):

internal/batches/executor/execution_cache.go
eseliger: Says the guy who named it _JSON_ lines cache before 😆

mrnugget: `JavaScriptObjectNotationLines` killed the linter

PR #659, opened 2021-11-23
Fix off-by-one error in JSON UI for skipping steps

eseliger:
Steps are 1-indexed in our JSON UI, but we forgot to add the obligatory +1 here that we do everywhere else.


mrnugget (approved):
I _trust_ you that this works because my brain can't tell

eseliger:
I trust my thorough debugging of this that took _at least_ 30 seconds.

PR #660, opened 2021-11-24
batches: support volume mode on Apple Silicon

courier-new:
Tested basic "hello world" batch spec with volume mount mode (both setting the workspace flag and on auto with the new default condition) and didn't see any issues with the execution or the diffs produced. The new condition will default to volume mode for Intel and M1 Macs and keep bind mode as the default for anything else. I updated some comments in the area to capture the discussion @LawnGnome and I had about eventually moving to volume mode as the default for everyone.

Closes https://github.com/sourcegraph/sourcegraph/issues/27549

### DIFF
--- a/internal/batches/workspace/workspace.go
+++ b/internal/batches/workspace/workspace.go
@@ -95,9 +95,9 @@ func BestCreatorType(ctx context.Context, images map[string]docker.Image) Creato
 	// if you have a batch change with steps that run as UID 1000 and then UID
 	// 2000, you'll get errors when the second step tries to write.
 
-	// For the time being, we're only going to consider volume mode on Intel
-	// macOS.
-	if runtime.GOOS != "darwin" || runtime.GOARCH != "amd64" {
+	// For the time being, we're only going to consider volume mode on Intel or
+	// M1 macOS.
+	if runtime.GOOS != "darwin" || (runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64") {
 		return CreatorTypeBind
 	}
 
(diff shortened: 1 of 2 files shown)


LawnGnome (approved):
:speedboat:

internal/batches/workspace/workspace.go
```diff
@@ -95,9 +95,12 @@ func BestCreatorType(ctx context.Context, images map[string]docker.Image) Creato
...
 
-	// For the time being, we're only going to consider volume mode on Intel
-	// macOS.
-	if runtime.GOOS != "darwin" || runtime.GOARCH != "amd64" {
+	// NOTE: For the time being, we're only going to consider volume mode on Intel or M1
+	// macOS. As we've generally not observed many issues from users using volume mode, and
+	// it is the faster of the options on non-Linux platforms (and only a touch slower on
+	// Linux itself), we should consider making it the default in the future for the sake of
+	// consistency.
+	if runtime.GOOS != "darwin" || (runtime.GOARCH != "amd64" && runtime.GOARCH != "arm64") {
```
LawnGnome, line 103: Maybe we should just remove the `GOARCH` checks altogether at this point, since macOS doesn't run anywhere else anyway.

eseliger (approved):

internal/batches/workspace/workspace.go
eseliger: 👍

courier-new: I kinda like keeping it in to make things explicit, juuuuuust in case someone comes in with a weird franken-machine that somehow is running on a different architecture. 😂

eseliger: I don't think Apple would ever distribute other binaries, but doesn't matter much to me 😁

LawnGnome: Yeah, I'm not too fussed, since I think we'll probably remove this check altogether soon, but I also think it's unlikely to be an issue in practice if we don't have it.

courier-new: sure sure. 🙂 I'll just leave it for now since, as Adam mentions, we'll probably remove this check entirely the next time we're thinking about it, anyway.

PR #661, opened 2021-11-24
Build changeset specs server side

eseliger:
This PR changes src exec's behavior to not upload changeset specs anymore. They aren't required anymore. 

See main PR for more details: https://github.com/sourcegraph/sourcegraph/pull/28158

### DIFF
--- a/cmd/src/batch_exec.go
+++ b/cmd/src/batch_exec.go
@@ -8,7 +8,6 @@ import (
 	"io"
 
 	"github.com/cockroachdb/errors"
-	"github.com/hashicorp/go-multierror"
 
 	"github.com/sourcegraph/src-cli/internal/batches/executor"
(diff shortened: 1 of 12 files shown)


cmd/src/batch_exec.go
```diff
@@ -109,27 +108,11 @@ func executeBatchSpecInWorkspaces(ctx context.Context, ui *ui.JSONLines, opts ex
...
-			ui.ParsingBatchSpecFailure(multiErr)
-			return cmderrors.ExitCode(2, nil)
-		} else {
-			// This shouldn't happen; let's just punt and let the normal
-			// rendering occur.
-			return err
-		}
-	}
-	ui.ParsingBatchSpecSuccess()
-
```
mrnugget, line 127: I love that this is now also gone. It was confusing to read that we're parsing this here (but we also parse it server-side, etc.)

mrnugget (approved):

PR #662, opened 2021-11-25
batches: support multiple target branches

LawnGnome:
The client side part of sourcegraph/sourcegraph#25228. Depends on https://github.com/sourcegraph/sourcegraph/pull/28168, and will not pass CI until that's merged and this PR gets an update to its `go.mod`.


internal/batches/service/workspaces.go
```diff
@@ -35,11 +35,6 @@ func findWorkspaces(
 	finder directoryFinder,
 	repos []*graphql.Repository,
 ) ([]RepoWorkspace, error) {
-	repoByID := make(map[string]*graphql.Repository)
-	for _, repo := range repos {
-		repoByID[repo.ID] = repo
-	}
```
LawnGnome, line 41: Fun fact: `repoByID` was never used again.

internal/batches/service/service.go
```diff
@@ -442,13 +442,17 @@ func (svc *Service) ResolveNamespace(ctx context.Context, namespace string) (str
 }
 
 func (svc *Service) ResolveRepositories(ctx context.Context, spec *batcheslib.BatchSpec) ([]*graphql.Repository, error) {
-	seen := map[string]*graphql.Repository{}
+	type seenRepo struct {
+		overridable bool
+		revisions   []*graphql.Repository
+	}
+	seen := map[string]*seenRepo{}
```
LawnGnome, line 449: This is obviously eerily similar to code in https://github.com/sourcegraph/sourcegraph/pull/28168 — my not-so-secret plan here is to unify these once we have generics available in Go, since the only real distinction is that the "revision" type is different in `src` and SSBC.

LawnGnome:
I'm opening this up for early review, but note the... uh, note in the PR text about this not being able to pass until the corresponding Sourcegraph PR lands.

mrnugget (approved):

internal/batches/service/workspaces.go
eseliger: Where you at golangci-lint when we need you 🧐

eseliger (approved):
In general LGTM, pending my comment on the algorithm on the main repo PR.

PR #664, opened 2021-11-26
Backport fix for workspace globbing

eseliger:
See https://github.com/sourcegraph/sourcegraph/pull/28230 for more details.

### DIFF
--- a/internal/batches/service/workspaces.go
+++ b/internal/batches/service/workspaces.go
@@ -43,9 +43,14 @@ func findWorkspaces(
 	// Pre-compile all globs.
 	workspaceMatchers := make(map[batcheslib.WorkspaceConfiguration]glob.Glob)
 	for _, conf := range spec.Workspaces {
-		g, err := glob.Compile(conf.In)
+		in := conf.In
+		// Empty `in` should fall back to matching all, instead of nothing.
+		if in == "" {
+			in = "*"
+		}
+		g, err := glob.Compile(in)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


LawnGnome (approved):
Approved, but same caveat as the main repo approval.

internal/batches/service/workspaces.go
```diff
@@ -43,9 +43,14 @@ func findWorkspaces(
 	// Pre-compile all globs.
 	workspaceMatchers := make(map[batcheslib.WorkspaceConfiguration]glob.Glob)
 	for _, conf := range spec.Workspaces {
-		g, err := glob.Compile(conf.In)
+		in := conf.In
+		// Empty `in` should fall back to matching all, instead of nothing.
```
mrnugget, line 47: Nitpick, but just wanted to mention it: we could also define something like

```go
type allMatcher struct{}
func (a allMatcher) Match(_ string) bool { return true }
```

and pass that in when `in == ""` since it satifies `glob.Glob`. We'd save a `glob.Compile`.

internal/batches/service/workspaces_test.go
```diff
@@ -118,6 +118,24 @@ func TestFindWorkspaces(t *testing.T) {
 				{Repo: repos[2], Steps: steps, Path: "d/e/f", OnlyFetchWorkspace: true},
 			},
 		},
+		"workspace configuration without in matches all": {
```
mrnugget, line 121:
```suggestion
		"workspace configuration without 'in' matches all": {
```

mrnugget (approved):

PR #666, opened 2021-11-26
go.mod: pull in sourcegraph/sourcegraph#28240 fix

LawnGnome:
As promised, this includes a changelog entry.


eseliger (approved):

mrnugget:
Thanks! I'm wondering whether a test in https://github.com/sourcegraph/src-cli/blob/main/internal/batches/ui/task_exec_tui_test.go could catch this, since it seems we'd only need to trigger the "shutdown"?

PR #668, opened 2021-11-30
batches: test the fix in #666

LawnGnome:
@mrnugget suggested this test, and quite right he was, too.

### DIFF
--- a/internal/batches/ui/task_exec_tui_test.go
+++ b/internal/batches/ui/task_exec_tui_test.go
@@ -48,14 +48,6 @@ func TestTaskExecTUI_Integration(t *testing.T) {
 	advanceClock := func(d time.Duration) { now = now.Add(d) }
 
 	buf := &ttyBuf{}
-	expectOutput := func(t *testing.T, buf *ttyBuf, want []string) {
-		t.Helper()
-
-		have := buf.Lines()
-		if !cmp.Equal(want, have) {
-			t.Fatalf("wrong output:\n%s", cmp.Diff(want, have))
-		}
-	}
 
 	out := output.NewOutput(buf, output.OutputOpts{
 		ForceTTY:    true,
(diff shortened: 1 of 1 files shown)


internal/batches/ui/task_exec_tui_test.go
```diff
@@ -205,6 +197,70 @@ func TestTaskExecTUI_Integration(t *testing.T) {
 	})
 }
 
+func TestProgressUpdateAfterComplete(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("Something emits different escape codes on windows.")
```
eseliger, line 202: Of course 😆

eseliger (approved):

internal/batches/ui/task_exec_tui_test.go
LawnGnome: In fairness, I copy/pasted that from the other integration test — but only after it failed on Windows. :window:

internal/batches/ui/task_exec_tui_test.go
```diff
@@ -205,6 +197,70 @@ func TestTaskExecTUI_Integration(t *testing.T) {
...
+		"├── github.com/sourcegraph/sourcegraph  echo Hello World > README.md          0s",
+		"└── github.com/sourcegraph/src-cli      Downloading archive                   0s",
+		"",
+	})
+
+	// Now mark the progress as complete.
+	printer.progress.Complete()
+
+	// Now send another update. This would panic before the relevant fix was
+	// merged in #666.
```
mrnugget, line 250: I just realised we forgot to celebrate #666, the PR of the beast, accordingly (by doing a little ritual in the comments or something) :(

mrnugget (approved):

internal/batches/ui/task_exec_tui_test.go
LawnGnome: From the current changelog:

image

mrnugget: Nice, that's the spirit (we summoned)

PR #671, opened 2021-12-15
batches: add `src batch remote -f spec.yml` for SSBC execution

LawnGnome:
This PR depends on sourcegraph/sourcegraph#30448.

Given the feedback from the draft version of this PR, the new command is `src batch remote -f spec.yml`. I promise that I'll figure out how to have commander handle optional sub-sub-commands if we decide to add a `src batch remote list` or similar later.

Although it's a few hundred lines, this is basically 95% boilerplate — the bulk of the logic is really in the backend PR, and backend service methods that were created early in the SSBC development process. We just have lots of (IMHO) uninteresting code to set up the command, wrap the GraphQL queries and mutations, and report progress.

Closes sourcegraph/sourcegraph#26503.


cmd/src/batch_remote.go
```diff
@@ -0,0 +1,37 @@
...
+)
+
+var batchRemoteCommands commander
+
+func init() {
+	usage := `'src batch remote' manages server side batch changes.
+
+Usage:
+
+    src batch remote command [command options]
```
mrnugget, line 15:
```suggestion
    src batch remote <command> [command options]
```

LawnGnome:
Just to reiterate the draftiness of all this: I think I'm going to move the get/create/replace logic into a new mutation on the backend, after sleeping on it. So, again, don't get too hung up on how this works just yet.

cmd/src/batch_remote.go
mrnugget: (We also did this wrong in other commands, so nitpicky)

mrnugget:
> But, there's one thing that is ready for us all to talk about: what noun and verb do we use in the command structure? I've chosen `remote` and `run`, respectively. `run` matches the UI. `remote` is there because users won't really think of "server" or "SSBC" as concepts, but "remote" feels like a good description of what's going on.

Why do we need a `noun` and a `verb`? I think `src batch run-remote -f foobar.yml` is perfectly fine. Or `remote-run`. Or just `src batch remote -f foobar.yml`? I also can't think of any other commands that would be nested under `remote` (and that wouldn't work as other standalone commands)

> An open question: should the existing commands move to a `src batch local` namespace, with everything aliased for now to make it work?

IMHO: No

> Another open question: should there even _be_ namespaces there? Malo's original concept in sourcegraph/sourcegraph#26503 used `-remote` or `-executor`, but I'm not convinced by this: I think it's going to be too easy for users to be confused about where their batch specs are being executed. I _could_ be convinced, though.

I'd use a different command, not a flag.

LawnGnome:
> Why do we need a `noun` and a `verb`? I think `src batch run-remote -f foobar.yml` is perfectly fine. Or `remote-run`. Or just `src batch remote -f foobar.yml`? I also can't think of any other commands that would be nested under `remote` (and that wouldn't work as other standalone commands)

I could see status commands being under `src batch remote` as well: something like `src batch remote status <ID>` feels plausible. That said, I'm struggling enough to think of a use case that we probably shouldn't decide based on that. :slightly_smiling_face: So I could be convinced on `src batch remote -f foobar.yml`, I think.

I don't love `run-remote` or `remote-run` — my concern here is that, conceptually, executing a batch change locally also implies that it's being "run". It feels like a weird special case that we only have this for remote.

@malomarrec, since you wrote the original issue here with a flag, what are your thoughts?

malomarrec:
>I could see status commands being under src batch remote as well: something like src batch remote status <ID> feels plausible. That said, I'm struggling enough to think of a use case that we probably shouldn't decide based on that. 🙂 So I could be convinced on src batch remote -f foobar.yml, I think.

I could see this being integrated in CI or another automation, and users using SSBC as part of a longer toolchain and scripting around it. That said, none of the options are incompatible with this, it's just that having `src batch list-remote` and `src batch get-remote <id>` etc etc sounds a bit heavy.

malomarrec:
One reason I had in adding the `remote` flag as opposed to changing the command was that, in my mind, running locally or remotely are the same operation, so they should be expressed similarly (with a modifier). But that's a debatable UX opinion.

courier-new:
I'm also on board with the proposal for `src batch remote apply` and `src batch remote preview`. Beyond even just pinging the status in CI or something, it seems to me that working with remote machines comes with a whole new set of concerns that `src` never previously had to worry about when everything was just running locally, and I'd hate to imagine that one day we want 20+ remote-specific commands that all have to be named `verb-remote`.

I also would have voted for using the `src batch local` namespace for existing commands, though, so I'm curious to hear other opinions on this. My thought is that we've agreed that as a product we want SSBC to be the new default, and I would expect alignment on that between the web GUI and the CLI. If `src batch apply` executes locally and I have to use a separate namespace for SSBC commands, though, that's at odds with "SSBC is the default".

Personally, I'd love if users were able to configure which environment the alias `src batch apply|preview` would target, so that someone who always wants to execute them locally and someone who always wants to execute them on a remote executor could just set up the command alias for `local|remote` once and then forget it. For parity with the current version of src-cli, the default could be local, but in some future version when SSBC goes GA, this default could potentially switch to SSBC.

mrnugget:
My question regarding `src batch remote [apply|preview]`: what would those do? Create the execution and then show progress until it's finished and _then_ apply or preview the batch spec?

(I somehow assumed we're only talking about a command that "submits" a batch spec to Sourcegraph and prints a URL to watch the execution.)

> I also would have voted for using the `src batch local` namespace for existing commands, though, so I'm curious to hear other opinions on this. My thought is that we've agreed that as a product we want SSBC to be the new default, and I would expect alignment on that between the web GUI and the CLI. If `src batch apply` executes locally and I have to use a separate namespace for SSBC commands, though, that's at odds with "SSBC is the default".

My thinking here is kinda: the CLI is a local tool (and not required for SSBC) so local execution should be the first-class citizen there and on Sourcegraph itself SSBC is the #1 thing, so it's the recommended way there.

internal/batches/service/remote.go
```diff
@@ -0,0 +1,138 @@
+package service
+
+import (
+	"context"
+
+	"github.com/cockroachdb/errors"
+)
+
+var ErrServerSideBatchChangesUnsupported = errors.New("server side batch changes are not available on this Sourcegraph instance")
```
courier-new, line 9: I had a meta question about the user-facing way that refer to the feature known internally as SSBC, but it's not at all a specific concern about this PR, so I realized it's probably better suited to a team sync and will put it there instead. 🙂

cmd/src/batch_remote.go
```diff
@@ -0,0 +1,138 @@
...
+			} else if res.State == "COMPLETED" {
+				break
+			}
+		}
+		ui.ResolvingWorkspacesSuccess()
+
+		// We have to enqueue this for execution with a separate operation.
+		//
+		// TODO: when the execute flag is wired up in the upsert mutation, just set
+		// it there and remove this.
```
eseliger, line 115: is this still a todo?

eseliger (approved):

cmd/src/batch_remote.go
LawnGnome: [Yep.](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@5a1824e22366d4e7ff5e4f0776509993bf1a9e3a/-/blob/cmd/frontend/graphqlbackend/batches.graphql?L1405:9)

eseliger: Oh, I think I misunderstood the comment. Makes sense!

PR #674, opened 2022-01-04
search: use .api/search/stream

stefanhengl:
Relates to https://github.com/sourcegraph/sourcegraph/issues/29399

This updates stream search to call .api/search/stream instead of
/search/stream. We plan to retire /search/stream as soon as all clients
have been migrated.

### DIFF
--- a/internal/streaming/search.go
+++ b/internal/streaming/search.go
@@ -21,7 +21,7 @@ type Opts struct {
 // response body.
 func Search(query string, opts Opts, client api.Client, decoder Decoder) error {
 	// Create request.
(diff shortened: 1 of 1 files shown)


tsenart (approved):
Have you tested this? Are these two routes identical (middleware and all)?

stefanhengl:
> Have you tested this? Are these two routes identical (middleware and all)?

I have tested unauthenticated requests against Cloud and token-based requests against a local instance in private mode. Both work fine. [The two routes (API vs APP) are not identical ](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/internal/cli/http.go?L37:6&subtree=true) but for the purpose of src-cli they should be equivalent.

PR #675, opened 2022-01-10
Handle missing flags and correctly allow files in repos and validate …

eseliger:
…commands

@mrnugget I tagged you here too as we two touched this last. Don't feel urged to review though if you don't have the capacity.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,12 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Removed
 
+## 3.35.2
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


courier-new (approved):
Thanks for fixing this!! It seems we haven't typically had much of a testing strategy when it comes to src-cli. Is this something we've had discussion about before?

eseliger:
A few dozen times 😆 We always wanted to get e2e tests set up but then the bug count was so low that it didn't really justify it, and ever since it comes up each times we have a bug but never in between.. 😬 Some day!

LawnGnome (approved):

mrnugget (approved):

PR #678, opened 2022-01-12
batches: match on abbreviated commit hash in version

courier-new:
Follow up to https://github.com/sourcegraph/sourcegraph/pull/29654, the version string for Sourcegraph in certain pre-release environments now will feature the 12-character abbreviated commit hash at the end.

See https://sourcegraph.slack.com/archives/C01N83PS4TU/p1642013850061100 for more.

Fixes [29618](https://github.com/sourcegraph/sourcegraph/issues/29618).

### DIFF
--- a/internal/api/version_check.go
+++ b/internal/api/version_check.go
@@ -22,7 +22,10 @@ func CheckSourcegraphVersion(version, constraint, minDate string) (bool, error) 
 		return true, nil
 	}
 
-	buildDate := regexp.MustCompile(`^\d+_(\d{4}-\d{2}-\d{2})_[a-z0-9]{7}$`)
+	// This should match the version format returned by Sourcegraph, which uses the 12
+	// character abbreviated commit hash:
+	// https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/enterprise/dev/ci/internal/ci/config.go?L96.
+	buildDate := regexp.MustCompile(`^\d+_(\d{4}-\d{2}-\d{2})_[a-z0-9]{12}$`)
 	matches := buildDate.FindStringSubmatch(version)
 	if len(matches) > 1 {
 		return matches[1] >= minDate, nil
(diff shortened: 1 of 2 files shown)


internal/api/version_check.go
```diff
@@ -22,7 +22,10 @@ func CheckSourcegraphVersion(version, constraint, minDate string) (bool, error)
 		return true, nil
 	}
 
-	buildDate := regexp.MustCompile(`^\d+_(\d{4}-\d{2}-\d{2})_[a-z0-9]{7}$`)
+	// This should match the version format returned by Sourcegraph, which uses the 12
+	// character abbreviated commit hash:
+	// https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/enterprise/dev/ci/internal/ci/config.go?L96.
+	buildDate := regexp.MustCompile(`^\d+_(\d{4}-\d{2}-\d{2})_[a-z0-9]{12}$`)
```
LawnGnome, line 28: This needs to continue to accept the old format as well for backward compatibility. Since we don't actually use the revhash for anything, how about:

```suggestion
	buildDate := regexp.MustCompile(`^\d+_(\d{4}-\d{2}-\d{2})_[a-z0-9]{7,}$`)
```

Which should accept anything seven characters or longer. (Git is planning to transition to a 64 character ID format, so I don't really want to put a length constraint on there, provided it's the last element in the version.)

If we do this, let's also add test cases for the common lengths (7, 12, and 40, at least).

LawnGnome (changes requested):
I think we'll need to make a change for BC reasons, but otherwise this looks good in terms of implementation. Good job!

internal/api/version_check.go
courier-new: That works for me, thanks for the suggestion!

LawnGnome (approved):
Thanks! Approved, pending the check failure going green.

PR #681, opened 2022-01-20
WIP: Script for Validating Sourcegraph Instance

mucles:
Help validating instance by performing the following checks ( from a custom json / yaml file or will run defualt values if none are specified): - Add a code host - clone a repo - make a search for the repo - make an unindexed search on the repo. *  fixes : the SleepBetweenTriesSeconds field not being read

PR #684, opened 2022-01-20
Update CHANGELOG.md

mucles:
relates to PR'S: https://github.com/sourcegraph/src-cli/pull/681 , https://github.com/sourcegraph/sourcegraph/pull/29991

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to `src-cli` are documented in this file.
 ### Added
 
 ### Changed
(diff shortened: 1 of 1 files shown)


daxmc99:
@sourcegraph/batchers Just to confirm, there are no more steps needed after this. The CLI will be released as part of the next monthly release, correct?

eseliger (approved):
LGTM, in the future you can also include changelog entries in the PR itself, no need to take the effort and create a second one :)

eseliger:
Yes, we sometimes release patch releases when necessary on our end, of if you tell us to, otherwise we will always cut a release close to the branch cut anyways.
Do you think this warrants a patch? It's relatively cheap and we can just tag it as 3.36.2 right away.

mucles:
Yes, this would warrant a patch, Thank you @eseliger :), greatly appreciate your help 🙏 
cc: @mike-r-mclaughlin

PR #687, opened 2022-02-01
batches: make `-f` optional

LawnGnome:
New behaviour (assuming `COMMAND` is any command that accepted `-f`, including `apply`, `preview`, `repos`, or `validate`):

* `src batch COMMAND` **on a TTY**: will wait for 250ms in case there's input on stdin; otherwise errors
* `src batch COMMAND` with a file redirected in: waits indefinitely for input
* `src batch COMMAND -f -`: waits indefinitely for input
* `src batch COMMAND FILE`: reads from `FILE`; equivalent to `src batch COMMAND -f FILE`
* `src batch COMMAND -f FILE`: works as before

In summary, this shouldn't break backward compatibility in any case except for the user _slowly_ typing in a batch spec on stdin.

Fixes #680.

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/cockroachdb/errors"
 	"github.com/hashicorp/go-multierror"
+	"github.com/mattn/go-isatty"
 
 	batcheslib "github.com/sourcegraph/sourcegraph/lib/batches"
 
@@ -97,7 +98,7 @@ func newBatchExecuteFlags(flagSet *flag.FlagSet, workspaceExecution bool, cacheD
 
 	flagSet.StringVar(
 		&caf.file, "f", "",
-		"The batch spec file to read.",
+		"The batch spec file to read, or - to read from standard input.",
 	)
 
 	flagSet.IntVar(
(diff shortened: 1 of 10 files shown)


internal/batches/ui/tty.go
```diff
@@ -0,0 +1,20 @@
+package ui
+
+import (
+	"time"
+
+	"github.com/creack/goselect"
+)
+
+func HasInput(fd uintptr, timeout time.Duration) (bool, error) {
+
```
eseliger, line 10:
```suggestion
```

cmd/src/batch_common.go
```diff
@@ -176,14 +194,31 @@ func batchDefaultTempDirPrefix() string {
 	return os.TempDir()
 }
 
-func batchOpenFileFlag(flag *string) (io.ReadCloser, error) {
-	if flag == nil || *flag == "" || *flag == "-" {
+func batchOpenFileFlag(flag string) (io.ReadCloser, error) {
```
eseliger, line 197: this is also used in the exec path - not sure we want this there? anyways, not super important.

eseliger (approved):

cmd/src/batch_common.go
LawnGnome: It's not particularly relevant, but I also don't think it'll cause problems. We can pull out the open/validate bits we need for `exec` at a later date if we want. (I don't mind doing that, but I also don't think it's on the critical path here.)

eseliger: yeah no, let's leave it as is.

PR #688, opened 2022-02-01
batches: fail earlier on dotcom

courier-new:
This PR is in response to https://github.com/sourcegraph/sourcegraph/issues/28252 and does two things:
- ~Lessens the arbitrarily large default count applied to the batch spec `repositoriesMatchingQuery` from 999999 to 99999, which is less likely to cause timeouts.~ Decided not to proceed with this change since we haven't heard of actual enterprise customers running into timeouts yet and this is expected to become a non-issue once we actually use `count: all`.
- Requests BC-specific information from the version GraphQL query so that it produces an error on dotcom or in an environment where batch changes is not available. This has the effect that `src preview|apply` will fail right away on the first thing it does, which is the version check, as opposed to much later after execution finishes and we go to upload changesets:
```
failed to query Sourcegraph version to check for available features: GraphQL errors: 1 error occurred:
   	* {
     "message": "access to batch changes on Sourcegraph.com is currently not available",
     "path": [
       "batchChanges"
     ]
   }
```

### DIFF
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -763,7 +763,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 
 var defaultQueryCountRegex = regexp.MustCompile(`\bcount:(\d+|all)\b`)
 
-const hardCodedCount = " count:99999"
+const hardCodedCount = " count:999999"
 
 func setDefaultQueryCount(query string) string {
 	if defaultQueryCountRegex.MatchString(query) {
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -57,15 +57,15 @@ func New(opts *Opts) *Service {
 // first thing we do.
 const sourcegraphVersionQuery = `query SourcegraphVersion {
 	site {
-	  productVersion
+		productVersion
 	}
 	batchChanges(first: 1) {
-    nodes {
-      id
-    }
-  }
+		nodes {
+			id
+		}
+	}
 }
-  `
+`
 
 // getSourcegraphVersion queries the Sourcegraph GraphQL API to get the
 // current version of the Sourcegraph instance.
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -52,6 +52,9 @@ func New(opts *Opts) *Service {
 	}
 }
 
+// The reason we ask for batchChanges here is to surface errors about trying to use batch
+// changes in an unsupported environment sooner, since the version check is typically the
+// first thing we do.
 const sourcegraphVersionQuery = `query SourcegraphVersion {
 	site {
 	  productVersion
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -760,7 +760,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 
 var defaultQueryCountRegex = regexp.MustCompile(`\bcount:(\d+|all)\b`)
 
-const hardCodedCount = " count:999999"
+const hardCodedCount = " count:99999"
 
 func setDefaultQueryCount(query string) string {
 	if defaultQueryCountRegex.MatchString(query) {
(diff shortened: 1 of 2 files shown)


internal/batches/service/service.go
```diff
@@ -755,7 +763,7 @@ func (svc *Service) FindDirectoriesInRepos(ctx context.Context, fileName string,
 
 var defaultQueryCountRegex = regexp.MustCompile(`\bcount:(\d+|all)\b`)
 
-const hardCodedCount = " count:999999"
+const hardCodedCount = " count:99999"
```
eseliger, line 766: This is meant to emulate `count:all` which didn't exist back then. In order not to change the behavior here (probably only for the biggest customers, but still) I think we should not tune down this number. That's just how BC is documented to operate on the search. 
I assume this is mainly a problem on dotcom with its massive scale and that's now already covered by the above change, isn't it?

internal/batches/service/service.go
```diff
@@ -52,11 +52,19 @@ func New(opts *Opts) *Service {
...
 const sourcegraphVersionQuery = `query SourcegraphVersion {
 	site {
 	  productVersion
 	}
+	batchChanges(first: 1) {
+    nodes {
+      id
+    }
   }
+}
```
eseliger, line 67: tabs vs spaces war going on here :)

eseliger: I wonder if this is a bit of a dirty hack? I don't have strong feelings about it though. Might be nicer to expose a property "batchChangesEnabled" or something like that that would on dotcom always be falsy and on instances where BC is turned off via site config as well. But it wouldn't work with older SG instances .. so .. well let's not overengineer this 🤷‍♂️

eseliger (approved):
LGTM pending we don't change the count thingy

internal/batches/service/service.go
LawnGnome: Yeah, I might prefer a new field that we can query for this just so we don't have to hit the database, but I don't think that's critical here, particularly since we'd have to "feature" it to depend on a particular version. Could be a nice "first issue" type sustaining ticket for someone else.

LawnGnome (approved):
This needs a changelog entry, but otherwise LGTM modulo the things Erik also said.

internal/batches/service/service.go
courier-new: Yeah I definitely agree this is a better idea but felt a bit overkill for the time being. 😅 I think there are other ways we could improve the UX if you try to run `src batch preview|apply` on dotcom/an unsupported environment beyond just failing earlier, so I'll log this under an umbrella issue for those other ideas I also had. 🙂

internal/batches/service/service.go
courier-new: Gotcha gotcha, that makes sense. The concern is just that our biggest customers might *also* hit timeout issues like the user reported in the original issue if their query actually approaches a results count of 999999, and we can't easily detect that and tell the user when it's happening, so it ends up just looking like flakiness instead. You're probably right that dotcom is still going to top the instances in terms of scale, though.

Related to this, though, how did we settle on the initial number of 999k as an emulation of `count:all`? How do we know 999k is "sufficiently large", 9M is too big, and 99999 is actually not large enough? Does search cap out at a certain count?

internal/batches/service/service.go
courier-new: https://github.com/sourcegraph/src-cli/issues/690

courier-new:
Omg I'm so out of practice with CHANGELOG because we never log anything for SSBC since it's all experimental. 😂 Thanks for the reminder, Adam.

internal/batches/service/service.go
eseliger: I think we arbitrarily hit the 9 key until it felt ludicrously large to be honest. That was when sourcegraph dotcom was on the order of 10k repos indexed. I've wanted to adopt the count:all flag instead forever but it isn't so long until we don't need this anymore at all because SSBC does use count:all so .. I guess not worth breaking our heads over this for too long 😅

courier-new: Haha okay fair! I'll leave it unchanged for now, then, since we haven't heard complaints about this from enterprise customers yet and it sounds like once we use `count:all` this would also be a non-issue. 🙂 Thanks Erik!

PR #691, opened 2022-02-04
Transfer less unused info for exec command

eseliger:


cmd/src/batch_common.go
```diff
@@ -342,7 +343,13 @@ func executeBatchSpec(ctx context.Context, ui ui.ExecUI, opts executeBatchSpecOp
 	})
 
 	ui.CheckingCache()
-	tasks := svc.BuildTasks(ctx, batchSpec, workspaces)
+	tasks := svc.BuildTasks(
+		ctx, &template.BatchChangeAttributes{
```
mrnugget, line 347: That's some funky indentation.

internal/batches/executor/coordinator.go
```diff
@@ -182,8 +182,8 @@ func (c Coordinator) buildChangesetSpecs(task *Task, result execution.Result) ([
 			BaseRev:     task.Repository.Rev(),
 		},
 		BatchChangeAttributes: task.BatchChangeAttributes,
```
mrnugget, line 184: You could now also get rid of this field, right?

internal/batches/executor/coordinator.go
```diff
@@ -254,37 +260,55 @@ func (c *Coordinator) cacheAndBuildSpec(ctx context.Context, taskResult taskResu
...
+func (c *Coordinator) doExecute(ctx context.Context, tasks []*Task, ui TaskExecutionUI) (results []taskResult, err error) {
+	var errs *multierror.Error
+
+	ui.Start(tasks)
+
+	// Run executor
+	c.exec.Start(ctx, tasks, ui)
+	results, err = c.exec.Wait(ctx)
+	if err != nil {
+		if c.opts.SkipErrors {
```
mrnugget, line 306: Not sure if it's new behaviour but I think errors should be reported in the UI in any case, right?

mrnugget (approved):

PR #693, opened 2022-02-08
asdf: Add some more commonly used tools

eseliger:
Found this useful while configuring my editor on the new laptop.


courier-new (approved):
Great idea. 🙂 Do you have shfmt and shellcheck integrated with your editor?

eseliger:
Yes, both extensions are in the recommended set of our main repo :)

PR #694, opened 2022-02-11
go.mod: use grafana/regexp, matching sourcegraph/sourcegraph#30948

LawnGnome:
I did a quick `src search`, and everything still looks (a) normal, and (b) garishly coloured, so I think we're good.


eseliger (approved):
😎 are you saying you are also migrating to our errors package????? (no, you don't need to 😆)

PR #695, opened 2022-02-11
batches: fix `src batch apply -f`

LawnGnome:
This code should have been the same in `src batch preview` and `src batch apply`, but apparently wasn't. My bad, sorry.


courier-new (approved):
Makes sense, thanks so much again for the quick turnaround. 🙏

PR #697, opened 2022-02-11
go.mod: run tidy

LawnGnome:
This caused the release to fail. (Side note: why on earth doesn't Go do this for you when you run `go get` as suggested?)


eseliger (approved):
OH NO, I ran into the same thing just this week 🤦

PR #698, opened 2022-02-18
all: migrate to github.com/sourcegraph/sourcegraph/lib/errors

bobheadxi:
Migrates all errors in `src-cli` to use the new error types introduced in https://github.com/sourcegraph/sourcegraph/pull/31466, and bans further imports of other error libraries.

Some benefits of this migration is that wrapped multierrors no longer obscure child errors, assertions work more consistently, and there is now an overall more ergonomic API for multierrors - see https://github.com/sourcegraph/sourcegraph/pull/31466 for more details.


eseliger (approved):
TYSM!

camdencheek (approved):

PR #705, opened 2022-03-02
batches: retry changeset spec upload by default

LawnGnome:
Fixes https://github.com/sourcegraph/sourcegraph/issues/30431.

This only adds retries to changeset spec upload — although the original ticket also mentions batch spec upload, I'm honestly less concerned about that. Let's deal with the _n_ upload case before the _+1_ upload case.

There's a little bit of plumbing here to handle the retries, which I think is mostly uninteresting.

### Test plan

I've added new unit tests to cover the retrier, `CreateChangesetSpec`, and tested this quickly by hand to ensure the overall commands still operate.

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -75,6 +75,7 @@ type batchExecuteFlags struct {
 	file          string
 	keepLogs      bool
 	parallelism   int
+	retries       int
 	timeout       time.Duration
 	workspace     string
 	cleanArchives bool
@@ -121,6 +122,10 @@ func newBatchExecuteFlags(flagSet *flag.FlagSet, workspaceExecution bool, cacheD
 	flagSet.IntVar(
 		&caf.parallelism, "j", runtime.GOMAXPROCS(0),
 		"The maximum number of parallel jobs. Default is GOMAXPROCS.",
+	)
+	flagSet.IntVar(
+		&caf.retries, "retries", 5,
+		"The maximum number of retries that will be made when sending a changeset spec.",
 	)
 	flagSet.DurationVar(
 		&caf.timeout, "timeout", 60*time.Minute,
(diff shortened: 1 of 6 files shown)


cmd/src/batch_common.go
```diff
@@ -122,6 +123,10 @@ func newBatchExecuteFlags(flagSet *flag.FlagSet, workspaceExecution bool, cacheD
 		&caf.parallelism, "j", runtime.GOMAXPROCS(0),
 		"The maximum number of parallel jobs. Default is GOMAXPROCS.",
 	)
+	flagSet.IntVar(
+		&caf.retries, "retries", 5,
+		"The maximum number of retries that will be made when sending a changeset spec.",
+	)
```
eseliger, line 129: I think 99% of users would assume this controls retries for step command executions, not a small implementation detail in one of the many subsets of a batch spec execution. 
I would probably prefer to be very explicit like `-changeset-spec-upload-retries`.

internal/retrier/retrier.go
```diff
@@ -0,0 +1,32 @@
...
+		err := f()
+		if err == nil {
+			return nil
+		}
+
+		errs = errors.Append(errs, err)
+		if len(errs.Errors()) > max {
+			return errs
+		}
+		time.Sleep(wait)
```
eseliger, line 29: Should this listen on context cancellations? Could (not tested) prevent immediate quits using SIGINTs during the upload stage

internal/batches/service/service.go
```diff
@@ -139,20 +141,31 @@ mutation CreateChangesetSpec($spec: String!) {
...
+		"spec": string(raw),
+	})
 	var result struct {
 		CreateChangesetSpec struct {
 			ID string
 		}
 	}
-	if ok, err := svc.newRequest(createChangesetSpecMutation, map[string]interface{}{
-		"spec": string(raw),
-	}).Do(ctx, &result); err != nil || !ok {
```
eseliger, line 155: that was a bug before, right? When ok is true and err is nil?

eseliger (approved):
Good improvement. I wonder if we could actually benefit from implementing some http middleware that retries based on the response code and body. 
Ie. we would definitely want to retry on 5xx errors and on network errors, but we would not actually want to retry on 413 errors and run into the same issue 5 times. 
That way, we could also get this for both the batch spec as well as the changeset spec. 

Just a thought though, I don't think this blocks merging this as-is.

chrispine:
This has been sitting here for a while; what's the plan for this? Does it still make sense to merge it?

PR #706, opened 2022-03-02
batches: fix workspace detection test on M1 Macs

LawnGnome:
### Test plan

This is just a test fix.

### DIFF
--- a/internal/batches/workspace/workspace_test.go
+++ b/internal/batches/workspace/workspace_test.go
@@ -13,7 +13,7 @@ import (
 
 func TestBestWorkspaceCreator(t *testing.T) {
 	ctx := context.Background()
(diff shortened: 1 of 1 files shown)


courier-new (approved):

eseliger (commented):
Thanks! Do you remember why we want to check the specific architecture here as well?

LawnGnome:
> Thanks! Do you remember why we want to check the specific architecture here as well?

It needs to match the actual behaviour in code, and right now we have specific checks for the architecture on macOS.

Of course, macOS only supports the two architectures in there, so it's arguable if this check is serving much purpose right now. Just getting ahead of the macOS port to RISC-V in 2030. 😆

eseliger:
Oh, I forgot that we also check for that. Makes total sense to me then :)

PR #709, opened 2022-03-04
Cache step results while steps are executing instead of caching all of them at the end

BolajiOlajide:
### Test plan

We adjusted the test suite to ensure that tasks are cached after execution of all steps, but the steps are cached immediately the individual execution is done.

We also tested this change on the UI by taking the following steps:

- Create a batch spec that'll fail at a particular step

```yaml
name: testerr
description: Add Hello World to READMEs

on:
  - repositoriesMatchingQuery: file:README.md count:100

# In each repository, run this command. Each repository's resulting diff is captured.
steps:
  - run: echo Hello World 123 | tee -a $(find -name README.md)
    container: ubuntu:18.04
  - run: exit 1
    container: ubuntu:18.04

changesetTemplate:
  title: Hello World
  body: My first batch change!
  commit:
```

- Ensure it fails, then retry the batch spec run and checking the previous passing steps was cached.

Closes [#27564](https://github.com/sourcegraph/sourcegraph/issues/27564)

### DIFF
--- a/internal/batches/executor/coordinator.go
+++ b/internal/batches/executor/coordinator.go
@@ -63,6 +63,8 @@ type NewCoordinatorOpts struct {
 func NewCoordinator(opts NewCoordinatorOpts) *Coordinator {
 	logManager := log.NewManager(opts.TempDir, opts.KeepLogs)
 
+	globalEnv := os.Environ()
+
 	exec := newExecutor(newExecutorOpts{
 		RepoArchiveRegistry: opts.RepoArchiveRegistry,
 		EnsureImage:         opts.EnsureImage,
@@ -73,7 +75,8 @@ func NewCoordinator(opts NewCoordinatorOpts) *Coordinator {
 		Timeout:     opts.Timeout,
 		TempDir:     opts.TempDir,
 		WriteStepCacheResult: func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error {
-			return writeToCache(ctx, opts.Cache, stepResult, task)
+			cacheKey := task.cacheKey(globalEnv)
+			return writeToCache(ctx, opts.Cache, stepResult, task, cacheKey)
 		},
 	})
 
@@ -224,10 +227,7 @@ func (c *Coordinator) loadCachedStepResults(ctx context.Context, task *Task, glo
 	return nil
 }
 
-func writeToCache(ctx context.Context, cache cache.Cache, stepResult execution.AfterStepResult, task *Task) error {
-	globalEnv := os.Environ()
-	cacheKey := task.cacheKey(globalEnv)
-
+func writeToCache(ctx context.Context, cache cache.Cache, stepResult execution.AfterStepResult, task *Task, cacheKey *cache.ExecutionKeyWithGlobalEnv) error {
 	key := cacheKeyForStep(cacheKey, stepResult.StepIndex)
 	if err := cache.SetStepResult(ctx, key, stepResult); err != nil {
 		return errors.Wrapf(err, "caching result for step %d in %q", stepResult.StepIndex, task.Repository.Name)
--- a/internal/batches/executor/coordinator.go
+++ b/internal/batches/executor/coordinator.go
@@ -236,7 +236,7 @@ func writeToCache(ctx context.Context, cache cache.Cache, stepResult execution.A
 	return nil
 }
 
-func (c *Coordinator) writeCache(ctx context.Context, taskResult taskResult, ui TaskExecutionUI) error {
+func (c *Coordinator) writeExecutionCacheResult(ctx context.Context, taskResult taskResult, ui TaskExecutionUI) error {
 	// Add to the cache, even if no diff was produced.
 	globalEnv := os.Environ()
 	cacheKey := taskResult.task.cacheKey(globalEnv)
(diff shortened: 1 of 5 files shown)


internal/batches/executor/coordinator.go
```diff
@@ -221,6 +224,18 @@ func (c *Coordinator) loadCachedStepResults(ctx context.Context, task *Task, glo
 	return nil
 }
 
+func writeToCache(ctx context.Context, cache cache.Cache, stepResult execution.AfterStepResult, task *Task) error {
+	globalEnv := os.Environ()
+	cacheKey := task.cacheKey(globalEnv)
```
LawnGnome, line 229: I'm not particularly concerned about this, but if we wanted to avoid the `os.Environ()` on each task, we could grab it in `NewCoordinator` and calculate the cache key in the `WriteToCache` callback before invoking this function.

Not worth changing unless we're back in here — `os.Environ` isn't free, but it's probably not expensive enough (one read lock and _n_ string copies) to prematurely optimise. Just something to keep in mind for the future.

internal/batches/executor/coordinator_test.go
```diff
@@ -399,23 +399,21 @@ func TestCoordinator_Execute_StepCaching(t *testing.T) {
 	task.Steps[1].Run = `echo "two modified"`
 	// Re-execution should start with the diff produced by steps[0] as the
 	// start state from which steps[1] is then re-executed.
-	execAndEnsure(t, coord, executor, batchSpec, task, assertCachedResultForStep(t, 0))
+	execAndEnsure(t, coord, executor, batchSpec, task, assertNoCachedResult(t))
```
LawnGnome, line 402: Do we need to update the comments in this test? I don't think they still line up with what's happening here.

LawnGnome (approved):
Generally LGTM — nice work! I have a question about a modified test, and a vague note about having _n_ `os.Environ()` calls, but nothing blocking.

internal/batches/executor/coordinator.go
eseliger: I think that's a good call actually. Doesn't seem super hard to do, @BolajiOlajide wdyt about that?

internal/batches/executor/coordinator.go
```diff
@@ -221,6 +224,18 @@ func (c *Coordinator) loadCachedStepResults(ctx context.Context, task *Task, glo
 	return nil
 }
 
+func writeToCache(ctx context.Context, cache cache.Cache, stepResult execution.AfterStepResult, task *Task) error {
```
eseliger, line 227: We now have a writeToCache and a (*Coordinator).writeCache method, can we rename both to make clearer what each are?

Maybe `writeToCache` => `writeStepCacheResult`, `(*Coordinator).writeCache` => `(*Coordinator).writeExecutionCacheResult`?

internal/batches/executor/coordinator.go
```diff
@@ -72,6 +72,9 @@ func NewCoordinator(opts NewCoordinatorOpts) *Coordinator {
 		Parallelism: opts.Parallelism,
 		Timeout:     opts.Timeout,
 		TempDir:     opts.TempDir,
+		WriteToCache: func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error {
```
eseliger, line 75:
```suggestion
		WriteStepCacheResult: func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error {
```

to align with the comment below?

internal/batches/executor/executor.go
```diff
@@ -59,9 +59,10 @@ type newExecutorOpts struct {
 	Logger              log.LogManager
 
 	// Config
-	Parallelism int
-	Timeout     time.Duration
-	TempDir     string
+	Parallelism  int
+	Timeout      time.Duration
+	TempDir      string
+	WriteToCache func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error
```
eseliger, line 65:
```suggestion
	WriteStepCacheResult func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error
```

eseliger (approved):
Great PR and an amazing FIRST PR! Congrats on your first, very meaningful even, contribution! 🚀

internal/batches/executor/coordinator.go
BolajiOlajide: Hey! Just catching up with this.
Yes, it's non-trivial and a non-breaking change so I can do it. It's awesome because it also reduces repetition in all tasks.

I'll update accordingly.

internal/batches/executor/coordinator_test.go
BolajiOlajide: Yes, I'll update this. Thanks

internal/batches/executor/coordinator.go
BolajiOlajide: Sounds good. I'll commit this suggestion.

internal/batches/executor/coordinator.go
BolajiOlajide: Done.

internal/batches/executor/run_steps.go
```diff
@@ -37,6 +37,8 @@ type executionOpts struct {
 	logger log.TaskLogger
 
 	ui StepsExecutionUI
+
+	writeToCache func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error
```
courier-new, line 41:
```suggestion
	writeStepCacheResult func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error
```

courier-new (approved):
Looks great, thanks for added tests!

Made one more suggestion in line with the same renaming Erik suggested about `writeToCache` vs. `writeStepCacheResult`, but otherwise I'm excited for this improvement with server-side execution!

internal/batches/executor/run_steps.go
BolajiOlajide: Sounds good to me.

LawnGnome (approved):
LGTM. Thanks for the changes!

eseliger (approved):

PR #713, opened 2022-03-17
add login into command list & alphabetize 

northyg:
Fixes https://github.com/sourcegraph/sourcegraph/issues/32450 and adds a bit more!

### Test plan

* Added `login` command to the list
* Noticed list was not alphabetized so went a step further and re-ordered the commands
* Confirmed changes in src-cli dev environment look visually correct (use spaces versus tabs 🙃 )

image

### DIFF
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -32,18 +32,18 @@ The options are:
 
 The commands are:
 
-	search          search for results on Sourcegraph
 	api             interacts with the Sourcegraph GraphQL API
-	repos,repo      manages repositories
-	users,user      manages users
-	orgs,org        manages organizations
-	config          manages global, org, and user settings
-	extsvc          manages external services
-	extensions,ext  manages extensions (experimental)
 	batch           manages batch changes
+	config          manages global, org, and user settings
+	extensions,ext  manages extensions (experimental)
+	extsvc          manages external services
+	login           authenticate to a Sourcegraph instance with your user credentials
 	lsif            manages LSIF data
-	login		authenticate to a Sourcegraph instance with your user credentials
+	orgs,org        manages organizations
+	repos,repo      manages repositories
+	search          search for results on Sourcegraph
 	serve-git       serves your local git repositories over HTTP for Sourcegraph to pull
+	users,user      manages users
 	version         display and compare the src-cli version against the recommended version for your instance
 
 Use "src [command] -h" for more information about a command.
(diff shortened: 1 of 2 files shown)


cmd/src/main.go
```diff
@@ -42,6 +42,7 @@ The commands are:
 	extensions,ext  manages extensions (experimental)
 	batch           manages batch changes
 	lsif            manages LSIF data
+	login		authenticate to a Sourcegraph instance with your user credentials
```
LawnGnome, line 45: Although this looks visually aligned on an 8 space tab width, the other lines actually use spaces rather than tabs between the command name and the description (ie between `login` and `authenticate` in this case). Would you be able to change this to match the other lines with spaces, please? (Alternatively, I'm happy to do it and push it to your branch, if you'd prefer!)

LawnGnome (changes requested):

malomarrec:
Thanks for doing this!

cmd/src/main.go
northyg: Hello @LawnGnome 👋 Thanks for having a look! Hmm, I added the line with spaces and double checked the file in VSC just now. It looks like there are spaces versus tabs, but perhaps I'm missing something? 🤔 

image

If I did, could you let me know where the issue is and I am happy to fix it :)

northyg:
Thanks @abeatrix 😄 ! I re-checked the file and replied above. 🤔 
Also, realized I forgot to update the CHANGELOG.md. Do you and @LawnGnome think this PR is worth adding to it? If so I can make that change as well.

cmd/src/main.go
abeatrix: This is what I see when I check out your branch:
image
It looks like it's still using 2 tabs instead of 11 spaces.

abeatrix:
Hey @northyg I just replied above. I'm wondering if we are seeing the unchanged file because you haven't pushed your local changes into the branch?

cmd/src/main.go
northyg: @LawnGnome  Aha! @abeatrix and I figured out the problem. I got turned around in the other branch I had made, and didn't push the changes correctly. 🐝 Helped me sort it out. I also made an addition to the Changelog. Can you have a look and see if its good to go ?

LawnGnome (approved):
Looks great, thanks!

PR #718, opened 2022-03-29
extensions: remove url from copied manifest

### LINKED ISSUE #717: src ext copy publishes extension with sourcegraph.com url
See sourcegraph/customer#784. `src ext copy` does not work for air-gapped instances since we do not remove the `url` property from the manifest, which points to the extension bundle on sourcegraph.com instead of the private registry's extension bundle (which is used when `url` is omitted). This makes it so `src ext copy` basically just pins a Cloud extension's version in a private registry instead of truly preventing requests to sourcegraph.com.

tjkandala:
Closes #717. 
- Update `updatePropertyInManifest` to remove property when value is an empty string.
- Remove `url` property from manifest before publishing to private registry.

This is a backwards-compatible change, so can it be part of a [patch release](https://github.com/sourcegraph/src-cli/blob/main/DEVELOPMENT.md#releasing)? And is the changelog entry unnecessary in this case then?

### Test plan

- Added a test case for removing properties from an extension manifest.
- Built with `go install` and tested `src ext copy` to confirm that the command still works and that `url ` is no longer included in the manifest.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- `src extensions copy` now removes the `url` property from extension manifests that points to the sourcegraph.com extension bundle, enabling use for air-gapped instances.
+
 ### Removed
 
 ## 3.38.0
--- a/cmd/src/extensions_copy.go
+++ b/cmd/src/extensions_copy.go
@@ -100,6 +100,11 @@ Copy an extension from Sourcegraph.com to your private registry.
 		if err != nil {
 			return err
 		}
+		// Remove sourcegraph.com bundle URL.
+		manifest, err = updatePropertyInManifest(manifest, "url", "")
+		if err != nil {
+			return err
+		}
 
 		response, err := http.Get(extensionResult.ExtensionRegistry.Extension.Manifest.BundleURL)
 		if err != nil {
--- a/cmd/src/extensions_publish.go
+++ b/cmd/src/extensions_publish.go
@@ -247,7 +247,11 @@ func updatePropertyInManifest(manifest []byte, property, value string) (updatedM
 	if o == nil {
 		o = map[string]interface{}{}
 	}
-	o[property] = value
+	if value == "" {
+		delete(o, property)
+	} else {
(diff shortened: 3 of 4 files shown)


LawnGnome (approved):
LGTM.

Yes, we can spin a patch release with this once merged, and yes, it should have a changelog entry (which looks fine).

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- `src extensions copy` now removes the `url` property from extension manifests that points to the sourcegraph.com extension bundle, enabling use for air-gapped instances.
```
Kelvin-Lee, line 22: Not sure if this too persnickety, or if you guys have strong opinions on how we communicate changes in the changelog, but...

Is `sourcegraph.com` always going to be the prefix of the `BundleURL` for `copy`'d extensions? Could one theoretically copy an extension from some other sourcegraph instance (e.g. cse-aws-test.sgdev.org copies from k8s.sgdev.org)? Very edge-casey, I suppose.

I guess I'd prefer language that's more like "resets the `BundleURL`" or something, if it's a possibility that extensions can be copied from not-sourcegraph.com instances.

Deferring entirely to your guys' judgement, just wanted to express my concerns.

eseliger (approved):

abeatrix (approved):
Lgtm! Great found!

CHANGELOG.md
tjkandala: Good thinking! For now, `copy` is [hardcoded](https://sourcegraph.com/github.com/sourcegraph/src-cli@tjk/ext-copy-remove-url/-/blob/cmd/src/extensions_copy.go?L23%3A65=) to use `sourcegraph.com`

Kelvin-Lee: Hey @tjkandala , the link you provided doesn't work for me anymore (I assume because the branch got deleted), but I'm assuming you intended to refer to [this LOC](https://github.com/sourcegraph/src-cli/blob/1c260bce6d91725ef2c17399cd5abdf1972be238/cmd/src/extensions_copy.go#L75)?

PR #721, opened 2022-03-31
lsif upload: Add gitlab-token

efritz:
See https://github.com/sourcegraph/sourcegraph/pull/33254. Adds `gitlab-token` to lsif upload command.

### Test plan

Tested locally as part of the related change in #33254.


efritz:
Waiting for other PR to land so we can update go.mod. Will also add a changelog entry pre-merge.

mrnugget (approved):

PR #729, opened 2022-04-14
lsif upload: Make better unauthorized error messages

### LINKED ISSUE #727: Improve hints on GitHub/GitLab access token related errors
Recently a customer carried out the following steps:

1. Created a personal access token under their Sourcegraph user settings.
2. Called `src lsif upload` with the SRC_ACCESS_TOKEN value based on that when calling src-cli.

They got an error: (as they wrote in the email)

```
❌ Failed to upload index file
unauthorized upload
You may need to specify or update your GitHub access token to use this endpoint.
```

Based on the code, it looks like they must've also got a hint pointing to https://docs.sourcegraph.com/cli/references/lsif/upload (which is great!) but it looks they weren't able to figure out a solution by reading the docs (or they didn't read the docs).

Perhaps we should:
1. If the `-github-token` argument is missing; explicitly recommend passing a `-github-token` argument.
2. If the `-github-token` argument is present; provide a hint that the token may have either expired or doesn't have the `public_repo` access scope.
 
(Similarly for GitLab.)

Context: [Slack thread 1](https://sourcegraph.slack.com/archives/C025T3B3NAH/p1649945814897659?thread_ts=1649945678.859089&cid=C025T3B3NAH), [Slack thread 2](https://sourcegraph.slack.com/archives/CHXHX7XAS/p1649948513728549?thread_ts=1649948423.598549&cid=CHXHX7XAS)

efritz:
Fixes #727.

### Test plan

Ran by hand:

```
$ src lsif upload
💡 Inferred arguments
   repo: github.com/sourcegraph/lsif-go
   commit: cde8136abd91c0a3bce9f5d2755bc006a2caaafb
   root:
   file: ./dump.lsif
   indexer: lsif-go
   indexerVersion: dev

✅ Index compressed
💡 Indexed compressed (6.65MB -> 1.06MB).
❌ Failed to upload index file
unauthorized upload

This Sourcegraph instance has enforced auth for LSIF uploads.
BLAH BLAH TEXT CHANGED SINCE I OPENED THE PR
See https://docs.sourcegraph.com/cli/references/lsif/upload.
```

### DIFF
--- a/cmd/src/lsif_upload.go
+++ b/cmd/src/lsif_upload.go
@@ -219,18 +219,35 @@ func (e errorWithHint) Error() string {
 	return fmt.Sprintf("%s\n\n%s\n", e.err, e.hint)
 }
 
-var errUnauthorizedHint = strings.Join([]string{
-	"You may need to specify or update your GitHub access token to use this endpoint.",
-	"See https://docs.sourcegraph.com/cli/references/lsif/upload.",
-}, "\n")
-
 // handleLSIFUploadError writes the given error to the given output. If the
 // given output object is nil then the error will be written to standard out.
 //
 // This method returns the error that should be passed back up to the runner.
 func handleLSIFUploadError(out *output.Output, err error) error {
 	if err == upload.ErrUnauthorized {
-		err = errorWithHint{err: err, hint: errUnauthorizedHint}
+		var actionableHints []string
+		if lsifUploadFlags.gitHubToken != "" {
+			actionableHints = append(actionableHints,
+				"The supplied -github-token did not give the expected access level for the target repository.",
+				"Please check that supplied token matches the code host and has expected access of the target repository.",
+			)
+		} else if lsifUploadFlags.gitLabToken != "" {
+			actionableHints = append(actionableHints,
+				"The supplied -gitlab-token did not give the expected access level for the target repository.",
+				"Please check that supplied token matches the code host and has expected access of the target repository.",
+			)
+		} else {
+			actionableHints = append(actionableHints,
+				"Please retry your request with a -github-token or -gitlab-token (matching the code host of the target repository).",
+				"This token will be used to check with the code host that the uploading user has write access to the target repository.",
+			)
+		}
+
+		err = errorWithHint{err: err, hint: strings.Join(mergeStringSlices(
+			[]string{"This Sourcegraph instance has enforced auth for LSIF uploads."},
+			actionableHints,
+			[]string{"See https://docs.sourcegraph.com/cli/references/lsif/upload."},
+		), "\n")}
 	}
 
 	if lsifUploadFlags.ignoreUploadFailures {
@@ -245,4 +262,13 @@ func handleLSIFUploadError(out *output.Output, err error) error {
 // emergencyOutput creates a default Output object writing to standard out.
 func emergencyOutput() *output.Output {
 	return output.NewOutput(os.Stdout, output.OutputOpts{})
+}
+
+func mergeStringSlices(ss ...[]string) []string {
+	var combined []string
+	for _, s := range ss {
+		combined = append(combined, s...)
+	}
+
+	return combined
 }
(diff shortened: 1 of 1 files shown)


cmd/src/lsif_upload.go
```diff
@@ -219,18 +219,35 @@ func (e errorWithHint) Error() string {
...
+				"Please check that supplied token matches the code host and has expected access of the target repository.",
+			)
+		} else if lsifUploadFlags.gitLabToken != "" {
+			actionableHints = append(actionableHints,
+				"The supplied -gitlab-token did not give the expected access level for the target repository.",
+				"Please check that supplied token matches the code host and has expected access of the target repository.",
+			)
+		} else {
+			actionableHints = append(actionableHints,
+				"Please retry your request with a -github-token or -gitlab-token (matching the code host of the target repository).",
```
varungandhi-src, line 241: Can we pass in the repo URL so that we can emit a precise hint here about -github-token or -gitlab-token?

cmd/src/lsif_upload.go
```diff
@@ -219,18 +219,35 @@ func (e errorWithHint) Error() string {
...
 //
 // This method returns the error that should be passed back up to the runner.
 func handleLSIFUploadError(out *output.Output, err error) error {
 	if err == upload.ErrUnauthorized {
-		err = errorWithHint{err: err, hint: errUnauthorizedHint}
+		var actionableHints []string
+		if lsifUploadFlags.gitHubToken != "" {
+			actionableHints = append(actionableHints,
+				"The supplied -github-token did not give the expected access level for the target repository.",
+				"Please check that supplied token matches the code host and has expected access of the target repository.",
```
varungandhi-src, line 232: "expected access level" seems quite vague/uninformative. 😐 Any specific reason to not repeat the exact requirements here? (I can see one reason to not duplicate the docs; that would make keeping them in sync trickier.)

cmd/src/lsif_upload.go
```diff
@@ -219,18 +219,35 @@ func (e errorWithHint) Error() string {
...
+			actionableHints = append(actionableHints,
+				"Please retry your request with a -github-token or -gitlab-token (matching the code host of the target repository).",
+				"This token will be used to check with the code host that the uploading user has write access to the target repository.",
+			)
+		}
+
+		err = errorWithHint{err: err, hint: strings.Join(mergeStringSlices(
+			[]string{"This Sourcegraph instance has enforced auth for LSIF uploads."},
+			actionableHints,
+			[]string{"See https://docs.sourcegraph.com/cli/references/lsif/upload."},
```
varungandhi-src, line 249:
```suggestion
			[]string{"For more details, see https://docs.sourcegraph.com/cli/references/lsif/upload."},
```

cmd/src/lsif_upload.go
efritz: We don't reach out to any other code host (because we can't supply any other tokens). We could do the [same check](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@a2c3de6ceb4a70c9ad9946797fd39c3fd6b2bf40/-/blob/enterprise/cmd/frontend/internal/codeintel/httpapi/auth_middleware.go?L25) we do in the frontend to figure out which code host it is, though.

cmd/src/lsif_upload.go
efritz: The scopes are different for github+gitlab. Should we just fork on each one and have explicit instructions here?

varungandhi-src: That sounds like a good approach. 👍

cmd/src/lsif_upload.go
varungandhi-src: > We don't reach out to any other code host (because we can't supply any other tokens).

Hmm, I hadn't looked at the code with too much attention; I'd assumed this same code path would apply for other code hosts too, and that we don't currently support access tokens for other code hosts.

Doing some check for a more targeted message (particularly around the access scopes part below) sounds good. 👍

efritz: We return a specific error in the non-github/non-gitlab case: https://sourcegraph.com/github.com/sourcegraph/sourcegraph@947423ea4f32fe4766d406b4ad1dff08407a1e20/-/blob/enterprise/cmd/frontend/internal/codeintel/httpapi/auth_middleware.go?L64

chrismwendt (approved):

PR #730, opened 2022-04-20
Fix Execution URL when using `src batch remote`

BolajiOlajide:
### Test plan

Run `src batch remote -f comby-go-refactor.batch.yml`

comby-go-refactor.batch.yml
```yaml
name: comby-go-fmt
description: Run `comby` and `go fmt`

on:
  - repositoriesMatchingQuery: lang:go fmt.Sprintf("%d", :[v]) patterntype:structural

steps:
  - run: comby -in-place 'fmt.Sprintf("%d", :[v])' 'strconv.Itoa(:[v])' ${{ join repository.search_result_paths " " }}
    container: comby/comby
  - run: goimports -w ${{ join previous_step.modified_files " " }}

changesetTemplate:
  title: Replacing fmt.Sprintf with strconv.Itoa
  body: This campiagn replaces `fmt.Sprintf` with `strconv.Itoa`
  branch: batch-changes/comby-go-fmt # Push the commit to this branch.
  commit:
    message: Replacing fmt.Sprintf with strconv.Iota
```

The result is an execution URL that doesn't 404.
Screenshot 2022-04-20 at 15 46 36

Closes [#33696](https://github.com/sourcegraph/sourcegraph/issues/33696)

### DIFF
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -480,12 +480,12 @@ query GetCurrentUserID {
 }
 `
 
-type NamespaceDetails struct {
+type Namespace struct {
 	ID  string
 	URL string
 }
 
-func (svc *Service) ResolveNamespace(ctx context.Context, namespace string) (NamespaceDetails, error) {
+func (svc *Service) ResolveNamespace(ctx context.Context, namespace string) (Namespace, error) {
 	if namespace == "" {
 		// if no namespace is provided, default to logged in user as namespace
 		var resp struct {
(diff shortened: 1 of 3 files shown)


internal/batches/service/service.go
```diff
@@ -128,14 +128,14 @@ func (svc *Service) CreateBatchSpec(ctx context.Context, namespace, spec string,
 
 const createChangesetSpecMutation = `
 mutation CreateChangesetSpec($spec: String!) {
```
eseliger, line 130: might want to revert to spaces, but not a blocker

courier-new (approved):
Nice! Glad it wasn't too challenging finding those other pieces we need in the URL! 🙌 😄

internal/batches/service/service.go
```diff
@@ -397,7 +397,7 @@ on:
 steps:
```
eseliger, line 397: this must be reverted to spaces, YAML doesn't like tabs.

eseliger (approved):

internal/batches/service/service.go
eseliger: (same applies for the instances of this further down)

internal/batches/service/service.go
courier-new: 😬 What auto-formatter is switching them back to tabs? Could we tell it to ignore this file in code somehow?

BolajiOlajide: Oh I missed that. I'll update now.

I guess I got confused because there was a mix of spaces and tabs so I just went with tabs for everything .

BolajiOlajide: Actually, i did that manually - there was an issue where i added a new field to the graphql query and that made use of tabs, while the rest of the query made use of spaces. I adjusted the whole file to use tabs and that's where the problem is, I'll just revert that commit so it doesn't break things.

BolajiOlajide: Reverted.

PR #733, opened 2022-04-26
batches: clarify on.repository error message when for incorrect repo path

adeola-ak:
Closes #33778

going back and fourth between this error message:
`no repository found: double-check your spelling and make sure to specify the repository in the format "codehost.com/owner/repo-name"` 
&
 `no repository found: did you check spelling and specify the repository in the format "codehost.com/owner/repo-name"? 
`

Thoughts?

Also, I believe the SSBC error for this operation may also not be clear enough and could benefit from being changed to whatever we decide above. It reads:
`Resolving "r:sourcegraph": repo not found: name="sourcegraph"
`

### Test plan
observed visual change in error text in src cli

### DIFF
--- a/internal/batches/service/service.go
+++ b/internal/batches/service/service.go
@@ -671,7 +671,9 @@ func (svc *Service) resolveRepositoryName(ctx context.Context, name string) (*gr
 		return nil, err
 	}
 	if result.Repository == nil {
-		return nil, errors.New("no repository found")
+
+		// no repository found: double-check your spelling and make sure to specify the repository in the format <>
+		return nil, errors.New("no repository found: did you check spelling and specify the repository in the format \"codehost.com/owner/repo-name\"? ")
 	}
 	return result.Repository, nil
 }


BolajiOlajide (approved):

internal/batches/service/service.go
```diff
@@ -671,7 +671,9 @@ func (svc *Service) resolveRepositoryName(ctx context.Context, name string) (*gr
 		return nil, err
 	}
 	if result.Repository == nil {
-		return nil, errors.New("no repository found")
+
+		// no repository found: double-check your spelling and make sure to specify the repository in the format <>
+		return nil, errors.New("no repository found: did you check spelling and specify the repository in the format \"codehost.com/owner/repo-name\"? ")
```
eseliger, line 676: I think this error message should include the queried repo name, otherwise, it's very hard to act on the error.

Also, what do people think about the suggested format? This is not required in Sourcegraph and some repos may actually be called just `sourcegraph`, depending on the instance settings. Would this be misleading?

adeola-ak: so this is the full error message 
```
Error:  resolving repositories: resolving "r:repo-name": no repository found
```

the repo name comes from higher up in the error change.

Also this is the original issue. In this instance that Malo was using, it seemed required. Not sure how much of the time the full path is required versus not. Maybe this needs more collaborating on the best way to word this that will take into consideration both cases and not confuse the user?

PR #735, opened 2022-05-02
Handle nil output when writing progress message

olafurpg:
Previously, the command `src lsif upload -no-progress` failed with a
nil panic when using LSIF Typed. I was tempted to refactor the code
so that `output` cannot be nil but decided not to in order to keep the
diff small and focused on fixing the panic.

### Test plan

Manually ran `src lsif upload -no-progress` and see it not panic.


Strum355 (approved):
Not the first person to hit this issue lol

PR #736, opened 2022-05-02
batches: clarify on.repository error message when for incorrect repo path

adeola-ak:
Closes https://github.com/sourcegraph/src-cli/pull/733 https://github.com/sourcegraph/sourcegraph/issues/33778

I was going back and fourth between this error message:

`no repository found: double-check your spelling and make sure to specify the repository in the format "<codehost_url>/owner/repo-name"`
&
`no repository found: did you check spelling and specify the repository in the format "<codehost_url>/owner/repo-name"?
`

Thoughts?

Also, I believe the SSBC error for this operation may also not be clear enough and could benefit from being changed to whatever we decide above. It reads:
Resolving "r:sourcegraph": repo not found: name="sourcegraph"

note: duplicate of this closed [issue](https://github.com/sourcegraph/src-cli/pull/733) after i realizing i was working in forked version

### Test plan
observed visual change in error text in src cli


eseliger:
For other people stopping by this PR: 
The complete error from src-cli:
CleanShot 2022-05-03 at 00 55 03@2x

I personally think that the main problem here is the `r:the-repo/thatdoesntexist` which doesn't refer to anything in the spec, and not the `resolving:` prefix. I would recommend we should change it to `repository: the-repo/thatdoesntexist`. 

Let me start a slack thread on this, so we can get it resolved quickly. :)

eseliger:
For other people stopping by this PR: 
The complete error from src-cli:
CleanShot 2022-05-03 at 00 55 03@2x

I personally think that the main problem here is the `r:the-repo/thatdoesntexist` which doesn't refer to anything in the spec, and not the `resolving:` prefix. I would recommend we should change it to `repository: the-repo/thatdoesntexist`. 

Let me start a slack thread on this, so we can get it resolved quickly. :)

adeola-ak:
hmm thats odd. I thought I did change it to say repository. This is what i see on my end:

Screen Shot 2022-05-02 at 9 28 16 PM

BolajiOlajide:
> hmm thats odd. I thought I did change it to say repository. This is what i see on my end:
> 
> 
> 
> Screen Shot 2022-05-02 at 9 28 16 PM
> 
> 
> 
> 

@eseliger can you confirm you built from her most recent commit because the message you get is weird.

courier-new:
I'm seeing the error in the same format as Eric, built on [f284388](https://github.com/sourcegraph/src-cli/pull/736/commits/f284388193b42a7841381f0e22d944e118506430). I believe the "r:sourcegraph" part is what `on.String()` produces [here](https://sourcegraph.com/github.com/sourcegraph/src-cli@aa/onrepo-error/-/blob/internal/batches/service/service.go?L557&utm_source=VSCode-2.2.1).

eseliger:
Yup, that's the place it gets rendered. However, what do people think in general? Would changing it from `r:` to `repository: ` improve the actionability here?

courier-new:
Yeah I like that! Assuming it doesn't break usage of that string in other places, I guess.

adeola-ak:
>You can trace the implementation to `OnQueryOrRepository`'s [`String()` method](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/lib/batches/batch_spec.go?L203%3A3-203%3A30=).

@courier-new So this part that you mentioned is actually the part I changed from r to repository. Its not showing up in my changed files accounted in the PR though. 🧐

adeola-ak:
@eseliger @courier-new any suggestions or ideas about why on my end it shows `repository` instead of `r` but doesnt reflect on your both end? or why the file isnt showing as updated in my PR even though I did update it? Does that file not get pushed to git?
Screen Shot 2022-05-03 at 8 52 27 PM

courier-new:
Ooh the classic "oops wrong repo!!" strike again 😅  Looks like that interface and function are from a package that lives in sourcegraph/sourcegraph, not sourcegraph/src-cli, so you'll have to open a corresponding PR over there!

adeola-ak:
@courier-new 😯 but how, if i was only working out of src-cli repo ?

courier-new:
We publish the package that `OnQueryOrRepository` is defined in from sourcegraph/sourcegraph on the Go package registry, so that we can import and use it in src-cli!

So if you found that function using VSCode's "Go to definition", it probably actually opened the installed package file. What's the file path from that screenshot you shared above? Did it look something like `.../.go/pkg/mod/github.com/sourcegraph/sourcegraph/lib@<someversion>/batches/batch_spec.go`? This is basically the Go equivalent of `node_modules` -- in other words, while making a change there might work on your machine, we actually want to make the change on the _source file_ for that package, so that we can publish an update to that package to get picked up by src-cli.

So I don't think the order of the PRs matters much, just as long as they both get merged before we release a new version of src-cli.

eseliger:
> So I don't think the order of the PRs matters much, just as long as they both get merged before we release a new version of src-cli.

The go lib package in sourcegraph/sourcegraph needs to be merged first, and then in this PR you need to bump the version of that package. Or at least in _a_ PR, because otherwise it won't be picked up on main. go.sum is a lockfile essentially. :) 
For that you can use `go get -u github.com/sourcegraph/sourcegraph/lib@<THE_LANDED_COMMIT_ON_MAIN_HASH>` in this repo.

courier-new:
Ah, duh, thanks for the reminder, Erik! 🙇‍♀️

courier-new (approved):

PR #742, opened 2022-05-13
Add support for SCIP + renaming.

varungandhi-src:
We use the name `index.scip` as the default. I've also renamed
a bunch of stuff which was referring to LSIF Typed.

Marking as draft for now since the code won't compile until
the SCIP repo is made public, and I also need to add tests.

### Test plan

Updated existing tests + added new tests.


cmd/src/lsif_upload_flags.go
```diff
@@ -155,35 +156,43 @@ type argumentInferenceError struct {
 	err      error
 }
 
-func handleLSIFTyped(out *output.Output) error {
-	if strings.HasSuffix(lsifUploadFlags.file, ".lsif-typed") {
-		// The user explicitly passed in a -file flag that points to an LSIF Typed index.
+func replaceExtension(oldPath string, newExtension string) string {
+	return oldPath[:len(oldPath)-len(path.Ext(oldPath))] + newExtension
```
efritz, line 160:
```suggestion
	return filepath.Join(filepath.Dir(oldPath), filepath.Base(oldPath) + newExtension)
```

efritz (approved):

cmd/src/lsif_upload_flags.go
varungandhi-src: I don't think that is correct; it will only attach the extension without removing the old extension.

efritz: Oh no I miscomprehended the entire stdlib 😓

PR #731, opened 2022-04-22
src debug command

DaedalusG:
### Src Debug

`src debug` is a little pet project I've been working on for a long time now as a way to learn golang. The intention here is to create a cli command that collects the data from common diagnostic commands to speed up debugging and make sharing information about a Sourcegraph instance easier for our support team.

The command is divided into three separate sub commands (`server`,`compose`, and `kube`) each specific to their deployment type and with some unique flags and values.

Here is an example of the return from a `src debug server` command:
```
warrengifford@Warrens-MacBook-Pro src-debug-test % src -v debug server -c vibrant_ritchie -o serv3
This command will archive docker-cli data for container: vibrant_ritchie
 SRC_ENDPOINT: https://cse-aws-test.sgdev.org
 Output filename: serv3.zip
Do you want to start writing to an archive? [y/N]: y
archiving file "serv3/inspect-vibrant_ritchie.txt" with 12294 bytes
archiving file "serv3/vibrant_ritchie.log" with 571587 bytes
archiving file "serv3/top-vibrant_ritchie.txt" with 84951 bytes
archiving file "serv3/config/siteConfig.json" with 3772 bytes
archiving file "serv3/config/external_services.txt" with 9480 bytes
```
_This is the simplest version of the command and output_

For an abridged example of kubernetes --
```
warrengifford@Warrens-MacBook-Pro src-debug-test % src debug kube -n ns-sourcegraph -o kube-test
Archiving kubectl data for 33 pods
 SRC_ENDPOINT: https://cse-aws-test.sgdev.org
 Context: gke_beatrix-test-overlay_us-central1-c_beatrix-test
 Namespace: ns-sourcegraph
 Output filename: kube-test.zip
Do you want to start writing to an archive? [y/N]: y
```

... after unzip

```
warrengifford@Warrens-MacBook-Pro src-debug-test % cd kube-test
warrengifford@Warrens-MacBook-Pro kube-test % ls
config  kubectl
warrengifford@Warrens-MacBook-Pro kube-test % ls config
external_services.txt   siteConfig.json
warrengifford@Warrens-MacBook-Pro kube-test % ls kubectl
events.txt                      persistent-volume-claims.txt    pods
getPods.txt                     persistent-volumes.txt
warrengifford@Warrens-MacBook-Pro kube-test % ls kubectl/pods
cadvisor-44nvf                                  jaeger-56fbb74f8-jmtjs
cadvisor-55b9d                                  minio-54b558cdd6-hlgbt
[...]

PR #743, opened 2022-05-13
chore: Follow up to #731

efritz:
Following up from additional comments in #731 (some were hidden by the long timeline in the github view pre merge, and @DaedalusG and I forgot to address these in our 1-1).

### Test plan

N/A.

### DIFF
--- a/cmd/src/debug_compose.go
+++ b/cmd/src/debug_compose.go
@@ -42,7 +42,7 @@ Examples:
 	var base string
 	var excludeConfigs bool
 	flagSet.StringVar(&base, "o", "debug.zip", "The name of the output zip archive")
-	flagSet.BoolVar(&excludeConfigs, "no-configs", false, "If true, include Sourcegraph configuration files. Defaults to false.")
+	flagSet.BoolVar(&excludeConfigs, "no-configs", false, "If true, exclude Sourcegraph configuration files. Defaults to false.")
 
 	handler := func(args []string) error {
 		if err := flagSet.Parse(args); err != nil {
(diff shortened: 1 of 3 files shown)


cmd/src/debug_compose.go
```diff
@@ -114,7 +113,7 @@ func archiveCompose(ctx context.Context, zw *zip.Writer, verbose, noConfigs bool
 	// setup channel for slice of archive function outputs
 	ch := make(chan *archiveFile)
 	g, ctx := errgroup.WithContext(ctx)
-	semaphore := semaphore.NewWeighted(8)
+	semaphore := semaphore.NewWeighted(int64(runtime.GOMAXPROCS(0)))
```
DaedalusG, line 116: This one is interesting, I'm not familiar with `runtime.GOMAXPROCS(0)`, will definitely do some research here 👍

efritz: It gives you the recommended number of processes/goroutines that can run in parallel (not just concurrently). It's relational to your cpu die count.

DaedalusG (approved):
This looks good and mostly seems like clarification of naming conventions. Except `semaphore := semaphore.NewWeighted(int64(runtime.GOMAXPROCS(0)))` which I'll need to learn more about. 

Has the command been tested with these changes? If not I'm happy to test these out before merge. 

Approving anyway

cmd/src/debug_compose.go
DaedalusG: Hmm, will this potentially reintroduce issues with open file limits? Or reduce the effectiveness of the semaphore in throttling remotely executed requests? In my understanding, this will scale the semaphore's limit in terms of the host machine running `src`'s resources?

efritz: Number of cpu cores is way under the Linux file handle limit by many orders of magnitude

LawnGnome (approved):

PR #748, opened 2022-05-18
Add src code-intel alias for src lsif

varungandhi-src:
Replaces #745. After discussing with Olaf and TJ, it was decided that
adding other commands would be easier if we had the in-between
command. We will add another alias depending on what Code Intel
is rebranded to. 😐

### Test Plan

Updated existing tests. Ran `src -h`, `src code-intel -h` and `src code-intel upload -h` manually.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ All notable changes to `src-cli` are documented in this file.
 ## Unreleased
 
 ### Added
+- `src code-intel` is a new alias for `src lsif`. [#745](https://github.com/sourcegraph/src-cli/pull/745)
 
 ### Changed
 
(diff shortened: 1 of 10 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ All notable changes to `src-cli` are documented in this file.
 ## Unreleased
 
 ### Added
+- `src code-intel` is a new alias for `src lsif`. [#745](https://github.com/sourcegraph/src-cli/pull/745)
```
efritz, line 15: <3

efritz (approved):

CHANGELOG.md
varungandhi-src: Oops, that is linking the wrong PR number.

PR #751, opened 2022-05-25
Bump SCIP version to include bugfix for textDocument/implementations

olafurpg:
The PR we want to include https://github.com/sourcegraph/scip/pull/27

### Test plan

See the CI go green. We already tested the upstream bugfix.


tjdevries (approved):
LGTM assuming green

PR #760, opened 2022-06-02
Make src-cli installable via npm.

### LINKED ISSUE #759: Make src-cli installable via npm
The Code Intel team is soon announcing two precise language indexers (TypeScript and Python) that will be installed via npm. To simplify CI installation, it would be nice if users of those indexers can also install src-cli via npm instead of via `curl` commands. Installing src-cli via npm also makes it easier for users to update to the latest src-cli version.

- https://www.npmjs.com/package/tree-sitter-cli Rust binary that's downloaded as a post-install hook. The benefit of this approach is that users only install one binary that's compatible with their OS/CPU.
- https://www.npmjs.com/package/@bazel/bazelisk Go binary that's embedded with the npm package. The downside of this approach is that users download multiple binaries for all supported OS/CPU.

I propose we use the same approach as tree-sitter-cli.

Related Slack conversation https://sourcegraph.slack.com/archives/C07KZF47K/p1653896066944809

olafurpg:
Fixes #759. Previously, the recommended way to install src-cli on Linux
was to download a static binary via `curl`. This PR makes it possible to
install src-cli via `npm install -g @sourcegraph/src`. The new npm
distribution will be particuarly helpful for users of our new
scip-typescript and scip-python indexers that are also distributed via
npm.

### Test plan

Manually ran `yarn install` in the `npm-distribution` directory to verify that it works as expected. There's no reliable way to test this change without trying to run the actual release step to npmjs.com.


npm-distribution/package.json
```diff
@@ -0,0 +1,26 @@
+{
+  "name": "@sourcegraph/src",
+  "version": "3.40.1",
```
eseliger, line 3: Aren’t we at 3.40.3 after this lands?

DEVELOPMENT.md
```diff
@@ -58,11 +58,13 @@ We adhere to the [general Sourcegraph principles for testing](https://docs.sourc
 
 1.  If this is a non-patch release, update the changelog. Add a new section `## $MAJOR.MINOR` to [`CHANGELOG.md`](https://github.com/sourcegraph/src-cli/blob/main/CHANGELOG.md#unreleased) immediately under `## Unreleased changes`. Add new empty `Added`, `Changed`, `Fixed`, and `Removed` sections under `## Unreleased changes`.
 2.  Find the latest version (either via the releases tab on GitHub or via git tags) to determine which version you are releasing.
-3.  `VERSION=9.9.9 ./release.sh` (replace `9.9.9` with the version you are releasing)
-4.  GitHub will automatically perform the release via the [goreleaser action](https://github.com/sourcegraph/src-cli/actions?query=workflow%3AGoreleaser). Once it has finished, **you need to confirm**:
+3.  Update the `"version"` field in the file `npm-distribution/package.json` to match the version you are releasing.
```
eseliger, line 61: Can the release script automate that?

.github/workflows/goreleaser.yml
```diff
@@ -33,4 +33,14 @@ jobs:
         env:
           # Use separate access token, because we need a scope:repo token to publish the brew formula.
           GITHUB_TOKEN: ${{ secrets.GH_PAT }}
- 
+  npm:
+    runs-on: ubuntu-latest
```
eseliger, line 37: Should this depend on the goreleaser step? Otherwise there will be a short moment in which it doesn’t work.

eseliger (approved):

DEVELOPMENT.md
olafurpg: Is it OK if the release step runs `git push origin main`? I already updated the release script to validate that the version in package.json matches but didn't feel comfortable pushing a commit to main without review/test-plan.

olafurpg: Given that we need to merge the changelog entries anwyays, i think it's OK to require a PR before running `VERSION=X ./release.sh`

olafurpg: I'm happy to follow up on this after merge. Gonna try to cut a new release to see it go live on npm

npm-distribution/package.json
olafurpg: Good catch! Fixed.

.github/workflows/goreleaser.yml
olafurpg: Good point. I updated the flow to depend on the goreleaser job.

olafurpg:
🍿  https://github.com/sourcegraph/src-cli/actions/runs/2427735253

CleanShot 2022-06-02 at 12 53 25

eseliger:
Woho, seems to have worked!

Re the additional manual step:
How do you feel about using a similar approach that the semanticrelease bot uses?

It has `0.0.0-dev` checked in as the version and replaces it inline when the release process runs.

olafurpg:
@eseliger That's a nice idea. I opened #763 implementing your suggestion.

PR #762, opened 2022-06-02
Use `setup-node` GitHub Actions step before running `npm publish`.

olafurpg:
The last `npm publish` step failed with an authorization error
https://github.com/sourcegraph/src-cli/runs/6707002077?check_suite_focus=true
```
npm ERR! need auth This command requires you to be logged in to https://registry.npmjs.org/
29
```

Our current hypothesis is that this change fixes the issue.


varungandhi-src (approved):

olafurpg:
Next attempt https://github.com/sourcegraph/src-cli/actions/runs/2427826231

PR #763, opened 2022-06-02
Simplify npm package release by setting the version

olafurpg:
Previously, you had to manually open a PR to bump up the version in package.json.
Now, the version is automatically bumped during the release pipeline.

### Test plan

Cut a new release and see if it works.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -5,21 +5,17 @@ set -euf -o pipefail
 read -p 'Have you read DEVELOPMENT.md? [y/N] ' -n 1 -r
 echo
 case "$REPLY" in
-    Y|y) ;;
-    *)
-        echo 'Please read the Releasing section of DEVELOPMENT.md before running this script.'
-        exit 1
-        ;;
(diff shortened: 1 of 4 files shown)


release.sh
```diff
@@ -5,22 +5,18 @@ set -euf -o pipefail
 read -p 'Have you read DEVELOPMENT.md? [y/N] ' -n 1 -r
 echo
 case "$REPLY" in
-    Y|y) ;;
```
olafurpg, line 8: I don't know why this is still getting formatted on file save when I've disabled format-on-save for all languages but go 😢

eseliger: 🤷‍♂️, honestly 😄

eseliger (approved):
Thanks for the quick improvement!

olafurpg:
Cut a new patch release 3.40.5

https://github.com/sourcegraph/src-cli/actions/runs/2432861873

PR #767, opened 2022-06-03
go.mod: update to latest sourcegraph/sourcegraph/lib

LawnGnome:
This includes a fixup for an import that has moved from sourcegraph/sourcegraph/lib to sourcegraph/scip since the last time we updated the dependency.

### Test plan

Tested Batch Changes functionality by hand, and the test suite still passes.


eseliger (approved):

Piszmog (approved):

varungandhi-src:
Thanks for catching this; I forgot to update this earlier while moving the code.

PR #768, opened 2022-06-06
Polish and use native workspaces in executors mode

eseliger:
This PR cleans up the src side of things for SSBC:
- Ensures to have a separate cache directory for each src running, so they don't collide while writing to them
- Ensures to have a tmp directory set within the executor workspace so it can be retained for inspection
- Removes useless API call to Sourcegraph to retrieve it's version
- Supports native repos cloned into the workspace
- Ensures no env var is used for cache key generation, those are not set on the server and it would cause cache key mismatches
- And most importantly, it doesn't require an access token anymore, so it's much safer, especially in less-isolated executors!

Closes https://github.com/sourcegraph/sourcegraph/issues/37079
Closes https://github.com/sourcegraph/sourcegraph/issues/36597
Closes https://github.com/sourcegraph/sourcegraph/issues/36187

### Test plan

Lots of manual validation. I will also shove it into k8s post-merge and see if it behaves well.


cmd/src/batch_common.go
```diff
@@ -82,6 +82,8 @@ type batchExecuteFlags struct {
 
 	// EXPERIMENTAL
 	textOnly bool
+
+	sourcegraphVersion string
```
eseliger, line 86: we pass this down so we can do feature detection without needing to talk to Sourcegraph

cmd/src/batch_exec.go
```diff
@@ -53,7 +56,7 @@ Examples:
 
 		err := executeBatchSpecInWorkspaces(ctx, &ui.JSONLines{}, executeBatchSpecOpts{
 			flags:  flags,
-			client: cfg.apiClient(flags.api, flagSet.Output()),
+			client: &deadClient{},
```
eseliger, line 59: There should never be requests made to Sourcegraph, so we guard against it.

cmd/src/batch_exec.go
```diff
@@ -131,18 +149,25 @@ func executeBatchSpecInWorkspaces(ctx context.Context, ui *ui.JSONLines, opts ex
...
 
-	// EXECUTION OF TASKS
+	// EXECUTION OF TASK
 	coord := svc.NewCoordinator(executor.NewCoordinatorOpts{
-		Creator:       workspaceCreator,
-		CacheDir:      opts.flags.cacheDir,
-		Cache:         &executor.ServerSideCache{Writer: ui},
-		SkipErrors:    opts.flags.skipErrors,
-		CleanArchives: opts.flags.cleanArchives,
+		Creator: workspaceCreator,
```
eseliger, line 154: lots of TODOs here to figure out.

internal/batches/executor/run_steps.go
```diff
@@ -42,13 +42,13 @@ type executionOpts struct {
...
 
 func runSteps(ctx context.Context, opts *executionOpts) (result execution.Result, stepResults []execution.AfterStepResult, err error) {
-	opts.ui.ArchiveDownloadStarted()
-	err = opts.task.Archive.Ensure(ctx)
-	opts.ui.ArchiveDownloadFinished(err)
-	if err != nil {
-		return execution.Result{}, nil, errors.Wrap(err, "fetching repo")
-	}
-	defer opts.task.Archive.Close()
+	// opts.ui.ArchiveDownloadStarted()
```
eseliger, line 45: TODO: Need to add a noop archive for exec mode.

internal/batches/executor/run_steps.go
```diff
@@ -286,9 +286,10 @@ func executeSingleStep(
 
 	// Where should we execute the steps.run script?
 	scriptWorkDir := workDir
-	if opts.task.Path != "" {
-		scriptWorkDir = workDir + "/" + opts.task.Path
-	}
+	// TODO: This breaks scripts that use `search_result_paths`
```
eseliger, line 289: TODO: put in separate PR, not strictly related.

internal/batches/workspace/bind_workspace.go
```diff
@@ -28,16 +28,22 @@ var _ Creator = &dockerBindWorkspaceCreator{}
 func (wc *dockerBindWorkspaceCreator) Type() CreatorType { return CreatorTypeBind }
 
 func (wc *dockerBindWorkspaceCreator) Create(ctx context.Context, repo *graphql.Repository, steps []batcheslib.Step, archive repozip.Archive) (Workspace, error) {
```
eseliger, line 30: todo: remove all of this, the custom archive implementation should fix that

internal/batches/workspace/git.go
```diff
@@ -25,9 +25,9 @@ func runGitCmd(ctx context.Context, dir string, args ...string) ([]byte, error)
 		"GIT_COMMITTER_EMAIL=batch-changes@sourcegraph.com",
 	}
 	cmd.Dir = dir
-	out, err := cmd.CombinedOutput()
+	out, err := cmd.Output()
```
eseliger, line 28: TODO: Why is this suddenly happening

internal/batches/service/service.go
```diff
@@ -281,12 +286,12 @@ func (svc *Service) BuildTasks(ctx context.Context, attributes *templatelib.Batc
 	return buildTasks(ctx, attributes, workspaces)
 }
 
-func (svc *Service) NewCoordinator(opts executor.NewCoordinatorOpts) *executor.Coordinator {
-	opts.RepoArchiveRegistry = repozip.NewArchiveRegistry(svc.client, opts.CacheDir, opts.CleanArchives)
+func (svc *Service) NewCoordinator(archiveRegistry repozip.ArchiveRegistry, logger log.LogManager, opts executor.NewCoordinatorOpts) *executor.Coordinator {
```
eseliger, line 289: TODO: Can be set on the opts thing directly.

internal/batches/executor/coordinator_test.go
```diff
@@ -228,7 +228,7 @@ func TestCoordinator_Execute(t *testing.T) {
 		},
 		{
 			name: "skip errors",
-			opts: NewCoordinatorOpts{Features: featuresAllEnabled(), SkipErrors: true},
+			opts: NewCoordinatorOpts{Features: featuresAllEnabled()}, // , SkipErrors: true},
```
eseliger, line 231: TODO: Investigate why this has no effect?

internal/batches/workspace/git.go
eseliger: I am seeing 

```
fatal: could not read Password for 'http://executor@localhost:3080': terminal prompts disabled
fatal: could not read Password for 'http://executor@localhost:3080': terminal prompts disabled
M  client/client-api/README.md
diff --git client/client-api/README.md client/client-api/README.md
index 7462bfb..e2eb036 100644
--- client/client-api/README.md
+++ client/client-api/README.md
@@ -1,3 +1,4 @@
 # Client API
 
 Sourcegraph Client API
+Hello World
```
for git status.. why would it ask for credentials for git status 🤔

internal/batches/executor/coordinator_test.go
```diff
@@ -226,41 +225,6 @@ func TestCoordinator_Execute(t *testing.T) {
 				}),
 			},
 		},
-		{
-			name: "skip errors",
-			opts: NewCoordinatorOpts{Features: featuresAllEnabled(), SkipErrors: true},
```
eseliger, line 231: This was not dependent on this variable, so this test tested nothing :D

Piszmog (approved):

internal/batches/log/logger.go
```diff
@@ -0,0 +1,18 @@
+package log
```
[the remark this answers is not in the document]
eseliger: Agreed! I'll check if I can make that happen some time this evening otherwise I'll follow up on that next week!

PR #770, opened 2022-06-07
batches: Mount paths into a container

Piszmog:
Closes [#31790](https://github.com/sourcegraph/sourcegraph/issues/31790).

Add the ability to mount a path on the local machine into a container during a `step`. 

A `path` can point to a file (`./foo.sh` or  `/bar/foo.sh`) or a directory (`./bar` or `/foo/bar/`). The `path` can be an absolute path or a relative path (for example `./foo.sh`). Regardless if the path is absolute or relative, the path must be within the same directory as the batch spec that is being ran (the batch spec directory is considered the "working directory").

Validation/handling of the `mount` field in a `step` is handled in the `src-cli` versus `batcheslib` in `sourcegraph/sourcegraph`. The reason for this was the avoid having conditional logic to figure out if `parseBatchSpec` is being called on a local machine vs behind an API. This also contains the file system logic within the context of `src-cli` which is dealing with the file system.

This PR also includes some general fixes I saw while in the code (for example, package and variable naming collision).

### Test plan

Go Unit Tests and manual testing.

#### Examples

Below are examples of mounting a path that contains a script (python or Go binary in this case) that is ran in the `run` field of a `step`.

##### Python

Example of running a python script that is mounted into the container and ran in a `step`.

###### Batch Spec

```yaml
name: python-script-test
description: Use a python script
on:
  - repositoriesMatchingQuery: file:README.md
steps:
  - run: python /tmp/updater.py
    container: python:latest
    mount:
      - path: ./updater.py
        mountpoint: /tmp/updater.py
changesetTemplate:
  title: Hello from Python
  body: Updated using a Python Script
  branch: python-script-test
  commit:
```

###### Script

```python
#!/usr/bin/env python3
import os.path


def main():
    exists = os.path.exists('README.md')
    if exists:
        with open('README.md', 'a') as f:
            f.write('\nHello from python')


if __name__ == "__main__":
    main()

```

###### Changesets

Screen Shot 2022-06-07 at 14 43 47

##### Binary

Example of running a binary (written in Go) that is mounted into the container and ran in a `step`.

###### Batch Spec

```yaml
name: binary-test
description: Use a binary
on:
  - repositoriesMatchingQuery: file:README.md
steps:
  - run: /tmp/updater
    container: alpine:latest
    mount:
      - path: ./updater # the linux binary
        mountpoint: /tmp/updater
changesetTemplate:
  title: Hello from a binary
  body: Updated using a binary
  branch: binary-test
  commit:
```

###### Binary

Binary was built from Go

```go
package main

import "os"

func main() {
	fileInfo, err := os.Stat("README.md")
	if err != nil {
		panic(err)
	}
	if fileInfo.IsDir() {
		panic("file is a directory")
	}

	f, err := os.OpenFile("README.md", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
	if err != nil {
		panic(err)
	}
	defer f.Close()

	if _, err = f.WriteString("\nhello from go"); err != nil {
		panic(err)
	}
}
```

Built with the following command,

`env GOOS=linux GOGOARCH=amd64 go build -ldflags="-s -w"`

###### Changesets

Screen Shot 2022-06-07 at 14 44 50

##### Directory

The above examples have the `path` pointing to a specific file. Directories can also be mounted in the container. This example reuses the above Python Example but mounts a directory that contains the script.

###### Batch Spec

```yaml
name: python-script-test
description: Use a python script
on:
  - repositoriesMatchingQuery: file:README.md
steps:
  - run: python /tmp/scripts/updater.py
    container: python:latest
    mount:
      - path: ./scripts
        mountpoint: /tmp/scripts
changesetTemplate:
  title: Hello from Python
  body: Updated using a Python Script
  branch: python-script-test
  commit:
```

###### Changesets

Screen Shot 2022-06-07 at 14 42 45


courier-new:
Thank you for such a thorough write-up of your testing examples!! 😄

internal/batches/service/service.go
```diff
@@ -380,9 +381,43 @@ func (svc *Service) ParseBatchSpec(data []byte) (*batcheslib.BatchSpec, error) {
 	if err != nil {
 		return nil, errors.Wrap(err, "parsing batch spec")
 	}
+	if err = handleMount(dir, spec); err != nil {
+		return nil, errors.Wrap(err, "parsing batch spec")
```
courier-new, line 385:
```suggestion
		return nil, errors.Wrap(err, "handling mount")
```

internal/batches/service/service_test.go
```diff
@@ -698,3 +702,268 @@ func TestEnsureDockerImages(t *testing.T) {
...
+			batchSpecDir: tempDir,
+			rawSpec: fmt.Sprintf(`
+name: test-spec
+description: A test spec
+steps:
+  - run: /tmp/sample.sh
+    container: alpine:3
+    mount:
+      - path: %s
+        mountpoint: /tmp/sample.sh
```
courier-new, line 750: Could it be worth adding a test case with multiple steps/multiple mounted files as well?

courier-new (approved):
I defer to others on the team for a heavier review of the docker-specific interactions, but the mounted file handling and other changes look great to me. 🙂

internal/batches/executor/run_steps.go
```diff
@@ -303,6 +303,11 @@ func executeSingleStep(
 		args = append(args, "--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", source.Name(), target))
 	}
 
+	// mount any paths on the local system to the docker container
+	for _, mount := range step.Mount {
+		args = append(args, "--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", mount.Path, mount.Mountpoint))
+	}
```
eseliger, line 309: Can we add a flag to `executionOpts` like `allowFileMounts` that we wire through the executor and coordinator layer, so that we can explicitly disable it in `batch_exec.go` (the entrypoint for SSBC?) Feels like a good safety measure, in case someone can still sneak a batch spec into the server that contains mounts (I think we currently don't reject them in the SSBC `createBatchSpecFromRaw` API, just in the editor UI)

internal/batches/service/service_test.go
```diff
@@ -698,3 +702,268 @@ func TestEnsureDockerImages(t *testing.T) {
...
+			spec, err := svc.ParseBatchSpec(test.batchSpecDir, []byte(test.rawSpec))
+			if test.expectedErr != nil {
+				assert.Equal(t, test.expectedErr.Error(), err.Error())
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, test.expectedSpec, spec)
+			}
+		})
+	}
+}
```
eseliger, line 969: great test suite!

internal/batches/service/service.go
```diff
@@ -380,9 +381,43 @@ func (svc *Service) ParseBatchSpec(data []byte) (*batcheslib.BatchSpec, error) {
...
+	}
 	return spec, nil
 }
 
+func handleMount(dir string, spec *batcheslib.BatchSpec) error {
```
eseliger, line 390: This sounds like a more precise description
```suggestion
func validateMountPaths(safeDirectory string, spec *batcheslib.BatchSpec) error {
```

cmd/src/batch_common.go
```diff
@@ -9,11 +9,13 @@ import (
 	"os/exec"
```
eseliger, line 9: There's a file called `batch_remote` in this directory that uploads the spec file for SSBC execution, we should add a gate there that the `validateMountPaths` method rejects if there is a mount entry present. The API should ideally reject this anyways (to be implemented), but a nice actionable error message can't hurt, I think.

cmd/src/batch_common.go
```diff
@@ -477,7 +479,13 @@ func parseBatchSpec(file string, svc *service.Service) (*batcheslib.BatchSpec, s
 		return nil, "", errors.Wrap(err, "reading batch spec")
 	}
 
-	spec, err := svc.ParseBatchSpec(data)
+	p, err := filepath.Abs(file)
+	if err != nil {
+		return nil, "", errors.Wrap(err, "batch spec path")
+	}
+	dir, _ := filepath.Split(p)
```
LawnGnome, line 486: For clarity, I probably would have gone with:

```suggestion
	dir := filepath.Dir(p)
```

Unless there's a behavioural quirk in one of `Dir` or `Split` that I've forgotten. (Which is entirely possible!)

cmd/src/batch_common.go
```diff
@@ -477,7 +479,13 @@ func parseBatchSpec(file string, svc *service.Service) (*batcheslib.BatchSpec, s
 		return nil, "", errors.Wrap(err, "reading batch spec")
 	}
 
-	spec, err := svc.ParseBatchSpec(data)
+	p, err := filepath.Abs(file)
```
eseliger, line 482: `file` can also be an empty string or `-`, which indicate "read from stdin". Should we in that case fall back to the working directory? Or error? Regardless, one of the two we need to do here.

eseliger (commented):
Great PR description and testing! I left a few comments inline

internal/batches/executor/coordinator.go
```diff
@@ -23,7 +23,7 @@ type taskExecutor interface {
 	Wait(context.Context) ([]taskResult, error)
 }
 
-// Coordinates coordinates the execution of Tasks. It makes use of an executor,
+// Coordinator coordinates the execution of Tasks. It makes use of an executor,
```
LawnGnome, line 26: :x: no longer marks the spot.

internal/batches/executor/executor.go
```diff
@@ -147,20 +148,20 @@ func (x *executor) do(ctx context.Context, task *Task, ui TaskExecutionUI) (err
 	ui.TaskStarted(task)
 
 	// Let's set up our logging.
-	log, err := x.opts.Logger.AddTask(util.SlugForPathInRepo(task.Repository.Name, task.Repository.Rev(), task.Path))
+	l, err := x.opts.Logger.AddTask(util.SlugForPathInRepo(task.Repository.Name, task.Repository.Rev(), task.Path))
```
LawnGnome, line 151: Good catch — I often shadow a package name inadvertently and then wonder why things break later.

internal/batches/service/service.go
```diff
@@ -380,9 +381,43 @@ func (svc *Service) ParseBatchSpec(data []byte) (*batcheslib.BatchSpec, error) {
...
+				// Try to build the absolute path since Docker will only mount absolute paths
+				p = filepath.Join(dir, p)
+			}
+			pathInfo, err := os.Stat(p)
+			if os.IsNotExist(err) {
+				return errors.Newf("step %d mount path %s does not exist", i+1, p)
+			} else if err != nil {
+				return errors.Wrapf(err, "step %d mount path validation", i+1)
+			} else {
+				if !strings.HasPrefix(p, dir) {
```
LawnGnome, line 404: I'm wondering if we'll have edge cases here around symlinks — for example, if the user has their spec in `/Users/foo/bar`, but there's a symlink `/Users/foo/quux` to it and they use the latter in an absolute path provided to `mount`.

The quick and easy solution here might be to call [`filepath.EvalSymlinks`](https://pkg.go.dev/path/filepath#EvalSymlinks) on `p` and `dir` before performing this check. (I guess we only really need to do `dir` once.)

cmd/src/batch_common.go
LawnGnome, line 482: We'll need to special case `file == "-"` and `file == ""`, per the code in `batchOpenFileFlag` to handle reading from stdin. (In that case, `dir` is presumably going to be the current working directory.)

internal/batches/executor/run_steps.go
```diff
@@ -303,6 +303,11 @@ func executeSingleStep(
 		args = append(args, "--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", source.Name(), target))
 	}
 
+	// mount any paths on the local system to the docker container
+	for _, mount := range step.Mount {
+		args = append(args, "--mount", fmt.Sprintf("type=bind,source=%s,target=%s,ro", mount.Path, mount.Mountpoint))
```
LawnGnome, line 308: Edge case: what happens if `mount.Path` or `mount.Mountpoint` include a literal comma?

LawnGnome (approved):
This looks really good! I have a few questions, but nothing fundamental to how this is implemented.

internal/batches/executor/run_steps.go
eseliger: That sounds scary 😬 I think the same question applies to the `files` property.

internal/batches/service/service.go
eseliger: Hmm, don't you think that could actually be intentional? 

Like I put a symlink down to use it in a different WD without copying it?

Or do I misunderstand what the problem is here?

LawnGnome: I guess my question is more: should we consider `/Users/foo/bar` and `/Users/foo/quux` to be the same for the purposes of this check, if they point to the same directory?

eseliger: I, having thought about this at least an entire 30 seconds would say we should not. 

If you symlink `/Users/foo/spec.yml` into `/Users/erik/workspace/spec.yml` and use that one, then the intentional workspace location IMO would be `/Users/erik/workspace` and not `/Users/foo`. I only symlinked the spec file and not it's surroundings. If I symlinked the entire directory `/Users/foo` to `/Users/erik/workspace`, then I should just be using relative paths or paths to the aliased name. 
At least to my 12:30am brain that sounds logical. Please challenge me 😬

internal/batches/service/service.go
Piszmog: Initially I went with `validate` in the function name to match the other functionality going on. But this is a bit more than "validate" It is also updating the batch spec (changing the mount path to the absolute path). So I felt `handle` was a bit closer to the functionality.

But thinking more about it, I think `validate` is "correct" since that is like 99% of the functionality

internal/batches/service/service.go
```diff
@@ -380,9 +381,43 @@ func (svc *Service) ParseBatchSpec(data []byte) (*batcheslib.BatchSpec, error) {
...
+					return errors.Newf("step %d mount path is not in the same directory or subdirectory as the batch spec", i+1)
+				}
+				// Mounting a directory on Docker must end with the separator. So, append the file separator to make
+				// users' lives easier.
+				if pathInfo.IsDir() && !strings.HasSuffix(p, string(filepath.Separator)) {
+					p += string(filepath.Separator)
+				}
+				// Update the mount path to the absolute path so building the absolute path (above) does not need to be
+				// redone when adding the mount argument to the Docker container.
+				step.Mount[j].Path = p
```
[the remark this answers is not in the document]
Piszmog: I don't see any issues with caching, but I am not super familiar with the cache yet. 

Actually, I was having issues with the cache where I would only make a change to the script file and would rerun the batch spec and nothing would change because of the cache (for example changing the text to append to the `README.md` from `Hello from python` to `Hello from python1`).

I was getting slightly annoyed with this issue, and I am sure another user will too. Is this something we want to deal with in this Issue or in a follow up Issue?

internal/batches/executor/run_steps.go
Piszmog: If a comma is provided, Docker gets angry because it thinks the value after the comma is a field and it is not in the right format.

```text
invalid argument "type=bind,source=/Users/randell/Documents/development/batch-spec-playground/some,bad/foo.py,target=/tmp/some/foo.py,ro" for "--mount" flag: invalid field 'bad/foo.py' must be a key=value pair

invalid argument "type=bind,source=/Users/randell/Documents/development/batch-spec-playground/foo.py,target=/tmp/some,bad/foo.py,ro" for "--mount" flag: invalid field 'bad/foo.py' must be a key=value pair
```

Regardless, I'll add a check for this character and any others.

internal/batches/service/service.go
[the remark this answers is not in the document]
Piszmog: I just saw the flag `-clear-cache`. So I guess, it is really not a problem? Maybe we should have some docs around the mount and cache and communicate that the `-clear-cache` flag is available when updating the script[s]?

internal/batches/executor/run_steps.go
Piszmog: Yea, we are not rejecting `mount` in the API layer at all.

internal/batches/service/service.go
```diff
@@ -380,9 +381,53 @@ func (svc *Service) ParseBatchSpec(data []byte) (*batcheslib.BatchSpec, error) {
...
+			}
+			pathInfo, err := os.Stat(p)
+			if os.IsNotExist(err) {
+				return errors.Newf("step %d mount path %s does not exist", i+1, p)
+			} else if err != nil {
+				return errors.Wrapf(err, "step %d mount path validation", i+1)
+			}
+			if !strings.HasPrefix(p, batchSpecDir) {
+				return errors.Newf("step %d mount path is not in the same directory or subdirectory as the batch spec", i+1)
+			}
```
eseliger, line 417: Can we move the static checks in here into the lib parse method, so SSBC can benefit from the same validation? Only works for the ones that don't need to access the FS, of course.

internal/batches/executor/executor.go
```diff
@@ -62,6 +63,7 @@ type newExecutorOpts struct {
 	Parallelism          int
 	Timeout              time.Duration
 	TempDir              string
+	AllowPathMounts      bool
 	WriteStepCacheResult func(ctx context.Context, stepResult execution.AfterStepResult, task *Task) error
```
eseliger, line 67: I think this one you also want to restrict when mounts are used. Otherwise, step-wise caching wouldn't trigger on changed files either.

eseliger (approved):
Great, very clean :)

internal/batches/service/service.go
Piszmog: I'll need to wait for the `sourcegraph` PR to be merged so I can update `go.mod` to the latest version

PR #773, opened 2022-06-08
batches: ensure .gitconfig email is properly inserted in batch spec

BolajiOlajide:
In `src-cli`, we use the [html/template](https://pkg.go.dev/html/template) library to substitute values, however, characters like `+` gets converted to it's unicode equivalent `&043;`

Before
CleanShot 2022-06-08 at 12 03 09@2x

After
CleanShot 2022-06-08 at 11 57 31@2x

### Test plan

* Enable email privacy on your Github account.
CleanShot 2022-06-08 at 11 40 23@2x
* Set your github-private email address in your global config with the command
```sh
git config --global add user.email <private_email_address>
```
* Create a batch spec with the command `src batch new -f batch.spec.yml`
* The email address should be inserted into the template as plain text instead of unicode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


BolajiOlajide:
> WOW that must've taken forever to find 😨

I was almost pulling my hair out at some point lol. Thankfully, all I needed was a 30-minute nap and I immediately checked the template import lol.

PR #777, opened 2022-06-12
Don't write execution results to SSBC cache and remove unnecessary additional call to workspace.Diff

eseliger:
This is the src-cli portion of https://github.com/sourcegraph/sourcegraph/pull/36758. It also optimizes the execution generally, because we run one less workspace.Diff. 
After this PR, nothing requires us to still write or read execution results from the cache. Since this is a bigger code cleanup/refactor, I filed a separate ticket to address this and put it in sustaining. This PR mostly focuses on getting the Diff call out of the execution and stops to write the unnecessary cache result to the JSON logs so we send less log data.

### Test plan

Tested together with the linked sg/sg PR. Works. Also, this code path is very well tested and it's still passing even without that additional call.


courier-new (approved):
Nice. 😎

LawnGnome (approved):
Obviously will need a fresh `go get` before we can merge, but otherwise LGTM.

Issue #755, opened 2022-05-27
batches: getting changed files in step: applying cached diff: error: unrecognized input: exit status 128


sfllaw:
Sometimes, when updating a batch change that has already been applied, the cached diff cannot be applied:

``` console
sfllaw@hostname:~$ src batch apply -v -f batch-changes.yaml 
✅ Parsing batch spec
✅ Resolving namespace
✅ Preparing container images  ██████████████████████████████████████████████
🚧 Workspace creator: bind
✅ Set workspace type
✅ Resolved 137 repositories
✅ Found 137 workspaces with steps to execute
✅ Found 0 cached changeset specs; 137 tasks need to be executed

✅ Executing... (137/137, 24 errored)  ██████████████████████████████████████

❌ 24 errors:
   github.com/sfllaw/repo:
   getting changed files in step: applying cached diff: 'git apply -p0 /home/sfllaw/.cache/sourcegraph/batch/bind-workspace-test-1611338046' failed: error: unrecognized input: exit status 128
   Log: /tmp/changeset-github.com-sfllaw-repo-5e3e8f2c3e2780ec35240790d1971b2c0b7ed9b0.3325954318.log
```

If we look into the cache, we see that the diff is empty!
``` console
sfllaw@hostname:~/.cache/sourcegraph/batch$ jq -C . < github.com-sfllaw-repo-5e3e8f2c3e2780ec35240790d1971b2c0b7ed9b0/j7ViTNOX3Gkk_8XnX-MbZQ.json
{
  "diff": "",
  "changedFiles": {
    "modified": null,
    "added": null,
    "deleted": null,
    "renamed": null
  },
  "outputs": {},
  "path": ""
}
```

This is exactly the behaviour you’d expect from an empty patch:
``` console
sfllaw@hostname:~/src/repo$ echo -n '' > patch
sfllaw@hostname:~/src/repo$ git apply -p0 patch
error: unrecognized input
sfllaw@hostname:~/src/repo$ echo $?
128
```

sfllaw:
cc: @sourcegraph/batchers (@eseliger @LawnGnome @courier-new @adeola-ak @BolajiOlajide @Piszmog @malomarrec @chrispine @danielmarquespt)

BolajiOlajide:
Hello @sfllaw,

Thanks for   this issue. 
We've added it to our planning discussion list and we will respond after.

Cheers!

LawnGnome:
[...]

PR #778, opened 2022-06-12
Only apply cache result progress if resulted in diff

eseliger:
A user reported this, we store execution step results in the cache also when no diff was generated (rightfully), but then we should only apply the patch when there is actually anything to do.

Also, I initially couldn't reproduce this locally because we swallowed this error in volume mode. It would only happen in bind mode.

Closes https://github.com/sourcegraph/src-cli/issues/755

### Test plan

Verified this does properly fail in volume mode now, too and don't fail anymore at all.

### DIFF
--- a/internal/batches/executor/run_steps.go
+++ b/internal/batches/executor/run_steps.go
@@ -121,8 +121,12 @@ func runSteps(ctx context.Context, opts *executionOpts) (result execution.Result
 			stepContext.Steps.Changes = previousStepResult.Files
 			stepContext.Outputs = opts.task.CachedResult.Outputs
 
-			if err := ws.ApplyDiff(ctx, []byte(opts.task.CachedResult.Diff)); err != nil {
-				return execResult, nil, errors.Wrap(err, "getting changed files in step")
+			// If the previous steps made any modifications in the workspace yet,
+			// apply them.
+			if opts.task.CachedResult.Diff != "" {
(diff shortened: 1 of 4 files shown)


internal/batches/executor/run_steps.go
```diff
@@ -119,8 +119,12 @@ func runSteps(ctx context.Context, opts *executionOpts) (result execution.Result
 			stepContext.Steps.Changes = previousStepResult.Files
 			stepContext.Outputs = opts.task.CachedResult.Outputs
 
-			if err := workspace.ApplyDiff(ctx, []byte(opts.task.CachedResult.Diff)); err != nil {
-				return execResult, nil, errors.Wrap(err, "getting changed files in step")
+			// If the previous steps made any modifications in the workspace yet,
+			// apply them.
+			if opts.task.CachedResult.Diff != "" {
```
Piszmog, line 124: Are there existing tests to cover when `CachedResult.Diff` is blank and not blank?

Piszmog (approved):
Looks good, just one question

courier-new (approved):
Wow we were REALLY swallowing that error in volume mode, huh? 😬 Thanks for pulling it back out, and fixing the error itself haha.

LawnGnome (approved):

internal/batches/executor/run_steps.go
eseliger: Another PR just found this bug, and it has a test for it so will merge this and then in the next PR you'll get that test :)

PR #781, opened 2022-06-14
Fix step caching

eseliger:
Src side change for https://github.com/sourcegraph/sourcegraph/pull/37240

This fixes two things, that I found today:
- It changes Stdout and Stderr on the StepResult struct to string, it is part of the marshalled cache result and *bytes.Buffer cannot be marshalled into JSON. This caused stdout/stderr to always be empty, thus breaking executions where `${{ previous_step.stdout }}` is used.
- It fixes a misconception of the `PreviousStepResult`. When running the actual execution, it will eventually _become_ the result of the previous step, but from the perspective of that particular cache result, it has to be the result from THAT step execution. Otherwise caching would only work from step 2 on and skip 1 step still.

### Test plan

Test case and manual verification.


internal/batches/executor/run_steps.go
```diff
@@ -179,10 +181,10 @@ func runSteps(ctx context.Context, opts *executionOpts) (result execution.Result
...
 		}
 		stepResult := execution.AfterStepResult{
-			StepIndex:          i,
-			Diff:               string(stepDiff),
-			Outputs:            make(map[string]interface{}),
-			PreviousStepResult: stepContext.PreviousStep,
+			StepIndex:  i,
+			Diff:       string(stepDiff),
+			Outputs:    make(map[string]interface{}),
+			StepResult: result,
```
eseliger, line 187: This was just weirdly wrong :D this meant that we got.. the second last result from the cache entry?

courier-new (approved):

internal/batches/executor/run_steps.go
mrnugget: Hmmmmm, I guess this is on me. This part of the code is really tricky - I hope we do have tests covering all the edge cases 😬

PR #783, opened 2022-06-15
docker: use a short timeout when inspecting images

LawnGnome:
Docker Desktop seems to be prone to occasionally entering states where it is nominally available (in the sense that it listens for connections), but can't actually do anything useful. Memory exhaustion events from previous `src batch` runs tend to be the culprit here.

This seems to be most easily detectable by looking at the behaviour of `docker image inspect`. This command doesn't have to hit the network, nor does it perform any real work, so it should always be "quick". If it's not, that implies that Docker is not in a happy place.

This commit adds a five second timeout to `docker image inspect` invocations, along with a secret environment variable to adjust this behaviour. If `docker` times out, then a (hopefully) helpful error message is generated to hint towards the things that the user should investigate next (basically, "is Docker really working?").

The only real concern I have with this is that I've basically picked five seconds out of thin air: practically, Docker for Mac seems to be able to do this in tens of milliseconds, but Sourcegraph has bought me a rather nice laptop. I have trouble imagining a scenario where multiple second delays are common where there aren't other serious issues, but the world is a large, weird place.

### Test plan

There's decent test coverage here, and I've added a new test case for this specific change.

I also tested this manually in these scenarios:

* Regular operation (to confirm nothing actually changed)
* `docker` fails to respond (by `killall -STOP dockerd` in the Docker Desktop VM)
* Set the timeout to `0` via the environment variable so `src` never actually waits and immediately fails

### DIFF
--- a/internal/batches/docker/image.go
+++ b/internal/batches/docker/image.go
@@ -4,8 +4,10 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"os"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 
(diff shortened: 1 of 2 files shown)


internal/batches/docker/image.go
```diff
@@ -66,19 +68,50 @@ func (image *image) Ensure(ctx context.Context) error {
 	image.ensureOnce.Do(func() {
 		image.ensureErr = func() (err error) {
 			inspectDigest := func() (string, error) {
-				// TODO!(sqs): is image id the right thing to use here? it is NOT
-				// the digest. but the digest is not calculated for all images
-				// (unless they are pulled/pushed from/to a registry), see
-				// https://github.com/moby/moby/issues/32016.
```
LawnGnome, line 72: This `TODO` was outdated: there's actually a lengthy comment earlier in the file explaining what "digest" means in this context, and how it relates to the public digest. Whoever last touched this really should have removed this then. 😬

courier-new (approved):
This seems like a great improvement. And worst case if 5s is too sensitive, it's fortunately very easy to release a patch release for src-cli to bump it up. 🙂

internal/batches/docker/image_test.go
```diff
@@ -76,6 +79,26 @@ func TestImage_Digest(t *testing.T) {
 			}
 		})
 	}
+
+	t.Run("docker timeout", func(t *testing.T) {
+		// We're more interested in the context error handling than the parsing
+		// of the timeout, and we don't want to slow down the test, so we're
+		// going to construct a context that has already exceeded its deadline
+		// at the point it is provided to Digest.
+		ctx, cancel := context.WithTimeout(context.Background(), -1*time.Second)
```
eseliger, line 88: Lol, didn't know that's valid!

eseliger (approved):
Nice approach!

Piszmog (approved):

internal/batches/docker/image_test.go
LawnGnome: [Turns out I didn't even invent it!](https://sourcegraph.com/search?q=context:global+lang:go+context.WithTimeout%28...%2C+-...%29&patternType=structural) I particularly like that it's used in the runtime's own unit tests.

eseliger: Nice, so they would break their own tests before it ships in a go release 😆

PR #786, opened 2022-06-15
batches: use Docker CPU count as default parallelism, not GOMAXPROCS

LawnGnome:
This PR changes the default parallelism of `src batch preview|apply` runs to be based on the CPU cores reported by `docker info`.

Practically, this means that Docker Desktop users will now have their parallelism based on the VM Docker runs within, not their host machine. The trade off here is that Docker Desktop users will likely have their default parallelism cut significantly, but this is also likely to make issues caused by the Docker VM having its resources exhausted considerably less likely.

### Implementation notes

* I've migrated the previous `docker version` preflight to use `docker info` instead, which means we can get the CPU count and preflight Docker at the same time.
* Since `docker info` should always be quick, this reuses the same timeout behaviour as #783, which has been renamed from being `image inspect` specific to being used for any "fast commands" issued to Docker. The behaviour is otherwise unchanged.
* In testing this, I found an issue where `os/exec` is inconsistent in terms of the error returned when a deadline is exceeded; I've updated both places we use the fast command context to look both at the returned error and the context error.

### Screenshot of the timeout

image

### Test plan

Added unit tests, and tested this in various scenarios locally (Docker not running; Docker running but unresponsive; Docker running normally).

### DIFF
--- a/cmd/src/batch_exec.go
+++ b/cmd/src/batch_exec.go
@@ -92,7 +92,12 @@ func executeBatchSpecInWorkspaces(ctx context.Context, ui *ui.JSONLines, opts ex
 	if err := checkExecutable("git", "version"); err != nil {
 		return err
 	}
-	if err := checkExecutable("docker", "version"); err != nil {
+
+	// In the past, we checked `docker version`, but now we retrieve the number
+	// of CPUs, since we need that anyway and it performs the same check (is
+	// Docker working _at all_?).
+	parallelism, err := getBatchParallelism(ctx, opts.flags.parallelism)
+	if err != nil {
 		return err
 	}
 
@@ -112,7 +117,7 @@ func executeBatchSpecInWorkspaces(ctx context.Context, ui *ui.JSONLines, opts ex
 	if len(input.Steps) > 0 {
 		ui.PreparingContainerImages()
 		images, err := svc.EnsureDockerImages(
-			ctx, input.Steps, opts.flags.parallelism,
+			ctx, input.Steps, parallelism,
 			ui.PreparingContainerImagesProgress,
 		)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


cmd/src/batch_exec.go
```diff
@@ -138,7 +143,7 @@ func executeBatchSpecInWorkspaces(ctx context.Context, ui *ui.JSONLines, opts ex
 		Cache:         &executor.ServerSideCache{Writer: ui},
 		SkipErrors:    opts.flags.skipErrors,
 		CleanArchives: opts.flags.cleanArchives,
-		Parallelism:   opts.flags.parallelism,
+		Parallelism:   parallelism,
```
eseliger, line 146: fun fact, it's unused in SSBC 😆 PR cleaning this mess here up coming soon :)

internal/batches/docker/info.go
```diff
@@ -0,0 +1,35 @@
...
+// NCPU returns the number of CPU cores available to Docker.
+func NCPU(ctx context.Context) (int, error) {
+	dctx, cancel, err := withFastCommandContext(ctx)
+	if err != nil {
+		return 0, err
+	}
+	defer cancel()
+
+	args := []string{"info", "--format", "{{ .NCPU }}"}
+	out, err := exec.CommandContext(dctx, "docker", args...).CombinedOutput()
```
eseliger, line 22: CombinedOutput or Output?

internal/batches/docker/info.go
```diff
@@ -0,0 +1,35 @@
...
+	if errors.IsDeadlineExceeded(err) || errors.IsDeadlineExceeded(dctx.Err()) {
+		return 0, newFastCommandTimeoutError(dctx, args...)
+	} else if err != nil {
+		return 0, err
+	}
+
+	dcpu, err := strconv.Atoi(string(bytes.TrimSpace(out)))
+	if err != nil {
+		return 0, err
```
eseliger, line 31:
```suggestion
		return 0, errors.Wrap(err, "parsing docker cpu count")
```

eseliger (approved):

internal/batches/docker/info.go
LawnGnome: Combined — Docker outputs all non-container-output on stderr.

eseliger: interesting!

Piszmog (approved):

PR #787, opened 2022-06-16
batches: upsert batch change when running `src batch remote

### LINKED ISSUE #780: SSBC: `src batch remote` preview link returns 404 if batch change did not already exist
In `src/batch_remote.go` we [create the batch spec](https://sourcegraph.com/github.com/sourcegraph/src-cli/-/blob/cmd/src/batch_remote.go?L82%3A48=) but we forget to make sure the batch change exists first. 🙃 It was previously possible to view a remote execution in isolation without a parent batch change, but this is no longer possible. However, with src-cli, it's possible to run a remote execution for a batch change that already exists or doesn't yet exist, whilst the mutation that the frontend UI uses to create the empty batch change (`CreateEmptyBatchChange`) fails if a batch change by that name already exists, so we either need to:

- Create a new mutation like `UpsertEmptyBatchChange` that doesn't fail if a batch change by that name already exists, or
- Add a new parameter to `CreateEmptyBatchChange` like `allowUpsert` that changes the behavior of the mutation so that it doesn't fail.

We should then call that mutation from `src/batch_remote.go` before we call `CreateBatchSpecFromRaw`, so that the batch change exists to load from for the preview link.

BolajiOlajide:
Closes #780 

### Test plan

* Create a server-side batch change with the command `src batch remote`
* When the command is done, the batch change should be accessible via the execution link outputted in the terminal

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
-- Execution URL when running `src batch remote` doesn't result in a 404. [sourcegraph/src-cli](https://github.com/sourcegraph/src-cli/pull/787)
+- The preview link shown when running `src batch remote` to create a new batch change no longer 404s. [sourcegraph/src-cli](https://github.com/sourcegraph/src-cli/pull/787)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- Execution URL when running `src batch remote` doesn't result in a 404. [sourcegraph/src-cli](https://github.com/sourcegraph/src-cli/pull/787)
+
 ### Removed
 
 ## 3.40.11
--- a/cmd/src/batch_remote.go
+++ b/cmd/src/batch_remote.go
@@ -78,6 +78,13 @@ Examples:
 		}
 		ui.ResolvingNamespaceSuccess(namespace.ID)
 
+		ui.SendingBatchChange()
+		err = svc.UpsertBatchChange(ctx, spec.Name, namespace.ID)
+		if err != nil {
+			return err
+		}
+		ui.SendingBatchChangeSuccess()
+
 		ui.SendingBatchSpec()
 		batchSpecID, err := svc.UpsertBatchSpecInput(
 			ctx,
--- a/cmd/src/batch_remote.go
+++ b/cmd/src/batch_remote.go
@@ -79,7 +79,7 @@ Examples:
 		ui.ResolvingNamespaceSuccess(namespace.ID)
 
 		ui.SendingBatchChange()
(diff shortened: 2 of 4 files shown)


eseliger (approved):

Piszmog (approved):

cmd/src/batch_remote.go
```diff
@@ -78,8 +78,15 @@ Examples:
...
+		ui.SendingBatchChange()
+		batchChangeName, err := svc.UpsertBatchChange(ctx, spec.Name, namespace.ID)
+		if err != nil {
+			return err
+		}
+		ui.SendingBatchChangeSuccess()
+
 		ui.SendingBatchSpec()
-		batchSpecID, err := svc.UpsertBatchSpecInput(
+		batchSpecID, err := svc.CreateBatchSpecFromRaw(
```
BolajiOlajide, line 89: @LawnGnome Had to switch this to `CreateBatchSpecFromRaw` because `UpsertBatchSpecInput` tries to delete the last spec belonging to the batch change, and that results in an error because of the FOREIGN constraint on `batch_changes_batch_spec_id_fkey`.

This is closely similar to what we have in the UI currently.

internal/batches/service/remote.go
```diff
@@ -15,6 +15,96 @@ func (svc *Service) areServerSideBatchChangesSupported() error {
...
+        allowIgnored: $allowIgnored,
+        allowUnsupported: $allowUnsupported,
+        noCache: $noCache,
+    ) {
+        id
+    }
+}
+`
+
+func (svc *Service) CreateBatchSpecFromRaw(
```
courier-new, line 77: Should we still keep `UpsertBatchSpecInput` in the service layer here if no one's using it anymore? No strong feelings, just a thought.

courier-new (approved):
Thanks for this!

cmd/src/batch_remote.go
LawnGnome: That's fine for me.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Fixed
 
+- Execution URL when running `src batch remote` doesn't result in a 404. [sourcegraph/src-cli](https://github.com/sourcegraph/src-cli/pull/787)
```
LawnGnome, line 22: I think a user would have trouble parsing this, mostly because I had trouble parsing it. 😄 

How about:

```suggestion
- The preview link shown when running `src batch remote` to create a new batch change no longer 404s. [sourcegraph/src-cli](https://github.com/sourcegraph/src-cli/pull/787)
```

LawnGnome (approved):
Nice work!

internal/batches/service/remote.go
BolajiOlajide: I have no strong feelings about.I'll remove it.

CHANGELOG.md
BolajiOlajide: Thanks. I was thinking about the best way to phrase that actually. hehe

PR #793, opened 2022-06-21
batches: handle interrupt signals in `batch preview`

### LINKED ISSUE #775: Control+C does not work when reading batch spec as standard input
A user is not able to exit with `ctrl+c` from `src-cli` when providing a batch spec as standard input.

### Steps

Run the following command,

`src batch preview -f -`

Try to exit the program with `ctrl+c` or `ctrl+x`.

BolajiOlajide:
Closes #775 

### Test plan

Confirm `Ctrl + C` terminates the `src batch preview -f -` command.

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"syscall"
 	"time"
 
 	"github.com/mattn/go-isatty"
@@ -533,7 +534,7 @@ func checkExecutable(cmd string, args ...string) error {
 func contextCancelOnInterrupt(parent context.Context) (context.Context, func()) {
 	ctx, ctxCancel := context.WithCancel(parent)
 	c := make(chan os.Signal, 1)
-	signal.Notify(c, os.Interrupt)
+	signal.Notify(c, os.Interrupt, syscall.SIGTERM)
 
 	go func() {
 		select {
--- a/cmd/src/batch_preview.go
+++ b/cmd/src/batch_preview.go
@@ -5,10 +5,10 @@ import (
 	"flag"
 	"fmt"
 
+	"github.com/sourcegraph/sourcegraph/lib/output"
+
 	"github.com/sourcegraph/src-cli/internal/batches/ui"
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
-
-	"github.com/sourcegraph/sourcegraph/lib/output"
 )
 
 func init() {


LawnGnome (approved):
Makes sense to me.

BolajiOlajide:
Did a bit more digging into this and the fix doesn't quite work as expected. 

Sending `Ctrl + C` sends the `SIGINT` signal, however for some reason it sends the `SIGTERM` signal when the terminal is frozen and the terminal wants to force quit the current session.
However, once the [execUI](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Asourcegraph%2Fsrc-cli.git&branch=bo%2Fhonor-sigterm-while-running-batch-preview&file=cmd%2Fsrc%2Fbatch_preview.go&editor=VSCode&version=2.2.4&start_row=47&start_col=2&end_row=53&end_col=3&utm_campaign=vscode-extension&utm_medium=direct_traffic&utm_source=vscode-extension&utm_content=vsce-commands) variable that holds an instance of the [`TUI`](https://sourcegraph.com/-/editor?remote_url=git%40github.com%3Asourcegraph%2Fsrc-cli.git&branch=bo%2Fhonor-sigterm-while-running-batch-preview&file=internal%2Fbatches%2Fui%2Ftui.go&editor=VSCode&version=2.2.4&start_row=27&start_col=0&end_row=34&end_col=1&utm_campaign=vscode-extension&utm_medium=direct_traffic&utm_source=vscode-extension&utm_content=vsce-commands) struct is instantiated, it doesn't honor the context as it has no idea of it, thus ignoring the call to `cancel` from the created `context.WithCancel`.

I updated the goroutine that handles the signal to

```go
go func() {
	select {
	case <-c:
		ctxCancel()
		os.Exit(0)
	case <-ctx.Done():
	}
}()
```

However, this means we always exit automatically after canceling the cotext and I'm not sure this is the right way to go about this. Does anyone have any insight as to the best approach to this?

mrnugget:
I got kinda nerd-sniped reading through this... My first thought was: "isn't there a way to abort an `io.ReadAll` call when a context is cancelled?" and then I found this comment here https://github.com/golang/go/issues/20280#issuecomment-655588450 and quickly copy&pasted&hacked together this: https://github.com/sourcegraph/src-cli/pull/794

It seems to do what it should do, but not sure about unintended side-effects etc. Just wanted to possibly add something to conversation 🙂 

screenshot_2022-06-23_10 26 51@2x

mrnugget:
(Also, have to mention this: the reading stops when you hit `Ctrl-D`, because [`Ctrl-D` sends the EOF value to the application](https://superuser.com/questions/169051/whats-the-difference-between-c-and-d-for-unix-mac-os-x-terminal), which means ends the input reading with an empty file. So, another way to avoid the ugly error would be to print to the user "Hit Ctrl-d to abort" or something, when we detect that input is `os.Stdin`)

BolajiOlajide:
> Cancel reading from stdin on Ctrl-C

Thanks @mrnugget . Testing it out now and reading the attached issue.

PR #796, opened 2022-06-28
batches: add SRC_HEADERS for passing multiple custom headers

### LINKED ISSUE #749: Replace `_` with `-` in custom headers
Headers usually use `-` as opposed to `_`, and some auth proxies may only support `-`. As setting environment variables with dashes is [more complex than underscores](https://unix.stackexchange.com/questions/23659/can-shell-variable-name-include-a-hyphen-or-dash), maybe it makes sense to replace them. This would technically be a breaking change if an auth proxy only supported underscores.

An example of a case where the proxy does not support underscores but does support dashes is Cloudflare Access (with service tokens, [more info](https://community.cloudflare.com/t/access-service-token-headers-underscore-inconsistency/385833)).

BolajiOlajide:
### Test plan

* Added Tests
* Tested on the following shell
- [x] bash
- [x] zsh
- [x] fish
- [x] Powershell

Closes #749

### DIFF
--- a/AUTH_PROXY.md
+++ b/AUTH_PROXY.md
@@ -7,3 +7,9 @@ SRC_HEADER_AUTHORIZATION="Bearer $(curl http://service.internal.corp)" SRC_HEADE
 ```
 
 In this example, the headers `authorization: Bearer my-generated-token` and `extra: metadata` will be threaded to all HTTP requests to your instance. Multiple such headers can be supplied.
+
+When passing multiple headers or headers with dashes, you can make use of the `SRC_HEADERS` environment variable as follow:
+
+```sh
+SRC_HEADERS="AUTHORIZATION:Bearer somerandom_string\nClient-ID:client-one"
+```
--- a/AUTH_PROXY.md
+++ b/AUTH_PROXY.md
@@ -7,6 +7,7 @@ SRC_HEADER_AUTHORIZATION="Bearer $(curl http://service.internal.corp)" SRC_HEADE
 ```
 
 In this example, the headers `authorization: Bearer my-generated-token` and `extra: metadata` will be threaded to all HTTP requests to your instance. Multiple such headers can be supplied.
+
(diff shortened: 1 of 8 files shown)


AUTH_PROXY.md
```diff
@@ -7,3 +7,10 @@ SRC_HEADER_AUTHORIZATION="Bearer $(curl http://service.internal.corp)" SRC_HEADE
 ```
 
 In this example, the headers `authorization: Bearer my-generated-token` and `extra: metadata` will be threaded to all HTTP requests to your instance. Multiple such headers can be supplied.
+An alternative to the above when passing in multiple headers or headers with dashes is to make use of the `SRC_HEADERS` environment variable as follow:
```
LawnGnome, line 10: I think we need an extra newline here for formatting purposes:

```suggestion

An alternative to the above when passing in multiple headers or headers with dashes is to make use of the `SRC_HEADERS` environment variable as follow:
```

cmd/src/headers.go
```diff
@@ -2,27 +2,51 @@ package main
...
+			headers = strings.ReplaceAll(headers, `\n`, "\n")
+			splitHeaders := strings.Split(headers, "\n")
+
+			for _, h := range splitHeaders {
+				p := strings.SplitN(h, ":", 2)
+				if len(parts) != 2 || p[1] == "" {
+					continue
+				}
+
+				additionalHeaders[strings.ToLower(p[0])] = strings.Trim(p[1], " ")
```
LawnGnome, line 45: We should probably trim `p[0]` as well in case somebody gives us something like `Authorization : foo bar`. (Some people really like spaces.)

LawnGnome (approved):
Couple of trivial nits, but LGTM otherwise.

Not a blocker, but probably worth at least thinking slightly about: do we have any idea how this would go on Windows? If the answer is "it won't work", that's fine, but we should probably call it out in `WINDOWS.md`.

Piszmog (approved):

PR #797, opened 2022-07-07
batches: Refactor caching to get rid of execution cache results altogether

eseliger:
This PR cleans up the caching mess that I left behind after the refactors for SSBC to get rid of execution results. We now finally don't have a distinction between the two at all. There are simply step results and if the step is the last step to execute, we are done and build the changeset specs from it. 
Additionally, we save a few passes to parse git diffs and such, which is always nice. 

Accompanied by https://github.com/sourcegraph/sourcegraph/pull/38311

### Test plan

Rely on test suite which is (except renames and some caching number tweaks) untouched and manually verified everything still works.

### DIFF
--- a/internal/batches/executor/run_steps.go
+++ b/internal/batches/executor/run_steps.go
@@ -13,7 +13,6 @@ import (
 
 	batcheslib "github.com/sourcegraph/sourcegraph/lib/batches"
 	"github.com/sourcegraph/sourcegraph/lib/batches/execution"
-	"github.com/sourcegraph/sourcegraph/lib/batches/git"
 	"github.com/sourcegraph/sourcegraph/lib/batches/template"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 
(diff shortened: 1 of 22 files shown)


internal/batches/service/service.go
```diff
@@ -419,6 +419,7 @@ func validateMount(batchSpecDir string, spec *batcheslib.BatchSpec, isRemote boo
 			}
 			// Update the mount path to the absolute path so building the absolute path (above) does not need to be
 			// redone when adding the mount argument to the Docker container.
+			// TODO: Can this mess with caching? We wouldn't be doing that server-side.
```
eseliger, line 422: TODO For later when we enable it for server-side, not for now.

internal/batches/executor/run_steps.go
```diff
@@ -26,73 +25,80 @@ import (
...
-			// Build the execution result from the step result.
-			changes, err := git.ChangesInDiff([]byte(opts.task.CachedResult.Diff))
-			if err != nil {
-				return execResult, nil, errors.Wrap(err, "parsing cached step diff")
-			}
-
-			execResult.Diff = opts.task.CachedResult.Diff
-			execResult.ChangedFiles = &changes
-			stepResults = append(stepResults, opts.task.CachedResult)
+			return stepResults, nil
```
eseliger, line 93: todo: we don't return any stepResults here, can this be problematic?

courier-new (approved):
Didn't do a super deep dive but generally looks good to me. 🙂 Especially appreciative of all the extra comments you added to `executor/run_steps.go`! 🙏

PR #798, opened 2022-07-08
Refactor execution to be able to use runSteps in exec mode directly

eseliger:
To reduce the overhead of things that are unnecessary for exec mode, I've refactored the code to be able to reuse the RunSteps method directly. This will be helpful when designing a potential k8s adapter for executors that will need to be able to split the work of a single workspace across multiple jobs.

This yields a few nice side-effects:
- Net reduction of code
- It's more clear now which flags are actually required to run the execution so I could get rid of a few more things we had to configure but weren't actually used in the end
- Got rid of an unnecessary call to the workspace to get the changed files - turns out we already had a library function to transform a diff into that list.

### Test plan

Test suite and manual testing.


internal/batches/workspace/workspace.go
```diff
@@ -39,15 +38,12 @@ type Workspace interface {
 	// delete the workspace when Close is called.
 	Close(ctx context.Context) error
 
-	// Changes is called after each step is executed, and should return the
-	// cumulative file changes that have occurred since Prepare was called.
-	Changes(ctx context.Context) (git.Changes, error)
```
eseliger, line 44: We don't actually need this call, nice. Saves a docker invocation.

internal/batches/executor/execution_cache.go
```diff
@@ -121,44 +121,3 @@ func (ExecutionNoOpCache) Set(ctx context.Context, key cache.Keyer, result execu
 func (ExecutionNoOpCache) Get(ctx context.Context, key cache.Keyer) (execution.AfterStepResult, bool, error) {
 	return execution.AfterStepResult{}, false, nil
 }
-
```
eseliger, line 124: we don't need a fully-fledged cache anymore now that we can run steps directly.

cmd/src/batch_validate.go
```diff
@@ -60,9 +60,7 @@ Examples:
 		out := output.NewOutput(flagSet.Output(), output.OutputOpts{Verbose: *verbose})
 		ui := &ui.TUI{Out: out}
 		svc := service.New(&service.Opts{
-			Client:           cfg.apiClient(apiFlags, flagSet.Output()),
-			AllowUnsupported: allowUnsupported,
-			AllowIgnored:     allowIgnored,
+			Client: cfg.apiClient(apiFlags, flagSet.Output()),
```
eseliger, line 63: those flags were specific to a single method on service, so they aren't set here anymore to avoid confusing people thinking they're always mandatory to set (myself mostly)

cmd/src/batch_exec.go
```diff
@@ -31,39 +30,29 @@ const (
 )
 
 type executorModeFlags struct {
```
eseliger, line 32: cache dir and sg version are not required after the refactor here

cmd/src/batch_exec.go
```diff
@@ -182,75 +155,65 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
...
+		// TODO: Should be slightly less than the executor timeout. Can we somehow read that?
+		Timeout:   flags.timeout,
+		TempDir:   tempDir,
+		GlobalEnv: globalEnv,
+		// Temporarily prevent the ability to sending a batch spec with a mount for server-side processing.
+		IsRemote:    isRemote,
+		RepoArchive: &repozip.NoopArchive{},
+		UI:          taskExecUI.StepsExecutionUI(task),
 	}
+	results, err := executor.RunSteps(ctx, opts)
```
eseliger, line 202: Running steps directly, no coordinator nor executor needed here! They're only concurrency manager wrappers that are overhead in SSBC and made it hard to understand which flags are actually required and if we might be hitting sg, hence the deadClient before.

cmd/src/batch_exec.go
```diff
@@ -290,33 +256,13 @@ func convertWorkspace(w batcheslib.WorkspacesExecutionInput) service.RepoWorkspa
...
 		},
-		Path:               w.Path,
-		Steps:              w.Steps,
-		OnlyFetchWorkspace: w.OnlyFetchWorkspace,
+		Path:                  w.Path,
+		Steps:                 w.Steps,
+		OnlyFetchWorkspace:    w.OnlyFetchWorkspace,
+		BatchChangeAttributes: &w.BatchChangeAttributes,
+		CachedStepResultFound: w.CachedStepResultFound,
+		CachedStepResult:      w.CachedStepResult,
```
eseliger, line 264: we now construct the task here already, so we don't need to read the file from disk again and might risk cache key mismatches.

internal/batches/executor/run_steps.go
```diff
@@ -25,47 +27,61 @@ import (
...
-	globalEnv []string
+	GlobalEnv []string
 }
 
-func runSteps(ctx context.Context, opts *runStepsOpts) (stepResults []execution.AfterStepResult, err error) {
-	opts.ui.ArchiveDownloadStarted()
-	err = opts.task.RepoArchive.Ensure(ctx)
-	opts.ui.ArchiveDownloadFinished(err)
+func RunSteps(ctx context.Context, opts *RunStepsOpts) (stepResults []execution.AfterStepResult, err error) {
+	// Set up our timeout.
```
eseliger, line 57: Moved to here from the executor so timeouts work when using runsteps directly, too.

LawnGnome (approved):
This is a big refactor, but it's one that feels overdue looking at it in totality — the diffstat alone kind of hints at that, but I think it had also been quite a while since any of us had really stepped back to thing about how the pieces of running steps really fit together, and there's obviously a cleaner pattern here for `src batch exec` that doesn't make our lives any harder for client side Batch Changes. Great job.

One historical note on the `Changes` method in the workspace interface — my memory is hazy on this, but I think the genesis of that was the mythical "marionette" workspace that [never made it past a draft PR](https://github.com/sourcegraph/src-cli/pull/418), where we'd have a long running container essentially orchestrating the Git workspace as a way of bypassing performance issues around calculating large diffs. Since a bunch of logic would then move out of the `src` process, the idea was to keep it _super_ agnostic to what a diff really looked like until we needed a real diff.

We never went ahead with that, and in practice, I don't think it's come up as a problem since. (And Docker for Mac is theoretically better at some of the filesystem stuff now, so chances are a user with big diff problems might actually be better trying `bind` mode now.)

So, basically, +1 on removing it and simplifying the whole shebang.

PR #801, opened 2022-07-13
Use new API to resolve workspaces server-side

eseliger:
Makes use of the newly introduced endpoint in src-cli. This is usually faster, and it means we can eventually have just one code path again. 

I think this is good as-is and a net-improvement that we should get in, but before we can remove the old code path, we should have a conversation about the `src batch repos` command:

- We cannot render the same output with the new resolver
- Because the current output is incorrect. It doesn't consider workspaces, nor ignored repos or any potential overwrites by a later rule in the `on:` section
Hence I propose that we shift this to run a proper workspace resolution and just print that outcome, like we do in the UI.

Closes https://github.com/sourcegraph/sourcegraph/issues/36261

### Test plan

Validated everything works as expected.


internal/batches/features.go
```diff
@@ -44,6 +45,7 @@ func (ff *FeatureFlags) SetFromVersion(version string) error {
 		{&ff.AllowOptionalPublished, ">= 3.30.0-0", "2021-06-21"},
 		{&ff.ServerSideBatchChanges, ">= 3.37.0-0", "2022-02-08"},
 		{&ff.BitbucketCloud, ">= 3.40.0-0", "2022-04-20"},
+		{&ff.ServerSideWorkspaceResolution, ">= 3.42.0-0", "2022-07-13"},
```
eseliger, line 48: TODO: Adjust date.

internal/batches/service/service.go
```diff
@@ -155,6 +155,110 @@ func (svc *Service) CreateChangesetSpec(ctx context.Context, spec *batcheslib.Ch
...
+		}
+	}
+
+	if unsupported.HasUnsupported() {
+		return workspaces, repos, unsupported
+	}
+
+	if ignored.HasIgnored() {
+		return workspaces, repos, ignored
+	}
```
eseliger, line 257: This just mimics what we currently do here. This deserves a refactor, too, but not in this PR :)

BolajiOlajide (approved):

LawnGnome (approved):
LGTM.

In terms of `src batch repos`: yeah, given that was a pre-workspaces "eh, seems useful" thing, we should rethink that. Maybe we add a new `src batch workspaces` and make `src batch repos` a deprecated command that just does that with a message saying it's deprecated?

courier-new (approved):
Agreed with your proposition and Adam's comment -- we can mark `src batch repos` as deprecated for now while we retain the old code path and have the proper resolution for `src batch workspaces` instead.

PR #807, opened 2022-07-26
search: use chunk matches to render multiline streamed results

rvantonder:
This switches `src search -stream 'query'` to use the chunk match result type. 

This means:

- Accurate result counts for patterns that match across newlines

- Multiline/structural search highlighting is now accurate (previously structural search was highlighting was broken)

### New JSON schema for `src search -stream -json 'query'`

<details>
  <summary><b>new schema</b></summary>

```json
{
  "type": "content",
  "path": ".golangci.yml",
  "repository": "github.com/sourcegraph/sourcegraph",
  "branches": [
    ""
  ],
  "commit": "994c2bc9fd0d5a5baaa6ed1664b3fa7bfa96e4aa",
  "chunkMatches": [
    {
      "content": "    - sheets.go",
      "contentStart": {
        "offset": 3092,
        "line": 98,
        "column": 0
      },
      "ranges": [
        {
          "start": {
            "offset": 3098,
            "line": 98,
            "column": 6
          },
          "end": {
            "offset": 3107,
            "line": 98,
            "column": 15
          }
        }
      ]
    },
    {
      "content": "    - slack.go",
      "contentStart": {
        "offset": 3108,
        "line": 99,
        "column": 0
      },
      "ranges": [
        {
          "start": {
            "offset": 3114,
            "line": 99,
            "column": 6
          },
          "end": {
            "offset": 3122,
            "line": 99,
            "column": 14
          }
        }
      ]
    }
  ]
}
```

</details>

<details>
  <summary><b>old schema</b></summary>

```json
{
  "type": "content",
  "path": ".golangci.yml",
  "repository": "github.com/sourcegraph/sourcegraph",
  "branches": [
    ""
  ],
  "commit": "477f4cf74b6deef0f78ba5ec316fd299bddc621c",
  "lineMatches": [
    {
      "line": "    - sheets.go",
      "lineNumber": 98,
      "offsetAndLengths": [
        [
          6,
          9
        ]
      ]
    },
    {
      "line": "    - slack.go",
      "lineNumber": 99,
      "offsetAndLengths": [
        [
          6,
          8
        ]
      ]
    }
  ]
}
```

</details>

### PR details:

I am making the decision to _remove line matches entirely_ for this option, because otherwise we can never move away from line matches/`preview` to adopt chunk matches as the only source of truth, which we really should do. It is possible that this could break clients/consumers that expect line matches for the `-stream -json` option, and the changelog is updated accordingly. 

Maybe this should be a major version bump, but looks like we keep src-cli releases in tandem with Sourcegraph? Or is this minor enough that we can ship this with a minor `src` release (to what extent have we encouraged using `search -stream`? consumers could in theory remove `-stream` and revert to using GQL to regain "old" behavior). If we don't feel comfortable with removing `LineMatches` in this PR, I would prefer holding out until 4.0 release so I can have the opportunity to piggy back on the "breaking change" version bump.

I think the follow up we want is:

- make `search -stream` the default `search`
- update the GQL search command to be an option like `search -gql` instead (only useful for clients who want to get JSON I imagine).

### Test plan
Updates tests


cmd/src/testdata/TestSearchStream/Text.golden
```diff
@@ -1,8 +1,8 @@
 [38;5;23morg/repo[0m › [38;5;69mpath/to/file[0m[38;5;2m (2 matches)[0m
 [38;5;239m--------------------------------------------------------------------------------
-[0m  [38;5;69m     5[0m[38;5;239m |  [0m[38;5;0m[48;5;11mfoo[0m bar [38;5;0m[48;5;11mfoo[0m
-
-[38;5;2msourcegraph/sourcegraph[0m[38;5;239m ([0m[38;5;23mhttp://127.0.0.1:55128/sourcegraph/sourcegraph[0m[38;5;239m)[0m[38;5;2m (1 match)[0m
+[0m  [38;5;69m     5[0m[38;5;239m |  [0m[38;5;0m[48;5;11mfoo[0m bar [38;5;0m[48;5;11mfoo[0m[38;5;239m
+  ------------------------------------------------------------------------------
```
rvantonder, line 4: There is a minor cosmetic change, that we will put a `----` line below the last result, and is a consequence of keeping the loop logic simple

cmd/src/search_stream.go
```diff
@@ -321,25 +317,44 @@ const streamingTemplate = `
...
+
+		addLineColumn := func(n int) []rune {
+			result = append(result, []rune("  ")...)
+			result = append(result, []rune(ansiColors["search-line-numbers"])...)
+			result = append(result, []rune(fmt.Sprintf("% 6d", match.ContentStart.Line+1+n))...)
+			result = append(result, []rune(ansiColors["nc"])...)
+			result = append(result, []rune(ansiColors["search-border"])...)
+			result = append(result, []rune(" |  ")...)
+			result = append(result, []rune(ansiColors["nc"])...)
+			return result
```
rvantonder, line 331: this logic preserves the line column formatting, which is based off of seeing `\n`, and this logic is no longer easy to express in the template. on the plus side, the template becomes simpler

cmd/src/search_stream.go
```diff
@@ -321,25 +317,44 @@ const streamingTemplate = `
...
+				result = append(result, []rune(ansiColors["search-match"])...)
+				continue
+			}
+			for _, rr := range match.Ranges {
+				if rr.Start.Offset-match.ContentStart.Offset+1 == offset {
+					result = append(result, []rune(ansiColors["search-match"])...)
+				} else if rr.End.Offset-match.ContentStart.Offset+1 == offset {
+					result = append(result, []rune(ansiColors["nc"])...)
+				}
+			}
```
rvantonder, line 352: sticking to the style of highlighting logic we currently, which is to check whether the current rune is within any highlightable range. slicing the highlighted versus non-highlighted parts out of content felt not fun and I didn't want to spend that much time to this effort.

keegancsmith:
If we version src-cli in tandem with sourcegraph this should be safe. If not, I worry that a customer is on an older version of sourcegraph and now src search -stream breaks.

Either way we need to document this loudly in the changelog. I know of customers using the json output with streaming.

> make search -stream the default search
> update the GQL search command to be an option like search -gql instead (only useful for clients who want to get JSON I imagine).

Sounds good to me! Additionally can we update the output of this so its less janky? I think you have done some of that, but man I really wish it didn't automatically have color output and the use of a pager. When using src-cli I am nearly always wanting to consume it via a tool and not as a human.

cmd/src/search_stream.go
camdencheek: Does this work with overlapping ranges? If I'm reading it correctly, we set a color if we find a range that starts at this rune, we unset the color if we find a range that ends at this rune. For an overlapping range, the end of the first range will cancel highlighted color early. I think this could be solved with just a "active ranges" counter. We practically only will have overlapping ranges in very few situations, but it's not impossible per the API definitions.

camdencheek (approved):

rvantonder:
@camdencheek nice catch, I remembered the part about overlapping ranges but naively thought it would unset and just re-enable highlighting. A test case proved otherwise. I implemented a test and the counter idea to fix this

@keegancsmith we are set with in-tandem compatibility. Is bold markdown enough for "loudly" communicating this or should we go harder with `###` headings?

> Additionally can we update the output of this so its less janky? I think you have done some of that, but man I really wish it didn't automatically have color output and the use of a pager. 

Unfortunately I have developed an allergy for drive-by requests and will have to decline 😛 I really just want to make sure we benefit from @camdencheek's hard work while it's fresh, and expedite our ability to get rid of `Preview`

rvantonder:
Alright I'm going to go ahead with merging this. I made the changelog entry louder. I will triage client/customer usage issues coming from this change (so you know who to call). I suspect this is not going to be a high impact change. We have really have to move forward at some point and break away from broken stuff for our own sanity.

rvantonder:
Also I checked the docs to see if this change should update anything there, and instead see we're still not guaranteeing backward compatibility anyway.

> NOTE: The Stream API is still evolving. Although parts of it can be considered stable, we don’t guarantee backward compatibility just yet. This means it is possible that fields are added, removed, or renamed. All backward incompatible changes to the event stream format will be documented in the [CHANGELOG](https://github.com/sourcegraph/sourcegraph/blob/main/CHANGELOG.md).

https://docs.sourcegraph.com/api/stream_api

PR #813, opened 2022-08-10
batches: fix Docker Desktop for Linux handling

### LINKED ISSUE #754: batches: Docker Desktop for Linux: Invalid mount config for type "bind"
[Docker Desktop for Linux](https://docs.docker.com/desktop/linux/) is out and looks like it has similar problems to #405.

Somehow, the underlying dockerd cannot find the `-tmp` directory used to unpack repositories.

This happens whether `-workspace bind` or `-workspace volume`. Running in `-v` mode gives the following output:

``` console
sfllaw@hostname:~$ src batch preview -v -workspace volume -f batch-change.yaml 
✅ Parsing batch spec
✅ Resolving namespace
✅ Preparing container images  █████████████████████████████████████████████████
🚧 Workspace creator: volume
✅ Set workspace type
✅ Resolved 5 repositories
✅ Found 5 workspaces with steps to execute
✅ Found 4 cached changeset specs; 1 task needs to be executed
✅ Executing... (1/1, 1 errored)  ██████████████████████████████████████████████
│                                                                               
└── github.com/sfllaw/repo  creating workspace: preparing local g...  6s

❌ Error:
   github.com/sfllaw/repo:
   creating workspace: preparing local git repo: preparing workspace: Docker output:
   docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /tmp/src-run-1024304234.
   See 'docker run --help'.: exit status 125
   Log: /tmp/changeset-github.com-sfllaw-repo-service-2ee289175889fff0ee96efc10192185f8ed8c861.3212197002.log
```

I have since reverted to the system dockerd, which has no such problem.

LawnGnome:
If the default temporary directory is in use _and_ Docker Desktop for Linux is the default Docker context, this changes the temporary directory to be one under `~/.cache/sourcegraph/batch-tmp` instead.

Fixes #754.

### Test plan

Tested manually in both workspace modes.

### DIFF
--- a/internal/batches/docker/info.go
+++ b/internal/batches/docker/info.go
@@ -10,6 +10,31 @@ import (
 	"github.com/sourcegraph/src-cli/internal/exec"
 )
 
+// CurrentContext returns the name of the current Docker context (not to be
+// confused with a Go context).
+func CurrentContext(ctx context.Context) (string, error) {
+	dctx, cancel, err := withFastCommandContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	defer cancel()
+
+	args := []string{"context", "inspect", "--format", "{{ .Name }}"}
+	out, err := exec.CommandContext(dctx, "docker", args...).CombinedOutput()
+	if errors.IsDeadlineExceeded(err) || errors.IsDeadlineExceeded(dctx.Err()) {
+		return "", newFastCommandTimeoutError(dctx, args...)
+	} else if err != nil {
+		return "", err
+	}
+
+	name := string(bytes.TrimSpace(out))
+	if name == "" {
+		return "", errors.New("no context returned from Docker")
+	}
+
+	return name, nil
+}
+
 // NCPU returns the number of CPU cores available to Docker.
 func NCPU(ctx context.Context) (int, error) {
 	dctx, cancel, err := withFastCommandContext(ctx)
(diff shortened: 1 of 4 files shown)


eseliger (approved):
Nice!

Piszmog (approved):

internal/batches/docker/info.go
```diff
@@ -10,6 +10,31 @@ import (
...
+// CurrentContext returns the name of the current Docker context (not to be
+// confused with a Go context).
+func CurrentContext(ctx context.Context) (string, error) {
+	dctx, cancel, err := withFastCommandContext(ctx)
+	if err != nil {
+		return "", err
+	}
+	defer cancel()
+
+	args := []string{"context", "inspect", "--format", "{{ .Name }}"}
```
courier-new, line 22: Wow, it makes me happy that this is so straightforward to get from Docker! 😄

courier-new (approved):
I'm so glad you have an actual linux box to test this on. Thanks for all the comments for future context, too!

internal/batches/docker/info.go
LawnGnome: I suspected the hardest part of this ticket was going to be detecting this case, so this was a very pleasant surprise.

PR #816, opened 2022-08-15
batches: Remote file mounting

Piszmog:

### DIFF
--- a/cmd/src/batch_exec.go
+++ b/cmd/src/batch_exec.go
@@ -224,7 +224,7 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags,
 
 	// Write all step cache results for all results.
 	for _, stepRes := range results {
-		cacheKey := task.CacheKey(globalEnv, mountDir, true, stepRes.StepIndex)
+		cacheKey := task.CacheKey(globalEnv, mountDir, stepRes.StepIndex)
 		k, err := cacheKey.Key()
 		if err != nil {
 			return errors.Wrap(err, "calculating step cache key")
--- a/cmd/src/batch_exec.go
+++ b/cmd/src/batch_exec.go
@@ -10,16 +10,18 @@ import (
 	"path/filepath"
 	"time"
 
-	"github.com/sourcegraph/sourcegraph/lib/errors"
-
+	"github.com/sourcegraph/src-cli/internal/api"
 	"github.com/sourcegraph/src-cli/internal/batches/docker"
-	"github.com/sourcegraph/src-cli/internal/batches/executor"
-	"github.com/sourcegraph/src-cli/internal/batches/graphql"
 	"github.com/sourcegraph/src-cli/internal/batches/log"
 	"github.com/sourcegraph/src-cli/internal/batches/repozip"
+	"github.com/sourcegraph/src-cli/internal/batches/workspace"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+
+	"github.com/sourcegraph/src-cli/internal/batches/executor"
+	"github.com/sourcegraph/src-cli/internal/batches/graphql"
 	"github.com/sourcegraph/src-cli/internal/batches/service"
 	"github.com/sourcegraph/src-cli/internal/batches/ui"
-	"github.com/sourcegraph/src-cli/internal/batches/workspace"
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
 
 	batcheslib "github.com/sourcegraph/sourcegraph/lib/batches"
@@ -30,18 +32,23 @@ const (
 )
 
 type executorModeFlags struct {
-	timeout time.Duration
-	file    string
-	tempDir string
-	repoDir string
+	timeout  time.Duration
+	file     string
+	tempDir  string
+	repoDir  string
+	mountDir string
+	api      *api.Flags
 }
 
 func newExecutorModeFlags(flagSet *flag.FlagSet) (f *executorModeFlags) {
-	f = &executorModeFlags{}
+	f = &executorModeFlags{
+		api: api.NewFlags(flagSet),
+	}
 	flagSet.DurationVar(&f.timeout, "timeout", 60*time.Minute, "The maximum duration a single batch spec step can take.")
 	flagSet.StringVar(&f.file, "f", "", "The workspace execution input file to read.")
 	flagSet.StringVar(&f.tempDir, "tmp", "", "Directory for storing temporary data.")
 	flagSet.StringVar(&f.repoDir, "repo", "", "Path of the checked out repo on disk.")
+	flagSet.StringVar(&f.mountDir, "mount", "", "Path of the files to mount on disk.")
 
 	return f
 }
@@ -56,6 +63,9 @@ func validateExecutorModeFlags(f *executorModeFlags) error {
 	if f.repoDir == "" {
 		return errors.New("repoDir parameter missing")
 	}
+	if f.mountDir == "" {
+		return errors.New("mountDir parameter missing")
+	}
 
 	return nil
 }
@@ -69,7 +79,7 @@ github.com/sourcegraph/sourcegraph/lib/batches.
 
 Usage:
 
-    src batch exec -f FILE -repo DIR [command options]
+    src batch exec -f FILE -repo DIR -mount DIR [command options]
 
 Examples:
 
@@ -96,9 +106,13 @@ Examples:
 		ctx, cancel := contextCancelOnInterrupt(context.Background())
 		defer cancel()
 
-		err := executeBatchSpecInWorkspaces(ctx, flags)
+		svc := service.New(&service.Opts{
+			Client: cfg.apiClient(flags.api, flagSet.Output()),
+		})
+
+		err := executeBatchSpecInWorkspaces(ctx, flags, svc)
 		if err != nil {
-			return cmderrors.ExitCode(1, nil)
+			return cmderrors.ExitCode(1, err)
 		}
 
 		return nil
@@ -115,7 +129,7 @@ Examples:
 	})
 }
 
-func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags) (err error) {
+func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags, svc *service.Service) (err error) {
 	ui := &ui.JSONLines{}
 
 	// Ensure the temp dir exists.
@@ -135,6 +149,15 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
 	repoDir := flags.repoDir
 	if !filepath.IsAbs(repoDir) {
 		repoDir, err = filepath.Abs(repoDir)
+		if err != nil {
+			return errors.Wrap(err, "getting absolute path for repo dir")
+		}
+	}
+
+	// Grab the absolute path to the mount contents.
+	mountDir := flags.mountDir
+	if !filepath.IsAbs(mountDir) {
+		repoDir, err = filepath.Abs(mountDir)
 		if err != nil {
 			return errors.Wrap(err, "getting absolute path for repo dir")
 		}
@@ -177,8 +200,7 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
 	ui.PreparingContainerImagesSuccess()
 
 	// Empty for now until we support secrets or env var settings in SSBC.
-	globalEnv := []string{}
-	isRemote := true
+	var globalEnv []string
 
 	// Set up the execution UI.
 	taskExecUI := ui.ExecutingTasks(false, 1)
@@ -191,11 +213,10 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
 		EnsureImage: imageCache.Ensure,
 		Task:        task,
 		// TODO: Should be slightly less than the executor timeout. Can we somehow read that?
-		Timeout:   flags.timeout,
-		TempDir:   tempDir,
-		GlobalEnv: globalEnv,
-		// Temporarily prevent the ability to sending a batch spec with a mount for server-side processing.
-		IsRemote:    isRemote,
+		Timeout:     flags.timeout,
+		TempDir:     tempDir,
+		MountDir:    mountDir,
+		GlobalEnv:   globalEnv,
 		RepoArchive: &repozip.NoopArchive{},
 		UI:          taskExecUI.StepsExecutionUI(task),
 	}
@@ -203,7 +224,7 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
 
 	// Write all step cache results for all results.
 	for _, stepRes := range results {
-		cacheKey := task.CacheKey(globalEnv, isRemote, stepRes.StepIndex)
+		cacheKey := task.CacheKey(globalEnv, mountDir, true, stepRes.StepIndex)
 		k, err := cacheKey.Key()
 		if err != nil {
 			return errors.Wrap(err, "calculating step cache key")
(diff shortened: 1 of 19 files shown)


go.mod
```diff
@@ -116,3 +117,5 @@ require (
 
 // See: https://github.com/ghodss/yaml/pull/65
 replace github.com/ghodss/yaml => github.com/sourcegraph/yaml v1.0.1-0.20200714132230-56936252f152
+
+replace github.com/sourcegraph/sourcegraph/lib v0.0.0-20220715132627-cb906941e8b4 => ../sourcegraph/lib
```
Piszmog, line 121: This will go away when `sourcegraph` PR is merged. This is here for my local development.

internal/batches/service/remote.go
```diff
@@ -98,13 +113,122 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+		"batchChange":      batchChange,
 	}).Do(ctx, &resp); err != nil || !ok {
 		return "", err
 	}
 
 	return resp.CreateBatchSpecFromRaw.ID, nil
 }
 
+// UploadMounts uploads file mounts to the server.
+func (svc *Service) UploadMounts(workingDir string, batchSpecID string, steps []batches.Step) error {
```
eseliger, line 125: when we land here, the existence of all files is already validated, correct?

internal/batches/service/remote.go
```diff
@@ -98,13 +113,122 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+		for _, mount := range step.Mount {
+			total, err := handlePath(w, workingDir, mount.Path, count)
+			if err != nil {
+				return err
+			}
+			count += total
+		}
+	}
+
+	if err := w.WriteField("count", strconv.Itoa(count)); err != nil {
```
eseliger, line 144: Can't count be derived from the payload? What is is required for here?

internal/batches/service/remote.go
```diff
@@ -98,13 +113,122 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+		return err
+	}
+
+	// Honestly, the most import thing to do. This adds the closing boundary to the request.
+	if err := w.Close(); err != nil {
+		return err
+	}
+
+	request, err := svc.client.NewHTTPRequest(context.Background(), http.MethodPost, fmt.Sprintf(".api/batches/mount/%s", batchSpecID), body)
```
eseliger, line 153: I think this should take a non-background context so we can handle cancellation right

internal/batches/service/remote.go
```diff
@@ -98,13 +113,122 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+	if err != nil {
+		return err
+	}
+	request.Header.Add("Content-Type", w.FormDataContentType())
+
+	resp, err := svc.client.Do(request)
+	if err != nil {
+		return err
+	}
+	if resp.StatusCode != http.StatusOK {
```
eseliger, line 163: should all 2xx codes be considered a success?

internal/batches/service/remote.go
```diff
@@ -98,13 +113,122 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
 	return resp.CreateBatchSpecFromRaw.ID, nil
 }
 
+// UploadMounts uploads file mounts to the server.
+func (svc *Service) UploadMounts(workingDir string, batchSpecID string, steps []batches.Step) error {
+	if err := svc.areServerSideBatchChangesSupported(); err != nil {
+		return err
+	}
+
+	body := &bytes.Buffer{}
```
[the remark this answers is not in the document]
eseliger: Also, this is a single request for all uploads, in one multipart body, correct? 

If so, this will be problematic for larger workspace uploads. Batch changes and code intel ran into that and for batch changes that meant uploading each changeset spec individually and then stitching the batch specs together once finished (because diffs can become quite huge we had to learn), and for code intel they implemented chunked uploading which in the backend stitches together larger uploads since lsif uploads can be large singular blobs (like a go binary, so not too far off of our use-case). They also have some mechanism to expire partial uploads at some stage once they are too old, to make sure we don't hoard partial data forever. Link to uploader code: https://sourcegraph.com/github.com/sourcegraph/sourcegraph-incremental-indexing-fork/-/blob/lib/codeintel/upload/upload.go?L18

internal/batches/executor/run_steps.go
```diff
@@ -38,16 +39,15 @@ type RunStepsOpts struct {
 	Task *Task
 	// TempDir points to where temporary files of the execution should live at.
 	TempDir string
+	// MountDir points to where files that should be mounted live at.
+	MountDir string
```
eseliger, line 43: Maybe `WorkspaceDir` describes this better? MountDir makes me think we mount a specific dir.

eseliger (commented):
Did an initial pass (since this is draft, not sure how deep I should go) and added thoughts on things that stood out to me as I went over the code here.
The biggest question to me is chunked uploading and memory usage while constructing the upload body.

cmd/src/batch_exec.go
```diff
@@ -115,7 +129,7 @@ Examples:
 	})
 }
 
-func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags) (err error) {
+func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags, svc *service.Service) (err error) {
```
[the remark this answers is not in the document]
Piszmog: I thought it was being left out so I decided to include it (jk)... I need to remove it. It's an artifact from an initial implementation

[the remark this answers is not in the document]
eseliger: there just wasn't great service in this file 😠

internal/batches/service/remote.go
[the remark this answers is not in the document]
Piszmog: I wrote this function to be generic (so I could upload x files). I want to limit the size of the files that can be upload and the number of files per request (se we are not trying to upload 100s of files in a single request).

One of the ideas I have been wrestling with when I did it this way is bulk upload files vs concurrent upload (1 file per request and send in parallel).

Work still needs to be done on performance... So this probably will change

internal/batches/ui/tui.go
```diff
@@ -239,6 +240,14 @@ func (ui *TUI) SendingBatchSpecSuccess() {
 	batchCompletePending(ui.pending, "Sending batch spec")
 }
 
+func (ui *TUI) UploadingMounts() {
+	ui.pending = batchCreatePending(ui.Out, "Uploading mounts")
```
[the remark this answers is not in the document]
eseliger: yeah ... let's brainstorm a bit in a team sync with 3 spare minutes at the end 😬

internal/batches/service/remote.go
Piszmog: That's correct

internal/batches/service/remote.go
Piszmog: This is the number of files being uploaded. It could be derived if I look at the `file_%d` field. But this would require to parse this (and potentially look at every field since order isn't guaranteed).

Piszmog: Yea, with this current implementation. With this bulk upload, it needs to know how many files are in the request (so it can lookup the files from the payload)

cmd/src/batch_exec.go
```diff
@@ -140,6 +147,15 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
...
 
+	// Grab the absolute path to the workspace files contents.
+	workspaceFilesDir := flags.workspaceFilesDir
+	if !filepath.IsAbs(workspaceFilesDir) {
+		workspaceFilesDir, err = filepath.Abs(workspaceFilesDir)
+		if err != nil {
+			return errors.Wrap(err, "getting absolute path for workspace files dir")
+		}
+	}
+
```
LawnGnome, line 158: I don't know that we need to do the `IsAbs` here; presumably `filepath.Abs(path)` on a path that's already absolute will just return the same thing anyway.

internal/batches/service/remote.go
```diff
@@ -96,6 +102,111 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+	w := multipart.NewWriter(pipeWriter)
+
+	// Write in a separate goroutine to properly chunk the file content. Writing to the pipe lets us not have
+	// to put the whole file in memory.
+	go func() {
+		defer pipeWriter.Close()
+		defer w.Close()
+
+		if err := createFormFile(w, workingDir, mountPath); err != nil {
+			// TODO
```
LawnGnome, line 154: What shall we do?

internal/batches/service/remote.go
```diff
@@ -96,6 +102,111 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+		if err != nil {
+			return err
+		}
+		return errors.New(string(p))
+	}
+	return nil
+}
+
+func createFormFile(w *multipart.Writer, workingDir string, mountPath string) error {
+	// TODO: limit file size
```
LawnGnome, line 180: I assume we're going to address this before merge.

LawnGnome (commented):
Looks like there's still a couple of bits to be finished off, but what's here LGTM.

eseliger:
> Looks like there's still a couple of bits to be finished off, but what's here LGTM.

Let me know when it's ready for review, I will take a look asap then :)

internal/batches/service/remote.go
Piszmog: Handling errors in goroutines is one of the biggest gripes I have...

Piszmog: Turns out this was easy with the `pipeWriter` - can just do `pipeWriter.CloseWithError`

Piszmog:
@eseliger should be good to look now. I think I caught all of my TODOs

internal/batches/service/remote.go
```diff
@@ -96,6 +102,121 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+// UploadBatchSpecWorkspaceFile uploads workspace files to the server.
+func (svc *Service) UploadBatchSpecWorkspaceFile(workingDir string, batchSpecID string, steps []batches.Step) error {
+	for _, step := range steps {
+		for _, mount := range step.Mount {
+			if err := svc.handleWorkspaceFile(workingDir, mount.Path, batchSpecID); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
```
eseliger, line 114: I think this has to deduplicate in the following case where the same thing is mounted to multiple steps:

```yaml
steps:
  - run: asdf
    mounts: [{tmp:/work/tmp}]
  - run: asdf 2
    mounts: [{tmp:/work/tmp}]
```

internal/batches/service/remote.go
```diff
@@ -96,6 +102,121 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+	go func() {
+		defer pipeWriter.Close()
+		defer multipartWriter.Close()
+
+		if err := createFormFile(multipartWriter, workingDir, mountPath); err != nil {
+			pipeWriter.CloseWithError(err)
+		}
+	}()
+
+	request, err := svc.client.NewHTTPRequest(context.Background(), http.MethodPost, fmt.Sprintf(".api/files/batch-changes/%s", batchSpecID), pipeReader)
```
eseliger, line 158: Should we hand down the proper context to here, so that we can better react to cancelation?

internal/batches/service/remote.go
```diff
@@ -96,6 +102,121 @@ func (svc *Service) CreateBatchSpecFromRaw(
...
+	if err != nil {
+		return err
+	}
+	request.Header.Set("Content-Type", multipartWriter.FormDataContentType())
+
+	resp, err := svc.client.Do(request)
+	if err != nil {
+		// Errors passed to pipeWriter.CloseWithError come through here.
+		return err
```
eseliger, line 167: that comment means that the file too large error would be handed up here, correct? If so, all perfect

internal/batches/service/remote_test.go
```diff
@@ -191,3 +201,407 @@ func TestService_CreateBatchSpecFromRaw(t *testing.T) {
...
+}
+
+type neverEnding byte
+
+func (b neverEnding) Read(p []byte) (n int, err error) {
+	for i := range p {
+		p[i] = byte(b)
+	}
+	return len(p), nil
+}
```
eseliger, line 607: wow, very cool

cmd/src/batch_exec.go
```diff
@@ -30,18 +32,23 @@ const (
...
+	repoDir           string
+	workspaceFilesDir string
+	api               *api.Flags
 }
 
 func newExecutorModeFlags(flagSet *flag.FlagSet) (f *executorModeFlags) {
-	f = &executorModeFlags{}
+	f = &executorModeFlags{
+		api: api.NewFlags(flagSet),
+	}
```
eseliger, line 46: I don't think we need this, the executor mode should never call SG APIs

cmd/src/batch_remote.go
```diff
@@ -100,6 +100,16 @@ Examples:
...
 
+		ui.UploadingWorkspaceFiles()
+		dir, err := getBatchSpecDirectory(file)
+		if err != nil {
+			return errors.Wrap(err, "batch spec path")
+		}
+		if err = svc.UploadBatchSpecWorkspaceFile(dir, batchSpecID, spec.Steps); err != nil {
+			return err
+		}
+		ui.UploadingWorkspaceFilesSuccess()
```
eseliger, line 111: would be nice to only log the messages here when there are actually any file matches, otherwise it might be confusing "Oh no what is it uploading??"

internal/batches/executor/executor.go
```diff
@@ -168,15 +169,15 @@ func (x *executor) do(ctx context.Context, task *Task, ui TaskExecutionUI) (err
...
-		RepoArchive: repoArchive,
+		Task:              task,
+		Logger:            l,
+		WC:                x.opts.Creator,
+		EnsureImage:       x.opts.EnsureImage,
+		TempDir:           x.opts.TempDir,
+		GlobalEnv:         x.opts.GlobalEnv,
+		Timeout:           x.opts.Timeout,
+		RepoArchive:       repoArchive,
+		WorkspaceFilesDir: x.opts.WorkingDirectory,
```
eseliger, line 180: In this file, there are two new fields on two structs, but for the same thing. Should they be named the same then, to make it easier to trace what is coming from where?

eseliger (commented):
Nice, looks good!

internal/batches/service/remote.go
Piszmog: Yep! I also have a test for this (and did functional testing as well to confirm)

cmd/src/batch_exec.go
```diff
@@ -30,10 +32,12 @@ const (
...
-	timeout time.Duration
-	file    string
-	tempDir string
-	repoDir string
+	timeout           time.Duration
+	file              string
+	tempDir           string
+	repoDir           string
+	workspaceFilesDir string
+	api               *api.Flags
```
eseliger, line 40: I think this is unused

eseliger (approved):
Let's go!!!11

cmd/src/batch_exec.go
Piszmog: Good catch. This was left over code when I was just trying to get things to work

internal/batches/service/remote_windows_test.go
```diff
@@ -0,0 +1,76 @@
...
+	mockclient "github.com/sourcegraph/src-cli/internal/api/mock"
+	"github.com/sourcegraph/src-cli/internal/batches/service"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/mock"
+	"github.com/stretchr/testify/require"
+
+	"github.com/sourcegraph/sourcegraph/lib/batches"
+)
+
+func TestService_UploadBatchSpecWorkspaceFiles_Windows_Path(t *testing.T) {
```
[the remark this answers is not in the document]
Piszmog: lol, I had a dream about windows paths. This is the result

PR #817, opened 2022-08-15
batches: Make sure no corrupted repozips are used

eseliger:
When the download of a zip was aborted, it could've left behind a corrupted archive zip. This fixes it by first storing it to a temporary location and then creating the actual file atomically by doing os.Rename.

Closes https://github.com/sourcegraph/sourcegraph/issues/39743

### Test plan

Verified things still work as expected.

### DIFF
--- a/internal/batches/repozip/fetcher.go
+++ b/internal/batches/repozip/fetcher.go
@@ -336,7 +336,7 @@ func fetchRepositoryFile(ctx context.Context, client HTTPClient, repo RepoRevisi
 		return false, fmt.Errorf("unable to fetch archive (HTTP %d from %s)", resp.StatusCode, req.URL.String())
 	}
 
-	f, err := os.CreateTemp(path.Dir(dest), fmt.Sprintf("%s-*.tmp", path.Base(dest)))
+	f, err := os.CreateTemp(filepath.Dir(dest), fmt.Sprintf("%s-*.tmp", filepath.Base(dest)))
 	if err != nil {
 		return false, err
 	}
--- a/internal/batches/repozip/fetcher.go
+++ b/internal/batches/repozip/fetcher.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"io"
-	"math/rand"
 	"net/http"
 	"os"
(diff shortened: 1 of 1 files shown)


internal/batches/repozip/fetcher.go
```diff
@@ -336,7 +337,10 @@ func fetchRepositoryFile(ctx context.Context, client HTTPClient, repo RepoRevisi
 		return false, fmt.Errorf("unable to fetch archive (HTTP %d from %s)", resp.StatusCode, req.URL.String())
 	}
 
-	f, err := os.Create(dest)
+	tmpFileDir := fmt.Sprintf("%s.%d.tmp", dest, rand.Int())
+	// Ensure the file doesn't exist.
+	_ = os.Remove(tmpFileDir)
+	f, err := os.Create(tmpFileDir)
```
LawnGnome, line 343: We should use [`os.CreateTemp`](https://pkg.go.dev/os#CreateTemp) here — it does essentially the same thing, but in a slightly safer way. (We also need to ensure it creates the file in the same directory that it will end up in, as I think you already know, since rename operations usually can't cross filesystem boundaries.)

BolajiOlajide (approved):

internal/batches/repozip/fetcher.go
eseliger: how about 8c04c68a196fe7b049cb6ff08406ed7378b92bff

LawnGnome: I think we'll need a `defer func(path string) { _ = os.Remove(path) }(f.Name())` after the `CreateTemp` call, since there's a possible return path before the `Rename`, but otherwise LGTM.

eseliger: Makes sense. I also made sure we close the file before we rename it, although on macOS at least that doesn't seem to make a difference.

LawnGnome: Yeah, I think that would only matter on Windows, and I'm not even sure there. No harm in being safe, though!

LawnGnome (approved):

PR #819, opened 2022-08-15
Remove src-cli workspace resolution

eseliger:
This is available as a faster API from the backend now, so no need to keep this duplicative, slower code. This is a 4.0 deprecation execution.

Comes with https://github.com/sourcegraph/sourcegraph/pull/40426.

Closes https://github.com/sourcegraph/sourcegraph/issues/38968.

### Test plan

Verified execution still works.


cmd/src/batch_repositories.go
```diff
@@ -90,31 +91,34 @@ Examples:
 			return err
 		}
 
-		repoCount := 0
-		for _, on := range spec.On {
-			repos, _, err := svc.ResolveRepositoriesOn(ctx, &on)
-			if err != nil {
-				return errors.Wrapf(err, "Resolving %q", on.String())
+		_, repos, err := svc.ResolveWorkspacesForBatchSpec(ctx, spec, allowUnsupported, allowIgnored)
```
eseliger, line 94: As discussed, this doesn't show the individual queries anymore. A follow-up will deprecate this command and create a new workspaces command as a replacement, that has a more elaborate output, and also properly supports workspaces as a native feature.

internal/batches/ui/exec_ui.go
```diff
@@ -22,11 +22,8 @@ type ExecUI interface {
 	DeterminingWorkspaceCreatorType()
 	DeterminingWorkspaceCreatorTypeSuccess(wt workspace.CreatorType)
 
-	ResolvingRepositories()
-	ResolvingRepositoriesDone(repos []*graphql.Repository, unsupported batches.UnsupportedRepoSet, ignored batches.IgnoredRepoSet)
-
 	DeterminingWorkspaces()
-	DeterminingWorkspacesSuccess(num int)
+	DeterminingWorkspacesSuccess(workspacesCount, reposCount int, unsupported batches.UnsupportedRepoSet, ignored batches.IgnoredRepoSet)
```
courier-new, line 26: Nit, but it looks like for `unsupported` and `ignored` we actually also only use them to get the counts with `len`. Could we just pass all 4 as counts?

courier-new (approved):
YAY SO MANY DELETIONS!! 🎉

Piszmog (approved):

internal/batches/ui/exec_ui.go
eseliger: The TUI prints them, so we need to keep it for this one

PR #822, opened 2022-08-16
Remove old feature gates from pre-4.0

eseliger:
Requires a newer Sourcegraph version from https://github.com/sourcegraph/sourcegraph/pull/40446.

This also adds a feature gate for sourcegraph 4.0 to src-cli to warn users when they use a too recent version of src-cli. 

Closes https://github.com/sourcegraph/sourcegraph/issues/38969
Closes https://github.com/sourcegraph/sourcegraph/issues/38967

### Test plan

Test suite and verified manually the 4.0 banner and previous functionality work.


Piszmog (approved):
LGTM

I'm mentally preparing myself for the merge conflict I'm going to have to go thru.

cmd/src/batch_common.go
```diff
@@ -604,3 +583,14 @@ func getBatchParallelism(ctx context.Context, flag int) (int, error) {
...
+
+func validateSourcegraphVersionConstraint(ctx context.Context, svc *service.Service) error {
+	ffs, err := svc.DetermineFeatureFlags(ctx)
+	if err != nil {
+		return err
+	}
+	if ffs.Sourcegraph40 {
+		return nil
+	}
+	return errors.Newf("\n\n * Warning:\n This version of src-cli requires Sourcegraph version 4.0 or newer. If you're not on Sourcegraph 4.0 or newer, please use any 3.x release of src-cli.\n\n")
```
LawnGnome, line 595: While I doubt anyone is _really_ going to language lawyer this into complaining that src-cli 3.2.0 doesn't work with Sourcegraph 3.43, how about:

```suggestion
	return errors.Newf("\n\n * Warning:\n This version of src-cli requires Sourcegraph version 4.0 or newer. If you're not on Sourcegraph 4.0 or newer, please use the 3.x release of src-cli that corresponds to your Sourcegraph version.\n\n")
```

cmd/src/batch_common.go
```diff
@@ -604,3 +583,14 @@ func getBatchParallelism(ctx context.Context, flag int) (int, error) {
 
 	return docker.NCPU(ctx)
 }
+
+func validateSourcegraphVersionConstraint(ctx context.Context, svc *service.Service) error {
```
LawnGnome, line 587: I worry a little that this may be too simplistic as we start adding things in 4.x again, but let's cross that bridge when we get to it.

internal/api/api.go
```diff
@@ -25,16 +26,9 @@ type Client interface {
 	// variables.
 	NewQuery(query string) Request
 
-	// NewRequest creates a GraphQL request.
+	// NewRequest creates a GraphQL request with gzip compression turned on.
```
LawnGnome, line 29: I actually wouldn't change this comment — this can be an implementation detail now.

LawnGnome (approved):
It's beautiful.

courier-new (commented):
DELETE 👏 IT 👏 ALL 👏

PR #825, opened 2022-08-17
Prepare CLI for extension sunsetting

philipp-spiess:
Part of sourcegraph/sourcegraph#39044
Part of sourcegraph/sourcegraph#39043

While auditing the CLI functionality and requirements for the upcoming extensions deprecation, I noticed two areas that are currently not working well:

- The `sg extensions` command doesn't explain the upcoming deprecation yet. I've changed it to show the same text as sourcegraph.com cc @jjinnii @ryankscott @muratsu 
- The `sg extensions list` command did not forward the GraphQL error probably. This happens because the endpoint does not use the normal GraphQL error pattern for errors. [See this for more context](https://sourcegraph.com/github.com/ilscipio/sourcegraph/-/blob/cmd/frontend/registry/api/extension_connection_graphql.go?L192-204&utm_source=VSCode-2.2.7).

All other extension endpoints work as expected and I made sure they forward an error message sent on the backend (I'll follow up with the backend changes to gate out the APIs in a separate commit [lol obviously because this is a different repo haha]).

### Test plan

Screenshot 2022-08-17 at 11 45 39
Screenshot 2022-08-17 at 12 59 44

### DIFF
--- a/cmd/src/extensions.go
+++ b/cmd/src/extensions.go
@@ -10,7 +10,8 @@ var extensionsCommands commander
 func init() {
 	usage := `'src extensions' is a tool that manages extensions in the extension registry on a Sourcegraph instance.
 
-EXPERIMENTAL: Extensions are experimental functionality on Sourcegraph and in the 'src' tool.
+DEPRECATED: We're in the process of removing Sourcegraph extensions with our September release.
+            Learn more: https://docs.sourcegraph.com/extensions/deprecation
 
 Usage:
 
--- a/cmd/src/extensions_list.go
+++ b/cmd/src/extensions_list.go
@@ -63,6 +63,7 @@ Examples:
       nodes {
         ...RegistryExtensionFields
       }
+			error
     }
   }
 }` + registryExtensionFragment
@@ -71,6 +72,7 @@ Examples:
 			ExtensionRegistry struct {
 				Extensions struct {
 					Nodes []Extension
+					Error string
 				}
 			}
 		}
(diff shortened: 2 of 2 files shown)


cmd/src/extensions.go
```diff
@@ -10,7 +10,8 @@ var extensionsCommands commander
 func init() {
 	usage := `'src extensions' is a tool that manages extensions in the extension registry on a Sourcegraph instance.
 
-EXPERIMENTAL: Extensions are experimental functionality on Sourcegraph and in the 'src' tool.
+DEPRECATED: We're in the process of removing Sourcegraph extensions with our September release.
+            Learn more: https://docs.sourcegraph.com/extensions/deprecation
```
philipp-spiess, line 14: cc @jjinnii @muratsu @ryankscott

eseliger: Do you want this to go out in a 3.43.x version?

cmd/src/extensions_list.go
```diff
@@ -63,6 +63,7 @@ Examples:
       nodes {
         ...RegistryExtensionFields
       }
+			error
```
eseliger, line 66: the indentation here is a bit off

eseliger (approved):
LGTM!

cmd/src/extensions.go
philipp-spiess: I _think_ we do, since we are also shipping this already on sourcegraph.com but this is a questions for the above people to answer 🙈

eseliger: Ok, just let me know when you have a decision, since I already cut 3.43.0 for the branch cut this evening, but can cut 3.43.1 right after merge.

cmd/src/extensions_list.go
philipp-spiess: Screenshot 2022-08-17 at 13 34 05

ah 🤦

cmd/src/extensions.go
philipp-spiess: Ah, I see. I don't think it's super urgent to get this out with 3.43 (the more important messaging is going out on the web app anyway), so I don't think we need to cut a minor release for that 🙂

PR #826, opened 2022-08-18
Add `src users clean` to src-cli

DaedalusG:
This draft PR tracks work on registering a new `src users` subcommand `src users clean` which allows admins to remove users in bulk given some conditions like admin status and days since last active use registered.
https://github.com/sourcegraph/sourcegraph/issues/40102

The command is still in development at the time of publishing this draft, it seems near MVP though

Two Additional nice to have features would be:
- output json file containing users removed and some data on the users (for use to send emails to removed users mostly)
- open an editor to compose general email to be sent to removed users and send. Functionality behind `-send-email` flag

_Please check carefully, I'm new at this!_
### Test plan
Manually test each command option against `cse-aws` and `cse-k8s`

Open to additional testing

### DIFF
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -58,7 +58,7 @@ fragment UserFields on User {
     }
     emails {
         email
-	verified
+		verified
     }
     usageStatistics {
         lastActiveTime
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -60,6 +60,10 @@ fragment UserFields on User {
         email
         verified
     }
+	usageStatistics {
+		lastActiveTime
+		lastActiveCodeHostIntegrationTime
+	}
     url
 }
 `
@@ -72,11 +76,17 @@ type User struct {
 	Organizations struct {
 		Nodes []Org
 	}
-	Emails []UserEmail
-	URL    string
+	Emails          []UserEmail
+	UsageStatistics UserUsageStatistics
+	URL             string
 }
 
 type UserEmail struct {
 	Email    string
 	Verified bool
+}
+
+type UserUsageStatistics struct {
+	LastActiveTime                    string
+	LastActiveCodeHostIntegrationTime string
 }
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -58,12 +58,12 @@ fragment UserFields on User {
     }
     emails {
         email
-		verified
+	verified
     }
-	usageStatistics {
-		lastActiveTime
-		lastActiveCodeHostIntegrationTime
-	}
+    usageStatistics {
+        lastActiveTime
+        lastActiveCodeHostIntegrationTime
+    }
     url
 }
 `
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -20,6 +20,7 @@ The commands are:
 	get        gets a user
 	create     creates a user account
 	delete     deletes a user account
+	clean      deletes inactive users
 	tag        add/remove a tag on a user
 
 Use "src users [command] -h" for more information about a command.
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -58,7 +58,7 @@ fragment UserFields on User {
     }
     emails {
         email
-        verified
+		verified
     }
 	usageStatistics {
 		lastActiveTime
--- a/cmd/src/users_clean.go
+++ b/cmd/src/users_clean.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"reflect"
 	"time"
 
 	"github.com/sourcegraph/src-cli/internal/api"
@@ -25,10 +24,11 @@ Examples:
 		fmt.Println(usage)
 	}
 	var (
-		daysToDelete = flagSet.Int("d", 365, "Returns the first n users from the list. (use -1 for unlimited)")
-		noAdmin      = flagSet.Bool("no-admin", false, "Omit admin accounts from cleanup")
-		toEmail      = flagSet.Bool("email", false, "send removed users an email")
-		apiFlags     = api.NewFlags(flagSet)
+		daysToDelete       = flagSet.Int("d", 365, "Returns the first n users from the list. (use -1 for unlimited)")
+		noAdmin            = flagSet.Bool("no-admin", false, "Omit admin accounts from cleanup")
+		toEmail            = flagSet.Bool("email", false, "send removed users an email")
+		removeNoLastActive = flagSet.Bool("removeNeverActive", false, "removes users with null lastActive value")
+		apiFlags           = api.NewFlags(flagSet)
 	)
 
 	handler := func(args []string) error {
@@ -49,30 +49,11 @@ Examples:
 query Users($first: Int, $query: String) {
 	users(first: $first, query: $query) {
 		nodes {
-			id
-			username
-			displayName
-			siteAdmin
-			organizations {
-				nodes {
-					id
-					name
-					displayName
-				}
-			}
-			emails {
-				email
-				verified
-			}
-			usageStatistics {
-				lastActiveTime
-				lastActiveCodeHostIntegrationTime
-			}
-			url
+			...UserFields
 		}
 	}
 }
-`
+` + userFragment
 
 		var result struct {
 			Users struct {
@@ -86,23 +67,32 @@ query Users($first: Int, $query: String) {
 
 		usersToDelete := make([]User, 0)
 		for _, user := range result.Users.Nodes {
-			daysSinceLastUse, err := computeDaysSinceLastUse(user)
+			daysSinceLastUse, wasLastActive, err := computeDaysSinceLastUse(user)
 			if err != nil {
 				return err
 			}
-			if daysSinceLastUse >= *daysToDelete {
-				usersToDelete = append(usersToDelete, user)
-				fmt.Printf("\nAdding %s to remove list: %d days since last active, remove after %d days inactive\n", user.Username, daysSinceLastUse, *daysToDelete)
+			if !wasLastActive && !*removeNoLastActive {
+				continue
 			}
+			if *noAdmin && user.SiteAdmin {
+				continue
+			}
+			if daysSinceLastUse <= *daysToDelete {
+				continue
+			}
+
+			usersToDelete = append(usersToDelete, user)
+			fmt.Printf("\nAdding %s to remove list: %d days since last active, remove after %d days inactive\n", user.Username, daysSinceLastUse, *daysToDelete)
 		}
 		for _, user := range usersToDelete {
-			removeUser(user)
+			if err := removeUser(user, client, ctx); err != nil {
+				return err
+			}
 			if *toEmail {
 				sendEmail(user)
 			}
 		}
-		fmt.Print(noAdmin)
-		fmt.Print(toEmail)
+
 		return nil
 	}
 
@@ -114,26 +104,25 @@ query Users($first: Int, $query: String) {
 	})
 }
 
-func computeDaysSinceLastUse(user User) (int, error) {
+func computeDaysSinceLastUse(user User) (timeDiff int, wasLastActive bool, _ error) {
 	timeNow := time.Now()
 	//TODO handle for null lastActiveTime = null
 	if user.UsageStatistics.LastActiveTime == "" {
 		fmt.Printf("\n%s has no lastActive value\n", user.Username)
-		return 9999, nil
+		wasLastActive = false
+		return 0, wasLastActive, nil
 	}
 	timeLast, err := time.Parse(time.RFC3339, user.UsageStatistics.LastActiveTime)
 	if err != nil {
 		fmt.Printf("failed to parse lastActive time: %s", err)
+		return 0, false, err
 	}
-	timeDiff := int(timeNow.Sub(timeLast).Hours() / 24)
-	if err != nil {
-		fmt.Printf("failed to diff lastActive to current time: %s", err)
-	}
+	timeDiff = int(timeNow.Sub(timeLast).Hours() / 24)
 
-	return timeDiff, err
+	return timeDiff, true, err
 }
 
-func removeUser(user User) error {
+func removeUser(user User, client api.Client, ctx context.Context) error {
 	query := `mutation DeleteUser(
   $user: ID!
 ) {
@@ -143,8 +132,13 @@ func removeUser(user User) error {
     alwaysNil
   }
 }`
-	reflect.TypeOf(query)
-	fmt.Printf("\nDeleted user: %s\n", user.Username)
+	vars := map[string]interface{}{
+		"user": user.ID,
+	}
+	if ok, err := client.NewRequest(query, vars).Do(ctx, nil); err != nil || !ok {
+		return err
+	}
+	fmt.Printf("\nDeleted user %s: %s\n", user.ID, user.Username)
 	return nil
 }
 
--- a/cmd/src/users_clean.go
+++ b/cmd/src/users_clean.go
@@ -50,7 +50,26 @@ Examples:
 		ctx := context.Background()
 		client := cfg.apiClient(apiFlags, flagSet.Output())
 
-		query := `
+		currentUserQuery := `
+query {
+	currentUser {
+		username
+	}
+}
+`
+		var currentUserResult struct {
+			Data struct {
+				CurrentUser struct {
+					Username string
+				}
+			}
+		}
+		if ok, err := cfg.apiClient(apiFlags, flagSet.Output()).NewRequest(currentUserQuery, nil).DoRaw(context.Background(), &currentUserResult); err != nil || !ok {
+			return err
+		}
+		fmt.Println(currentUserResult)
+
+		usersQuery := `
 query Users() {
 	users() {
 		nodes {
@@ -61,20 +80,25 @@ query Users() {
 ` + userFragment
 
 		// get users to delete
-		var result struct {
+		var usersResult struct {
 			Users struct {
 				Nodes []User
 			}
 		}
-		if ok, err := client.NewRequest(query, nil).Do(ctx, &result); err != nil || !ok {
+		if ok, err := client.NewRequest(usersQuery, nil).Do(ctx, &usersResult); err != nil || !ok {
 			return err
 		}
+		fmt.Println(usersResult)
 
 		usersToDelete := make([]UserToDelete, 0)
-		for _, user := range result.Users.Nodes {
+		for _, user := range usersResult.Users.Nodes {
 			daysSinceLastUse, wasLastActive, err := computeDaysSinceLastUse(user)
 			if err != nil {
 				return err
+			}
+			// never remove user issuing command
+			if user.Username == currentUserResult.Data.CurrentUser.Username {
+				continue
 			}
 			if !wasLastActive && !*removeNoLastActive {
 				continue
--- a/cmd/src/users_clean.go
+++ b/cmd/src/users_clean.go
@@ -97,10 +97,18 @@ query Users($first: Int, $query: String) {
 	})
 }
 
-func delete_users_not_active_in(users_data string, days_threshold int) error {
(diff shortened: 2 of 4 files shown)


DaedalusG:
Noting from @mike-r-mclaughlin  that there should be a flag to bypass scanln the safety check, so users can run this programmatically ✅

unknwon:
@DaedalusG  Do you actually want a review now? The PR description seems to imply we have more work to do on it.

DaedalusG:
Yes please @unknwon, I need to add a lower bound for the -d flag after talking with @ryphil but aside for some styling I think this is about ready to merge 👍

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+`
+
+	flagSet := flag.NewFlagSet("clean", flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		daysToDelete = flagSet.Int("d", 365, "Day threshold on which to remove users, defaults to 365")
```
unknwon, line 28:
```suggestion
		daysToDelete = flagSet.Int("days", 365, "Days threshold on which to remove users, defaults to 365")
```

I think it is more important to be clear than short, unless we can provide both (long and short flag names).

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+		removeNoLastActive = flagSet.Bool("removeNeverActive", false, "removes users with null lastActive value")
+		skipConfirmation   = flagSet.Bool("skip-conf", false, "skips user confirmation step allowing programmatic use")
+		// TODO: Write json file containing users to be deleted, last active usage, site-admin status, and emails
+		// json               = flagSet.Bool("json", false, "Write json file containing users to be deleted, last active usage, site-admin status, and emails")
+		apiFlags = api.NewFlags(flagSet)
+	)
+
+	handler := func(args []string) error {
+		if err := flagSet.Parse(args); err != nil {
+			fmt.Printf("failed to parse command args with error: %s", err)
```
unknwon, line 41: Shouldn't we return the error here instead of printing?

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+			if *noAdmin && user.SiteAdmin {
+				continue
+			}
+			if daysSinceLastUse <= *daysToDelete && wasLastActive {
+				continue
+			}
+			deleteUser := UserToDelete{user, daysSinceLastUse}
+
+			usersToDelete = append(usersToDelete, deleteUser)
+			//fmt.Printf("\nAdding %s to remove list: %d days since last active, remove after %d days inactive\n", user.Username, daysSinceLastUse, *daysToDelete)
```
unknwon, line 89: Do we want to keep this line? If not we can just delete it.

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+	timeNow := time.Now()
+	// handle for null lastActiveTime returned from
+	if user.UsageStatistics.LastActiveTime == "" {
+		wasLastActive = false
+		return 0, wasLastActive, nil
+	}
+	timeLast, err := time.Parse(time.RFC3339, user.UsageStatistics.LastActiveTime)
+	if err != nil {
+		fmt.Printf("failed to parse lastActive time: %s", err)
+		return 0, false, err
```
unknwon, line 147: We should either return or only print error, unless there is very good reason to do both.

unknwon (commented):
There is quite a bit of TODOs, could you collect them to a GitHub issue and be linked from these TODOs?

cmd/src/users.go
```diff
@@ -57,8 +58,12 @@ fragment UserFields on User {
     }
     emails {
         email
-        verified
+		verified
     }
+	usageStatistics {
+		lastActiveTime
+		lastActiveCodeHostIntegrationTime
+	}
```
mrnugget, line 66:
```suggestion
	verified
    }
    usageStatistics {
        lastActiveTime
        lastActiveCodeHostIntegrationTime
    }
```

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+	"github.com/sourcegraph/src-cli/internal/api"
+)
+
+func init() {
+	usage := `
+Examples:
+
+	$ src users clean -d 182
+
+`
```
mrnugget, line 19: I'd add an explanation of what that command does

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		daysToDelete = flagSet.Int("d", 365, "Day threshold on which to remove users, defaults to 365")
+		noAdmin      = flagSet.Bool("no-admin", false, "Omit admin accounts from cleanup")
+		// TODO: Add an email functionality to email removed users, open editor like in git commit
+		// toEmail            = flagSet.Bool("email", false, "send removed users an email")
+		removeNoLastActive = flagSet.Bool("removeNeverActive", false, "removes users with null lastActive value")
+		skipConfirmation   = flagSet.Bool("skip-conf", false, "skips user confirmation step allowing programmatic use")
```
mrnugget, line 33:
```suggestion
		skipConfirmation   = flagSet.Bool("force", false, "skips user confirmation step allowing programmatic use")
```

I think `--force` is more idiomatic to skip confirmation. See [`rm -f`](https://man7.org/linux/man-pages/man1/rm.1.html#top_of_page)

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		daysToDelete = flagSet.Int("d", 365, "Day threshold on which to remove users, defaults to 365")
+		noAdmin      = flagSet.Bool("no-admin", false, "Omit admin accounts from cleanup")
+		// TODO: Add an email functionality to email removed users, open editor like in git commit
+		// toEmail            = flagSet.Bool("email", false, "send removed users an email")
+		removeNoLastActive = flagSet.Bool("removeNeverActive", false, "removes users with null lastActive value")
```
mrnugget, line 32: This uses `camelCase` vs. the [kebab case](https://www.theserverside.com/definition/Kebab-case) of `no-admin` above

cmd/src/users_clean.go
```diff
@@ -0,0 +1,202 @@
...
+	handler := func(args []string) error {
+		if err := flagSet.Parse(args); err != nil {
+			fmt.Printf("failed to parse command args with error: %s", err)
+		}
+
+		ctx := context.Background()
+		client := cfg.apiClient(apiFlags, flagSet.Output())
+
+		vars := map[string]interface{}{
+			"-d": api.NullInt(*daysToDelete),
```
mrnugget, line 48: That's not used, right?

cmd/src/users_clean.go
DaedalusG: Changing this and also adding a minimum value of 60 per @ryphil's feedback

cmd/src/users_clean.go
DaedalusG: ty missed it 👍

cmd/src/users_clean.go
DaedalusG: This was actually some copy pasta that my linter said I should have error handling for I think. Going ahead and returning err

cmd/src/users_clean.go
DaedalusG: Removing the print error, this was mostly just for debugging null values in a users `lastActiveTime`

cmd/src/users_clean.go
DaedalusG: 👍

cmd/src/users_clean.go
DaedalusG: ✅

cmd/src/users_clean.go
DaedalusG: right 😓 - and now I better understand how the `api` package works 😁

unknwon:
Please re-request review once you think it's ready again 😁

cmd/src/users_clean.go
DaedalusG: ✅

DaedalusG:
Hey @unknwon @mrnugget addressed many of your review comments here. The current output from the command is as below:
```
warrengifford@Warrens-MacBook-Pro cmd % src users clean
Users to remove from instance at https://cse-k8s.sgdev.org
╭───────────────────┬──────────────────────────────────┬────────────────────────╮
│ USERNAME          │ EMAIL                            │ DAYS SINCE LAST ACTIVE │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ adeola            │ adeola@sourcegraph.com           │                    300 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ nonso             │ nonso.obiano@sourcegraph.com     │                    160 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ sourcegraph       │ cs-integrations@sourcegraph.com  │                    273 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ g-bot             │ warrenbruceg@gmail.com           │                    253 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ Keegan            │ keegan@sourcegraph.com           │                    211 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ testAtSourcegraph │ testAtSourcegraph@mailinator.com │                    224 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ Malo              │ malo@sourcegraph.com             │                    191 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ ErikSeliger       │ erik@sourcegraph.com             │                    162 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ support           │                                  │                    103 │
╰───────────────────┴──────────────────────────────────┴────────────────────────╯
Do you  wish to proceed with user removal [y/N]: n
Aborting removal
warrengifford@Warrens-MacBook-Pro cmd % src users clean -noAdmin
Users to remove from instance at https://cse-k8s.sgdev.org
╭───────────────────┬──────────────────────────────────┬────────────────────────╮
│ USERNAME          │ EMAIL                            │ DAYS SINCE LAST ACTIVE │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ sourcegraph       │ cs-integrations@sourcegraph.com  │                    273 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ g-bot             │ warrenbruceg@gmail.com           │                    253 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ testAtSourcegraph │ testAtSourcegraph@mailinator.com │                    224 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ Malo              │ malo@sourcegraph.com             │                    191 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ ErikSeliger       │ erik@sourcegraph.com             │                    162 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ support           │                                  │                    103 │
╰───────────────────┴──────────────────────────────────┴────────────────────────╯
Do you  wish to proceed with user removal [y/N]: n
Aborting removal
warrengifford@Warrens-MacBook-Pro cmd % src users clean -days 1
-days flag must be set to 60 or greater
warrengifford@Warrens-MacBook-Pro cmd % src users clean -days 1000 -force
warrengifford@Warrens-MacBook-Pro cmd %
```

cmd/src/users_clean.go
```diff
@@ -0,0 +1,193 @@
...
+	// handle for null lastActiveTime returned from
+	if user.UsageStatistics.LastActiveTime == "" {
+		wasLastActive = false
+		return 0, wasLastActive, nil
+	}
+	timeLast, err := time.Parse(time.RFC3339, user.UsageStatistics.LastActiveTime)
+	if err != nil {
+		return 0, false, err
+	}
+	timeDiff = int(timeNow.Sub(timeLast).Hours() / 24)
```
mrnugget, line 138: I prefer to have variable declarations as close to their first use as possible (there's exceptions of course, such as group of declarations, etc.) and in this case I don't think we need the declaration at all.

```suggestion
	// handle for null lastActiveTime returned from
	if user.UsageStatistics.LastActiveTime == "" {
		wasLastActive = false
		return 0, wasLastActive, nil
	}
	lastActive, err := time.Parse(time.RFC3339, user.UsageStatistics.LastActiveTime)
	if err != nil {
		return 0, false, err
	}
	timeDiff = int(time.Since(timeLast).Hours() / 24)
```

cmd/src/users_clean.go
```diff
@@ -0,0 +1,193 @@
...
+func removeUser(user User, client api.Client, ctx context.Context) error {
+	query := `mutation DeleteUser(
+  $user: ID!
+) {
+  deleteUser(
+    user: $user
+  ) {
+    alwaysNil
+  }
+}`
```
mrnugget, line 153:
```suggestion
	query := `mutation DeleteUser($user: ID!) {
  deleteUser(user: $user) {
    alwaysNil
  }
}`
```

cmd/src/users_clean.go
```diff
@@ -0,0 +1,193 @@
...
+}
+` + userFragment
+
+		// get users to delete
+		var result struct {
+			Users struct {
+				Nodes []User
+			}
+		}
+		if ok, err := client.NewRequest(query, nil).Do(ctx, &result); err != nil || !ok {
```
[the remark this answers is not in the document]
DaedalusG: @mrnugget this is great feedback. This request does always return all users, and that makes sense because we need to check every users activity for this. That being said I'd be way down to learn about altering the graphQL schema -- and just generally exploring the n+1 problem for the backend. This does need to be stable on very large instances as its being developed in direct relation to requests from such orgs 👍 _My naive understanding is that the client side computation should be too big a deal though, since its just basic dates and arithmetic?_

Will throw some time on to discuss with you further 👍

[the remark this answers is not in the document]
mrnugget: > My naive understanding is that the client side computation shouldn't be too big a deal though, since its just basic dates and arithmetic?

True. The computation here is not the problem. It would be totally fine to compute this locally if you already had the data locally.

In this case, though, you first need to fetch the data. And if possible it's good to reduce amount of data being sent over the wire. Especially here, since you're loading _all users_ just to filter them (imagine a scenario in which you'd want to compute "average time since last activity" - for that you'd need to fetch all users locally).

On top of that: that particular query is slow, because I assume it's doing an N+1 query. The 1 query: it loads all users. The N queries: load usageStatistics for each users. So on an instance with 5000 users you end up running 5001 SQL queries, just to fetch the list of users and their `usageStatistics`. (I haven't looked at the code yet, so might be completely off).

[Try running this query on sourcegraph.com and see how long it takes.](https://sourcegraph.com/api/console#%7B%22query%22%3A%22query%20Users%20%7B%5Cn%20%20users%20%7B%5Cn%20%20%20%20nodes%20%7B%5Cn%20%20%20%20%20%20id%5Cn%20%20%20%20%20%20username%5Cn%20%20%20%20%20%20displayName%5Cn%20%20%20%20%20%20usageStatistics%20%7B%5Cn%20%20%20%20%20%20%20%20lastActiveTime%5Cn%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%20%20%20%20pageInfo%20%7B%5Cn%20%20%20%20%20%20hasNextPage%5Cn%20%20%20%20%20%20endCursor%5Cn%20%20%20%20%7D%5Cn%20%20%20%20%5Cn%20%20%7D%5Cn%7D%5Cn%22%2C%22operationName%22%3A%22Users%22%7D)

We can fix the N+1. That might make the request faster. But I'd still think it'd nicer to do the filtering on the server and send less data. The database would have to do less work, the internet would have to transfer less data, the client would need to do less.

Let's use an extreme example to kinda highlight the pattern: say you want to build something like `src download-archives -repo-has-file=horse.txt` that downloads ZIP archives of all repositories that contain `horse.txt` files. You could download _all_ archives of all repositories and then check them locally and throw away the rest. Even ignoring that it'd take up disk space locally, that feels a bit wasteful, right? So instead we can move the compution to the data (instead of moving data to the computation) and ask our server to find repositories that contain this `horse.txt` file (since the server already has the files on disk) and only return those.

cmd/src/users.go
```diff
@@ -57,7 +58,11 @@ fragment UserFields on User {
     }
     emails {
         email
-        verified
+	verified
```
unknwon, line 61:
```suggestion
		verified
```

cmd/src/users_clean.go
```diff
@@ -0,0 +1,188 @@
...
+
+	flagSet := flag.NewFlagSet("clean", flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		daysToDelete       = flagSet.Int("days", 60, "Days threshold on which to remove users, must be 60 days or greater and defaults to this value ")
+		noAdmin            = flagSet.Bool("noAdmin", false, "Omit admin accounts from cleanup")
```
unknwon, line 35: nit: I think giving the opposite option is better, ie. skip admin by default and use the flag `includeAdmin` instead. It feels weird a user clean up ends up deleting admin accounts 🤔

unknwon (approved):
Agree with @mrnugget's other comment on pushing computation to the backend, but I think could be a followup PR if you commit 😁

cmd/src/users_clean.go
DaedalusG: This totally makes sense 👍, changed it

cmd/src/users_clean.go
```diff
@@ -0,0 +1,188 @@
...
+	flagSet := flag.NewFlagSet("clean", flag.ExitOnError)
+	usageFunc := func() {
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+	var (
+		daysToDelete       = flagSet.Int("days", 60, "Days threshold on which to remove users, must be 60 days or greater and defaults to this value ")
+		removeAdmin        = flagSet.Bool("removeAdmin", false, "clean admin accounts")
+		removeNoLastActive = flagSet.Bool("removeNeverActive", false, "removes users with null lastActive value")
```
mrnugget, line 36:
```suggestion
		removeNoLastActive = flagSet.Bool("remove-never-active", false, "removes users with null lastActive value")
```

cmd/src/users_clean.go
```diff
@@ -0,0 +1,188 @@
...
+			if daysSinceLastUse <= *daysToDelete && wasLastActive {
+				continue
+			}
+			deleteUser := UserToDelete{user, daysSinceLastUse}
+
+			usersToDelete = append(usersToDelete, deleteUser)
+		}
+
+		if *skipConfirmation {
+			for _, user := range usersToDelete {
```
mrnugget, line 94:
```suggestion
			for _, user := range usersToDelete {
			  if user.ID == currentUser.ID {
			  	continue
			  }
```

cmd/src/users_clean.go
DaedalusG: Fixed these in latest PR

cmd/src/users_clean.go
DaedalusG: @mrnugget this was bugging me so I took care of it in the last commit here: https://github.com/sourcegraph/src-cli/pull/826/commits/817d78e41d851981cabf226406b603932feac41f

PR #827, opened 2022-08-18
Update release process until 4.0 stable is released.

LawnGnome:
This PR will need to be reverted immediately before 4.0 is released, which will be accomplished by merging #828.

In terms of mechanics, we're relying on GoReleaser's [automatic detection of pre-releases](https://goreleaser.com/customization/release/#github), combined with [`.api/src-cli` automatically skipping pre-release versions of src-cli](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@a00ffa24df05ad716def5d7a053ee5586c6b3194/-/blob/internal/src-cli/version.go?L116:31&subtree=true). The only actual changes we need to make are:

1. Choosing appropriate version numbers, which are now documented in `DEVELOPMENT.md`, and
2. Not pushing the `sourcegraph/src-cli` Docker image to the `latest` tag on `main` until we're ready to release 4.0.0 stable.

Closes sourcegraph/sourcegraph#40575.

### Test plan

I'm going to make 3.43.1 and 4.0.0-rc.0 releases immediately after merging this.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -50,7 +50,8 @@ dockers:
     - "sourcegraph/src-cli:{{ .Tag }}"
     - "sourcegraph/src-cli:{{ .Major }}"
     - "sourcegraph/src-cli:{{ .Major }}.{{ .Minor }}"
-    - "sourcegraph/src-cli:latest"
+    # Disabled until 4.0 stable is released.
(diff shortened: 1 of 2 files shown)


eseliger (approved):
awesome, thanks for figuring this out!

PR #828, opened 2022-08-18
Revert 4.0-pre changes

LawnGnome:
This PR reverts #827, and must be merged before 4.0.0 stable is released.

At the same time this is merged, we should ensure the changelog reflects any releases made on `3.43`.

### Test plan

If #827 works, this will also work.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -50,8 +50,7 @@ dockers:
     - "sourcegraph/src-cli:{{ .Tag }}"
     - "sourcegraph/src-cli:{{ .Major }}"
     - "sourcegraph/src-cli:{{ .Major }}.{{ .Minor }}"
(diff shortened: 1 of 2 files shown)


eseliger (dismissed):
Nice! Making sure we don't accidentally merge :)

LawnGnome:
This was reviewed by @eseliger last month, but he marked it "changes required" so we couldn't accidentally merge this. Re-requesting an approval just to unblock the machinery.

camdencheek (approved):

unknwon (approved):

unknwon:
CleanShot 2022-09-21 at 10 10 09@2x
In case you're in hurry

LawnGnome:
> In case you're in a hurry

You're a gem, Joe. Thanks!

PR #831, opened 2022-08-22
src doc: add code-intel and remove lsif

mrnugget:
See [this Slack thread](https://sourcegraph.slack.com/archives/C03CSAER9LK/p1661140435839809). 

### Test plan

- Ran it locally to test that it generates new files

### DIFF
--- a/cmd/src/doc.go
+++ b/cmd/src/doc.go
@@ -58,7 +58,7 @@ Examples:
 			"config":       &configCommands,
 			"extensions":   &extensionsCommands,
 			"extsvc":       &extsvcCommands,
-			"lsif":         &lsifCommands,
+			"code-intel":   &codeintelCommands,
 			"orgs":         &orgsCommands,
 			"orgs members": &orgsMembersCommands,
 			"repos":        &reposCommands,


mrnugget:
@LawnGnome testing this locally in `sourcegraph/sourcegraph` I found out that we'd need to delete the existing `doc/cli/references/code-intel.md` file for `doc/cli/references/code-intel/index.md` to be used by `docsite`. Any ideas on how to best do that? Manual delete in a PR and then re-generate docs in another PR? (Not sure how all of that fits into your release timelines for src-cli)

cmd/src/doc.go
```diff
@@ -58,7 +58,7 @@ Examples:
 			"config":       &configCommands,
 			"extensions":   &extensionsCommands,
 			"extsvc":       &extsvcCommands,
-			"lsif":         &lsifCommands,
+			"code-intel":   &codeintelCommands,
```
olafurpg, line 61: Is there a way to keep supporting `src lsif` but hide it from `-help`? I'm not convinced this backwards breaking change is worth it because it punishes to customers who have gone through the effort of setting up precise intel in their CI pipelines. We only introduced `src code-intel` a few months ago, I'd expect the deprecation cycle to last at least 1-2 years for an enterprise product like our and I honestly think it's fine to support `src lsif` forever as long as it doesn't cause a maintenance burden.

olafurpg (changes requested):

cmd/src/doc.go
mrnugget: How is this backwards breaking? It changes the docs on dcos.sourcgraph.com from what you see on the left to what you see on the right. It doesn't change anything about how the commands or src-cli works.

screenshot_2022-08-22_10 40 53@2x

varungandhi-src: Yeah, there's no need to remove the old one, there is no major benefit of making the breaking change.

olafurpg (approved):
LGTM 👍 I thought this PR removed the actual command itself.

mrnugget:
> @LawnGnome testing this locally in `sourcegraph/sourcegraph` I found out that we'd need to delete the existing `doc/cli/references/code-intel.md` file for `doc/cli/references/code-intel/index.md` to be used by `docsite`. Any ideas on how to best do that? Manual delete in a PR and then re-generate docs in another PR? (Not sure how all of that fits into your release timelines for src-cli)

Turns out this is not an issue. I only ran into this because I didn't run the `doc.go` file in its entirety. That file cleans the directory so this problem doesn't show up. Still, waiting for review, Adam 😄

varungandhi-src (approved):
Thanks!

LawnGnome (approved):

PR #836, opened 2022-09-06
readme: add instructions for docker

keegancsmith:
We received feedback that some users prefer using src-cli via docker. We publish an image for the purpose of serve-git, so was only documented there. This includes basic information for those inclined to use docker.

Test Plan: run the supplied command in docs

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -63,6 +63,14 @@ chmod +x /usr/local/bin/src
 
 See [Sourcegraph CLI for Windows](WINDOWS.md).
 
+### Installation: Docker
+
+`sourcegraph/src-cli` is published to DockerHub. You can use the `latest` tag or a specific version such as `3.43`. To see all versions view [sourcegraph/src-cli tags](https://hub.docker.com/r/sourcegraph/src-cli/tags).
+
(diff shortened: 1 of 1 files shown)


beyang (approved):

README.markdown
```diff
@@ -63,6 +63,14 @@ chmod +x /usr/local/bin/src
 
 See [Sourcegraph CLI for Windows](WINDOWS.md).
 
+### Installation: Docker
+
+`sourcegraph/src-cli` is published to DockerHub. You can use the `latest` tag or a specific version such as `3.43`. To see all versions view [sourcegraph/src-cli tags](https://hub.docker.com/r/sourcegraph/src-cli/tags).
```
LawnGnome, line 68: (nit, sorry)

```suggestion
`sourcegraph/src-cli` is published to Docker Hub. You can use the `latest` tag or a specific version such as `3.43`. To see all versions view [sourcegraph/src-cli tags](https://hub.docker.com/r/sourcegraph/src-cli/tags).
```

LawnGnome (approved):

PR #842, opened 2022-09-16
batches: Fix remote command - Batch Change ID not provided

### LINKED ISSUE #841: Unable to run remote command
When running `src batch remote -f {some spec file}` the following error occurs,

```
GraphQL errors: {
  "locations": [
    {
      "column": 5,
      "line": 9
    }
  ],
  "message": "Field \"createBatchSpecFromRaw\" argument \"batchChange\" of type \"ID!\" is required but not provided."
}
```

Looks like we are not passing the required batchChange ID to `createBatchSpecFromRaw`.

Piszmog:
Closes #841 

Changes to the GraphQL API broke the `batch remote` command. Updated `upsertEmptyBatchChange` to return the batch change ID and updated `createBatchSpecFromRaw` to accept the batch change ID.

### Test plan

Added Go Tests and did functional testing.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -93,6 +93,7 @@ require (
 	github.com/spf13/cobra v1.4.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/ssor/bom v0.0.0-20170718123548-6386211fdfcf // indirect
+	github.com/stretchr/objx v0.4.0 // indirect
 	github.com/xeipuuv/gojsonpointer v0.0.0-20190905194746-02993c407bfb // indirect
 	github.com/xeipuuv/gojsonreference v0.0.0-20180127040603-bd5ef7bd5415 // indirect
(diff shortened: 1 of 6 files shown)


go.mod
```diff
@@ -93,6 +93,7 @@ require (
 	github.com/spf13/cobra v1.4.0 // indirect
 	github.com/spf13/pflag v1.0.5 // indirect
 	github.com/ssor/bom v0.0.0-20170718123548-6386211fdfcf // indirect
+	github.com/stretchr/objx v0.4.0 // indirect
```
eseliger, line 96: Wonder where that is coming from 🤔

eseliger (approved):
Thanks, also for the tests!

go.mod
Piszmog: This is from me using testify's mock package

BolajiOlajide (approved):

PR #844, opened 2022-09-26
batches: check that docker is running when running src batch

BolajiOlajide:
Closes [#37594](https://github.com/sourcegraph/sourcegraph/issues/37594)

### Test plan

Quit `Docker Desktop` on your PC and run a `src batch` command, you should get a descriptive error message that explains the need for `docker` to make use of `src batch`.

CleanShot 2022-09-26 at 22 27 57@2x


cmd/src/batch_common.go
```diff
@@ -268,9 +268,14 @@ func executeBatchSpec(ctx context.Context, ui ui.ExecUI, opts executeBatchSpecOp
...
-	// Docker working _at all_?).
+	// In the past, we relied on `getBatchParallelism` to ascertain if docker is running,
+	// however, we don't always check for the number of CPUs (especially when the -j parallelis)
+	// flag is passed. This is a more explicit check to confirm docker is working.
+	if err := checkExecutable("docker", "version"); err != nil {
+		fmt.Println("docker info thing")
+		return err
+	}
+
 	parallelism, err := getBatchParallelism(ctx, opts.flags.parallelism)
```
BolajiOlajide, line 279: [One of the reasons why this wasn't working for the docker check is that, when the `-j` flag is passed, we don't bother to check the number of CPUs on the user's machine, therefore failing to confirm if docker is running or not](https://sourcegraph.sourcegraph.com/github.com/sourcegraph/src-cli/-/blob/cmd/src/batch_common.go?L579:1&popover=pinned).

LawnGnome (approved):
Great catch on the `-j` behaviour!

cmd/src/batch_common.go
LawnGnome: I particularly appreciate the updated comment. 😆

adeola-ak (approved):

Piszmog (approved):

cmd/src/batch_common.go
```diff
@@ -268,9 +268,13 @@ func executeBatchSpec(ctx context.Context, ui ui.ExecUI, opts executeBatchSpecOp
...
 
-	// In the past, we checked `docker version`, but now we retrieve the number
-	// of CPUs, since we need that anyway and it performs the same check (is
-	// Docker working _at all_?).
+	// In the past, we relied on `getBatchParallelism` to ascertain if docker is running,
+	// however, we don't always check for the number of CPUs (especially when the -j parallelis)
+	// flag is passed. This is a more explicit check to confirm docker is working.
+	if err := checkExecutable("docker", "version"); err != nil {
+		return err
+	}
```
eseliger, line 276: I think this makes src-cli prone to the same "hanging" issue again that we were running into before, `getBatchParallelism` uses `withFastCommandContext(ctx)` to get around that. @LawnGnome any thoughts?

BolajiOlajide: Oh okay. 

Would a better alternative be to create a method that still runs `docker version` but with the `withFastCommandContext(ctx)` ? WDTY

eseliger: Yes I think so!

LawnGnome: Yep, that would be the best thing to do. Sorry I missed that yesterday!

BolajiOlajide: No problem, that's what reviews are for - to catch this kind of things.
I've implemented the check using `withFastCommandContext(ctx)` though.

internal/batches/docker/context.go
```diff
@@ -94,3 +95,20 @@ func fastCommandTimeout() (time.Duration, error) {
 
 	return fastCommandTimeoutData.timeout, fastCommandTimeoutData.err
 }
+
+// executeFastCommand creates a fastCommandContext used to execute docker commands
+// with a timeout for docker commands that are supposed to be fast (e.g docker info).
+func executeFastCommand(ctx context.Context, args ...string) ([]byte, error) {
```
eseliger, line 101: can the other commands we run against docker that currently use `withFastCommandContext` also use this helper or would that not work?

eseliger (approved):

internal/batches/docker/context.go
BolajiOlajide: It would. I wanted to do that in a separate PR incase there are other considerations to be made.
I'll do that in another PR.

PR #851, opened 2022-10-02
code-intel: Fix double word typo

abitrolly:
### Test plan

Well, changing the message should not fail anything.


LawnGnome (approved):
Thanks!

abitrolly:
@eseliger looks like GitHub is still broken.

LawnGnome:
@abitrolly Sorry! I think this is something funky with the workflow being initiated by a fork owned by a non-Sourcegrapher. I'll file an issue for us to figure that out later, but for now I'll override the check and merge this. Thanks again.

PR #861, opened 2022-10-12
batches: Workspace file upload for local runs

Piszmog:
Closes [#42885](https://github.com/sourcegraph/sourcegraph/issues/42885).

@BolajiOlajide caught this when working on [#31791](https://github.com/sourcegraph/sourcegraph/issues/31791). Luckily it was a quick implementation.

### Test plan

No tests. Did functional testing.

https://user-images.githubusercontent.com/38407415/195401837-ec1ab497-4207-4143-b94b-ade1eff1a4f5.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Added
 
+- Mounted file are now uploaded to the Sourcegraph instance when running `src batch preview` and `src batch apply`. [#861](https://github.com/sourcegraph/src-cli/pull/861)
+
 ### Changed
 
 ### Fixed
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -491,9 +491,11 @@ func executeBatchSpec(ctx context.Context, ui ui.ExecUI, opts executeBatchSpecOp
 	if hasWorkspaceFiles {
 		ui.UploadingWorkspaceFiles()
 		if err = svc.UploadBatchSpecWorkspaceFiles(ctx, batchSpecDir, string(id), batchSpec.Steps); err != nil {
-			return err
+			// Since failing to upload workspace files should not stop processing, just warn
+			ui.UploadingWorkspaceFilesWarning(errors.Wrap(err, "uploading workspace files"))
+		} else {
+			ui.UploadingWorkspaceFilesSuccess()
 		}
-		ui.UploadingWorkspaceFilesSuccess()
 	}
 
 	if !opts.applyBatchSpec {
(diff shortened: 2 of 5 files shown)


eseliger:
I think I am missing a bit of context, what do we require these files for on the backend?

BolajiOlajide:
> I think I am missing a bit of context, what do we require these files for on the backend?

We need to save them in the database so it's displayable in the UI. Currently only `src batch remote ...` saves the file in the database.

eseliger:
`preview` is applicable from the UI as well, so probably they should behave similarly?

eseliger:
I'll tag @malomarrec and @danielmarquespt for some further comments while we figure something out that just occurred to me:

We have to do feature detection here. Likely also for the batch remote command where we already added this. This may only be called if the backend is 4.1 or later. src-cli usually is "compatible with all", in that it wouldn't break with other versions. 
Don't worry it's not as bad as it sounds, we've done that many times and just recently dropped a lot of flags for 4.0: https://github.com/sourcegraph/src-cli/pull/822/files#diff-eb73e64d9518d33b94d2ce1a93bc7ca28fc4d66c4db8f673d194da4c2dfc0067R11

malomarrec:
My take: this is a very valuable addition, and will make batch changes more reproducible as a whole (as someone that reads someone else's batch spec, I also get the scripts without which the spec is kinda meaningless.)

As Erik pointed out, we don't really know how folks use it. So I'd recommend being loud about this changes. It's actually an additive breaking change :trollface:. So in practice:
- add to changelog
- add to in-product changelog a message like "Scripts mounted with src-cli will now be uploaded and shown in the UI. You shouldn't hardcode secrets in scripts, use `env` instead.

Overall, all of this is gonna be much smoother when we support https://github.com/sourcegraph/sourcegraph/issues/27926

adeola-ak (approved):

Piszmog:
@malomarrec do you think that local runs should always should be successful even if there is an issue uploading files to the database?

So we have successfully ran locally, we generated all the changesets, and we start uploading everything to Sourcegraph. While uploading the workspace files, turns out a file used to generate the changesets exceeds our limit (10MB). Do you think we should reject the entire run or have something in the UI to communicate that a file did not make it up?

malomarrec:
I think a warning in the UI is much better: what failed is not blocking the batch change, so it shouldn't fail

Piszmog:
Per @danielmarquespt 

> Ok, so I think we should only display anything when it fails to be on the UI. Otherwise I’m afraid we might worry the user with a warning message that has no impact for them.
>
> We can create the logic to prepare the UI work, because then we need to show it failed when UI displays the mounted files, but right now, if it hasnt any impact then I dont think we should show anything

danielmarquespt:
Regarding the choice of not displaying the notice:

AFAIK right now, a file failing to upload has no impact on anything. So I think we shouldn't display a warning and worry the user that something failed, when it has no impact.

In the future, when we display the files in the UI _then_ we need to disclose that something failed.

Piszmog:
I have update the UI to log a warning. But the error does not stop the execution.

LawnGnome (approved):
I'm probably -0 on the actual change (I think the downsides of uploading local, potentially private files outweigh the advantages here), but since everyone else is fine with it, LGTM technically.

BolajiOlajide (approved):

Issue #862, opened 2022-10-13
Automate versioned Homebrew formula publishing


courier-new:
Follow up to https://github.com/sourcegraph/src-cli/issues/852.

Currently our release process for src-cli only publishes to the main (unversioned) Homebrew formula, which overwrites whatever the last latest version was so that it can be automatically installed with `brew install sourcegraph/src-cli/src-cli`. Ideally, it is possible to install the latest with the main formula, or install any prior version with a specific version command, e.g. `brew install sourcegraph/src-cli/src-cli@4.0.2`.

From this [blog post](https://observiq.com/blog/creating-homebrew-formulas-with-goreleaser/), we understand that publishing a versioned formula and publishing the main formula require different goreleaser configs, because the naming scheme needs to change. We are also constrained in that we cannot easily re-run a release for a given version multiple times without deleting the artifacts uploaded to GitHub and updating all of SHA256 references to them.

To get around these constraints, I think we could implement the following in our GitHub actions workflow:

1. Fetch the previous latest release version tag.
2. Copy the current main release formula file (`src-cli.rb`) to a versioned formula file matching that tag (`e.g. src-cli@4.0.2.rb`), and update the class name to match. Commit the result.
3. Run goreleaser like normal; this updates the main formula again.
4. Remove the existing symlink for the last version and create a new one for the new version.

The next time we release a new version of src-cli, this should automatically get us:

- a versioned formula for the previous release (that you could install with `brew install sourcegraph/src-cli/src-cli@4.0.2`)
- a new versioned formula for 4.0.3 (that you could install with `brew install sourcegraph/src-cli/src-cli@4.0.3`)
- an updated main formula (e.g. what you get when you `brew install sourcegraph/src-cli/src-cli`) that points to 4.0.3 instead of 4.0.2

PR #863, opened 2022-10-14
[wip] automate versioned Homebrew formula publishing

courier-new:
This is the initial, simpler solution to https://github.com/sourcegraph/src-cli/issues/862 which assumes we continue to publish src-cli the way we _typically_ have, i.e. only ever releasing a new _latest version_. It assumes that the newer a tag is, the later the version is. This is absolutely not a good assumption to make most of the time, however it turns out to be remarkably realistic for our repository. With the exception of the 4.0.0-rc release, we have not had a patch release on an older major version since [March of 2020](https://github.com/sourcegraph/src-cli/releases/tag/3.10.13).

It adds a `goreleaser-pre` step which:
- Fetches the current/new and previous version tags
- Converts the current "main" formula in the Homebrew tap repository into a versioned formula

It then lets `goreleaser` run like normal, which will update the main Homebrew formula. 

It then also adds a `goreleaser-post` step which:
- Updates the symlink for the new latest version

### Test plan

- Set up [fork of Homebrew tap repo](https://github.com/courier-new/homebrew-src-cli-test) as if 4.0.1 was the latest release (main formula is for 4.0.1, versioned formulas up to 4.0.0, symlink for 4.0.1)
- Update GH action to skip `goreleaser` step (don't want to actually publish anything new), point pre and post steps at fork repo
- Trigger action
- Observed expected results (main formula is converted into versioned formula, symlink for 4.0.2, no other side effects)


courier-new:
Closing as superseded by #864.

PR #864, opened 2022-10-14
Automate versioned Homebrew formula publishing, fix patch releases

courier-new:
Closes https://github.com/sourcegraph/src-cli/issues/862.

### Context

These GH action workflow changes enable us to automatically version Homebrew formulas for future src-cli releases while also preserving the "main" (unversioned) formula.

Before this, any new release would just overwrite the main formula. This means there was no way to get an older version of src-cli with `brew` unless someone manually published a versioned formula. But what's worse is this also means that if I wanted to cut a patch release for 3.43.3 with a backwards-compatible bug fix that was introduced after 4.0, and I used our standard release process, the next person who tried to install src-cli with `brew install sourcegraph/src-cli/src-cli` wouldn't get 4.0.2, they'd get 3.43.3. Not ideal.

### Implementation overview

This is accomplished with 3 additional jobs to our GH action:

The first step, `release_type` determines what type of release we're doing: patch vs. update latest. This is because we need to make changes to the [Homebrew tap repo](https://github.com/sourcegraph/homebrew-src-cli) ourselves if we're releasing a new latest version, and we need to use a different goreleaser config for either case due to Homebrew's naming convention expectations.

The second step, `goreleaser_pre` only runs if we're releasing a new latest version. It copies the current "main" formula in our Homebrew tap for the previous latest release to a versioned formula, so that it is preserved when goreleaser runs and overwrites the main formula for the latest build. It is not necessary to do this step if the latest version did not change.

Then the normal `goreleaser` step runs. If we're not releasing a new latest version, we use a different config which just publishes a versioned formula, ensuring we don't overwrite the main one. It also skips publishing the Docker image for src-cli:latest. If we _are_ releasing the latest version, we just use the same config as before.

The third additional step, `goreleaser_post` also only runs if we're releasing a new latest version. It simply updates the name of the symlink to the main formula in the Homebrew tap repo. This enables someone to install the latest version either with or without the version. For example right now for 4.0.2:

```bash
# These will do the same thing
brew install sourcegraph/src-cli/src-cli
brew install sourcegraph/src-cli/src-cli@4.0.2
```

It is not necessary to update the symlink if the latest version did not change.

I added lots of comment inline to give clarity to what these steps are and why they are necessary. Ideally this would be something that goreleaser could handle for us automatically, but at the moment it's One Name Pattern to Rule Them All, and we'd rather not forego versions for a main formula, or vice versa. 

I also updated our docs, as I realized we didn't even mention in the README that you can install src-cli with npm/npx.

### Test plan

Obviously I haven't tested the full end-to-end workflow with a full release. It's possible something could still break.

- Set up [fork of Homebrew tap repo](https://github.com/courier-new/homebrew-src-cli-test) as if 4.0.1 was the latest release (main formula is for 4.0.1, versioned formulas up to 4.0.0, symlink for 4.0.1)
- Update GH action to skip goreleaser and npm steps (don't want to actually publish anything new), point pre and post steps at fork Homebrew tap repo
- Trigger action with fake github tag for an older patch release
- Trigger action again with latest tag
- [Example](https://github.com/sourcegraph/src-cli/actions/runs/3252073542) GH actions workflow when releasing an older patch release version
  - Step to copy main formula is skipped
  - Goreleaser runs with patch config file
  - Step to update symlink is skipped 	
- [Example](https://github.com/sourcegraph/src-cli/actions/runs/3252661950) GH actions workflow when releasing a new latest version
  - Main formula is converted to versioned formula and result is committed (except not because `--dry-run`, but y'know)
  - Goreleaser runs with normal config file
  - Symlink is updated and result it committed (except not because `--dry-run` again)


.github/workflows/goreleaser.yml
```diff
@@ -1,16 +1,166 @@
-name: GoReleaser
+name: Build and Release
 
 on:
   push:
-    tags:
+    tags:	
       - '*'
-      - '!latest'
```
courier-new, line 7: We don't use this tag anymore, so I deleted it so as not to be confusing.

.github/workflows/goreleaser.yml
```diff
@@ -1,16 +1,166 @@
...
+        uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
+      - name: Fetch most recent tag, latest (highest version) tag, and second latest tag
+        # For latest and second latest tags, we can't use git tag --sort=version:refname
+        # because git doesn't have a concept of pre-release versions and thus mis-sorts
+        # versions like 4.0.0-rc.0 *after* 4.0.0.
+        run: |
+          echo "most_recent_tag=$(git tag --sort=taggerdate | tail -1)" >> $GITHUB_ENV
+          echo "latest_tag=$(git tag | tr - \~ | sort -V | tr \~ - | tail -1)" >> $GITHUB_ENV
```
courier-new: As the comment above explains, `git tag --sort=version:refname` sorts by semantic version number but doesn't understand pre-release versions, so it sorts versions like `4.0.0-rc.0` after `4.0.0` when they should come before. I shamelessly stole this from a comment on [this gist](https://gist.github.com/loisaidasam/b1e6879f3deb495c22cc), but it worked reliably from my testing. To break it down, since I hate obtuse bash commands:

- `git tag` lists all tags in src-cli.
- `tr - \~` replaces hyphens with tildes, so `4.0.0-rc.0` becomes `4.0.0~rc.0` (this is apparently what's typically understood as pre-releases by certain Linux distros)
- `sort --version-sort` does natural sort by version number of the tags
- `tr \~ -` replaces the tildes with hyphens again
- `tail -1` takes the last item, which based on the sorting should be the latest version tag

The command for the second-latest tag is the same, except it takes the second-to-last item instead.

.github/workflows/goreleaser.yml
```diff
@@ -1,16 +1,166 @@
...
+      - name: Compare tags
+        # If the most recent tag is also the latest tag, semver compare returns 0. If it's
+        # a patch of an older version, semver compare will return -1.
+        run: |
+          if [[ $(semver compare ${{ env.latest_tag }} ${{ env.most_recent_tag }}) == 0 ]]
+          then
+            echo "is_latest_version=1" >> $GITHUB_ENV
+          else
+            echo "is_latest_version=0" >> $GITHUB_ENV
+          fi
```
courier-new, line 74: I couldn't figure out a way to write this as a one-liner but if someone who's better at bash-fu wants to give it a go, be my guest. 😄

.github/workflows/goreleaser.yml
```diff
@@ -1,16 +1,166 @@
...
+      - name: Log variables
+        run: |
+          echo "Most recent (current) tag: ${{ env.most_recent_tag }}"
+          echo "Latest version tag: ${{ env.latest_tag }}"
+          echo "Building current version as new latest?: ${{ env.is_latest_version }}"
```
courier-new, line 79: I could probably get rid of all this logging but it doesn't hurt to keep it, at least until we actually test an actual release. 🙂

DEVELOPMENT.md
```diff
@@ -69,22 +69,11 @@ We adhere to the [general Sourcegraph principles for testing](https://docs.sourc
...
-For example, suppose we have the the recommended versions.
-
-| Sourcegraph version | Recommended src-cli version |
-| ------------------- | --------------------------- | 
-| `3.100`             | `3.90.5`                    |
-| `3.99`              | `3.85.7`                    |
-
-If a new feature is added to a new `3.91.6` release of src-cli and this change requires only features available in Sourcegraph `3.99`, then this feature should also be present in a new `3.85.8` release of src-cli. Because a Sourcegraph instance will automatically select the highest patch version, all non-breaking changes should increment only the patch version. 
-
-Note that if instead the recommended src-cli version for Sourcegraph `3.99` was `3.90.4` in the example above, there is no additional step required, and the new patch version of src-cli will be available to both Sourcegraph versions.
```
courier-new, line 87: This example just confused me, since 99% of the time the recommended src-cli version matches the Sourcegraph version, so I decided to just remove it. Perhaps this was more common historically.

.github/workflows/goreleaser.yml
```diff
@@ -1,16 +1,166 @@
...
+        with:
+          fetch-depth: 0
+      - name: Fetch most recent tag, latest (highest version) tag, and second latest tag
+        # For latest and second latest tags, we can't use git tag --sort=version:refname
+        # because git doesn't have a concept of pre-release versions and thus mis-sorts
+        # versions like 4.0.0-rc.0 *after* 4.0.0.
+        run: |
+          echo "most_recent_tag=$(git tag --sort=taggerdate | tail -1)" >> $GITHUB_ENV
+          echo "latest_tag=$(git tag | tr - \~ | sort --version-sort | tr \~ - | tail -1)" >> $GITHUB_ENV
+          echo "second_latest_tag=$(git tag | tr - \~ | sort --version-sort | tr \~ - | tail -2 | sed -n 1p)" >> $GITHUB_ENV
```
courier-new, line 60: As the comment above explains, `git tag --sort=version:refname` sorts by semantic version number but doesn't understand pre-release versions, so it sorts versions like `4.0.0-rc.0` after `4.0.0` when they should come before. I shamelessly stole this from a comment on [this gist](https://gist.github.com/loisaidasam/b1e6879f3deb495c22cc), but it worked reliably from my testing. To break it down, since I hate obtuse bash commands:

- `git tag` lists all tags in src-cli.
- `tr - \~` replaces hyphens with tildes, so `4.0.0-rc.0` becomes `4.0.0~rc.0` (this is apparently what's typically understood as pre-releases by certain Linux distros)
- `sort --version-sort` does natural sort by version number of the tags
- `tr \~ -` replaces the tildes with hyphens again
- `tail -1` takes the last item, which based on the sorting should be the latest version tag

The command for the second-latest tag is the same, except it takes the second-to-last item instead.

.github/workflows/goreleaser.yml
Piszmog: I would prefer this over a one-liner. Easier to read.

Piszmog (approved):

BolajiOlajide (approved):

PR #866, opened 2022-10-18
temp: only trigger release workflow on explicit bad tag

courier-new:
For https://github.com/sourcegraph/src-cli/issues/878.

Releasing src-cli today didn't go exactly as planned because the action found the wrong tag for the most recent one. It turns out the way the src-cli repo is cloned into the container for the first job, it changes the git tag `taggerdate`s, so `git tag --sort=taggerdate` isn't a reliable sort mechanism for getting the most recent tag.

I should be able to fix this by using a GitHub environment variable that is available during the action , but I can only test it by pushing more tags. Thus, my plan to fix this up is as follows:

1. Merge this PR, which disables this action from running except when if the tag "DO-NOT-PUSH-ME" is pushed. This enables me to push test tags and delete them later, without triggering new releases. I already manually finished the release for src-cli 4.1.0, so there shouldn't be any need to push a legitimate real tag in the meantime.
2. Push up a branch that sets `most_recent_tag` with the GitHub environment variable, and disables the rest of the release workflow except the first job that determines the release type
3. Create a temporary test tag for 4.1.1 on that commit and trigger the workflow
4. Verify the correct most recent version is set and the release type is flagged as new latest rather than patch

### Test plan

Described above.

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -2,8 +2,8 @@ name: Build and Release
 
 on:
   push:
-    tags:	
-      - '*'
+    tags:
+      - 'DO-NOT-PUSH-ME'
 
 # There are two cases where this GitHub action will run:
 #


Piszmog (approved):

courier-new:
Oh that's awesome, thanks Bolaji!! I even tried to look for something like this last week. I'll give that a shot first and see if I can get around needing to disable this on `main`.

Issue #878, opened 2022-11-02
src-cli: fix latest version release detection


courier-new:
Releasing src-cli 4.1 didn't go exactly as planned because the release action identified the wrong tag for the most recent one. It turns out that the way the src-cli repo is cloned into the container for the first job, it resets the git tag taggerdates, so `git tag --sort=taggerdate` isn't a reliable sort mechanism for getting the most recent tag.

We should be able to fix this by using the GitHub environment variable that is available during the action, `GITHUB_REF_NAME`, instead. But this is slightly annoying, because the change is only testable _when a new tag is pushed_.

This issue encompasses the work to make the fix, set up a state to test it in, and teardown afterwards.

PR #879, opened 2022-11-04
Use `GITHUB_REF_NAME` to determine current tag for release workflow

courier-new:
(Hopefully) closes https://github.com/sourcegraph/src-cli/issues/878.

We'll now use the GitHub-provided environment variable `GITHUB_REF_NAME` to determine the current (most recent) tag, rather than the unreliable sort. According to the [docs](https://docs.github.com/en/actions/learn-github-actions/environment-variables#default-environment-variables), this value will be:

> The short ref name of the branch or tag that triggered the workflow run. This value matches the branch or tag name shown on GitHub. For example, `feature-branch-1`.

I was able to test this locally with [act](https://github.com/nektos/act) (thanks Bolaji!!) using simulated tag push events. I tested that a new tag triggered the "new latest" version of the workflow and that an older patch version tag triggered the patch version. As a caveat, since I could only test with a simulated event, we of course can't guarantee that a real GitHub event wouldn't behave differently. But I have higher confidence, at least.

I also tried to clarify a couple comments, mostly because after staring at the words "latest" and "current" and "most recent" for a couple days, they all started to blur together in my mind. 🤪

### Test plan

I'll take point on our 4.2 release (or any sooner patch, if it comes up) later this month to verify the workflow succeeds.

### DIFF
--- a/.github/workflows/goreleaser.yml
+++ b/.github/workflows/goreleaser.yml
@@ -35,9 +35,9 @@ on:
 # goreleaser_post [if case 1.]  - updates symlink to latest version
 # npm             [always]      - publishes to npm
 jobs:
-  # release_type compares the most recently created tag to the highest versioned tag
-  # available on the repo to determine if this release is for a new latest version or a
-  # patch of an older version.
+  # release_type compares the current tag to the highest versioned tag available on the
+  # repo to determine if this release is for a new latest version or a patch of an older
+  # version.
   release_type:
     name: Determine release type
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


eseliger (approved):
I trust you that this makes sense, my bash knowledge is about the same of a 2 year old 😬

.github/workflows/goreleaser.yml
```diff
@@ -50,33 +50,42 @@ jobs:
...
-      - name: Fetch most recent tag, latest (highest version) tag, and second latest tag
+      - name: Set current tag, latest (highest version) tag, and second latest tag
         # For latest and second latest tags, we can't use git tag --sort=version:refname
         # because git doesn't have a concept of pre-release versions and thus mis-sorts
         # versions like 4.0.0-rc.0 *after* 4.0.0.
         run: |
-          echo "most_recent_tag=$(git tag --sort=taggerdate | tail -1)" >> $GITHUB_ENV
+          echo "current_tag=${GITHUB_REF_NAME}" >> $GITHUB_ENV
           echo "latest_tag=$(git tag | tr - \~ | sort --version-sort | tr \~ - | tail -1)" >> $GITHUB_ENV
           echo "second_latest_tag=$(git tag | tr - \~ | sort --version-sort | tr \~ - | tail -2 | sed -n 1p)" >> $GITHUB_ENV
```
BolajiOlajide, line 60: I'm not sure storing these variables in `$GITHUB_ENV` will work because of how `GITHUB_ENV` works. 
CleanShot 2022-11-04 at 12 40 25

It is unique for each step, so it's possible the tags set here won't be accessible in the `Compare tags` step.

Then again the best way to know is by testing it, but I know I've run into this problem in the past that's why I thought to highlight it.

BolajiOlajide (approved):

.github/workflows/goreleaser.yml
courier-new: Thanks for flagging! Thankfully, I've been able to test this part cross-step successfully -- the items from `Compare tags` and `Log variables` successfully get the values for items set in `GITHUB_ENV` so long as they're part of the same job. Though the wording of this blurb you shared certainly makes me doubt that I had those observations, haha. 😅

PR #887, opened 2022-11-24
Support binary diffs

eseliger:
Patches can contain non UTF-8 characters (who knew). Since we JSON-encode the patch as a string field, we lose that encoding over the wire to Sourcegraph. This PR adds support for a byte slice encoded (so base64 over the wire) mode from Sourcegraph 4.4 onwards, which retains original encoding.

### Test plan

Validated this src-cli version still works with older Sourcegraph backends (then without this fix), and with a current Sourcegraph backend where this fixes patches with non UTF-8. Tested with src-cli execution and SSBC.


.tool-versions
```diff
@@ -1,3 +1,3 @@
-golang 1.18.1
+golang 1.19.3
```
eseliger, line 1: Had to upgrade, Sourcegraph now requires it.

go.mod
```diff
@@ -117,3 +117,5 @@ require (
 
 // See: https://github.com/ghodss/yaml/pull/65
 replace github.com/ghodss/yaml => github.com/sourcegraph/yaml v1.0.1-0.20200714132230-56936252f152
+
+replace github.com/sourcegraph/sourcegraph/lib => /Users/erik/Code/sourcegraph/sourcegraph/lib
```
eseliger, line 121: TODO: Upgrade once sg/sg is up to date

internal/batches/features.go
```diff
@@ -28,6 +28,7 @@ func (ff *FeatureFlags) SetFromVersion(version string) error {
 		// Example usage:
 		// {&ff.FlagName, ">= 3.23.0-0", "2020-11-24"},
 		{&ff.Sourcegraph40, ">= 4.0.0-0", "2022-08-24"},
+		{&ff.BinaryDiffs, ">= 4.3.0-0", "2022-11-24"},
```
eseliger, line 31: TODO: Adjust date prior to merge.

LawnGnome (approved):
I haven't tested this yet — I'll do that after I've reviewed the backend PR — but this LGTM. I'm sort of looking at the `[]byte()` conversions that are getting removed here and there and kicking myself that we didn't do this (much) sooner.

PR #889, opened 2022-11-26
snapshot: add pg_dump command builder with 'snapshot databases'

bobheadxi:
Add simple utility for generating commands that can then be run/modified to create database dumps.

Right now, it's mostly for reference - as we build experience with different setups we can see if this can be made more flexible in the right ways. I believe a templating command like this is important to have, however, since we want to make sure all databases are exported and the correct flags are used with `pg_dump`, and documentation + manual copy-paste is error-prone.

Closes https://github.com/sourcegraph/customer/issues/1611 for now.

### Test plan

image

Also tested the kubectl commands on k8s.sgdev.org

### DIFF
--- a/cmd/src/snapshot_databases.go
+++ b/cmd/src/snapshot_databases.go
@@ -54,8 +54,13 @@ TARGETS FILES
 			var commandBuilder pgdump.CommandBuilder
 			switch builder {
 			case "pg_dump", "":
+				targetKey = "local"
 				commandBuilder = func(t pgdump.Target) (string, error) {
-					return pgdump.Command(t), nil
+					cmd := pgdump.Command(t)
+					if t.Target != "" {
+						return fmt.Sprintf("%s --host=%s", cmd, t.Target), nil
+					}
+					return cmd, nil
 				}
 			case "docker":
 				commandBuilder = func(t pgdump.Target) (string, error) {
@@ -106,6 +111,23 @@ TARGETS FILES
 
 // predefinedDatabaseDumpTargets is based on default Sourcegraph configurations.
 var predefinedDatabaseDumpTargets = map[string]pgdump.Targets{
+	"local": {
+		Primary: pgdump.Target{
+			DBName:   "sg",
+			Username: "sg",
+			Password: "sg",
+		},
+		CodeIntel: pgdump.Target{
+			DBName:   "sg",
+			Username: "sg",
+			Password: "sg",
+		},
+		CodeInsights: pgdump.Target{
+			DBName:   "postgres",
+			Username: "postgres",
+			Password: "password",
+		},
+	},
 	"docker": { // based on deploy-sourcegraph-managed
 		Primary: pgdump.Target{
 			Target:   "pgsql",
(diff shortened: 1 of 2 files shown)


cmd/src/snapshot_databases.go
```diff
@@ -0,0 +1,151 @@
...
+				}
+			case "docker":
+				commandBuilder = func(t pgdump.Target) (string, error) {
+					// docker exec -it pgsql sh -c 'pg_dump --no-owner --format=p --no-acl --username=sg --dbname=sg'
+					return fmt.Sprintf("docker exec -it %s sh -c '%s'", t.Target, pgdump.Command(t)), nil
+				}
+			case "kubectl":
+				targetKey = "k8s"
+				commandBuilder = func(t pgdump.Target) (string, error) {
+					// docker exec -it pgsql sh -c 'pg_dump --no-owner --format=p --no-acl --username=sg --dbname=sg'
```
michaellzc, line 68:
```suggestion
					// kubect exec -it pgsql -- bash -c 'pg_dump --no-owner --format=p --no-acl --username=sg --dbname=sg'
```

bobheadxi: oops, did not intend to keep this docstring

internal/pgdump/pgdump.go
```diff
@@ -0,0 +1,64 @@
...
+type Targets struct {
+	Primary      Target `yaml:"primary"`
+	CodeIntel    Target `yaml:"codeintel"`
+	CodeInsights Target `yaml:"codeinsights"`
+}
+
+// Target represents a database for pg_dump to export.
+type Target struct {
+	// e.g. container, deployment, statefulset, etc.
+	Target string `yaml:"entity"`
```
michaellzc, line 19:
```suggestion
	// Target is the name of the database deployment
	// When built-in database is used,
	// docker: the name of the database container, e.g. pgsql, codeintel-db, codeinsights-db
	// k8s: the name of the deployment or statefulset, e.g. deploy/pgsql, sts/pgsql, 
	Target string `yaml:"entity"`
```

I think we could also interpret `target` as hostname of the database host. for example, a customer could be using an external database

michaellzc: also, should we rename the key name in yaml to `target` instead of `entity`? I feel like the word `entity` is a bit confusing

bobheadxi: Ah yes the yaml should definitely be `target`!

bobheadxi:
Current dependencies on/for this PR:
* main
  * **PR #889** <a href="https://app.graphite.dev/github/pr/sourcegraph/src-cli/889" target="_blank">Graphite</a>  👈
    * **PR #890** <a href="https://app.graphite.dev/github/pr/sourcegraph/src-cli/890" target="_blank">Graphite</a> 

This comment was auto-generated by [Graphite](https://app.graphite.dev/github/pr/sourcegraph/src-cli/889?utm_source=stack-comment).

michaellzc (approved):

PR #893, opened 2022-11-30
snapshot: trim extension statements from pgdump when uploading

bobheadxi:
When uploading a snapshot, trim extension statements unsupported by Google Cloud SQL. We do this on upload so that the customer's copy of the dump remains unmodified (in case something goes wrong, we still have the original to work with - trimming can be disabled with `trim-extensions=false`).

Closes https://github.com/sourcegraph/customer/issues/1613

### Test plan

```
go run ./cmd/src snapshot upload -credentials='../cloud-data-migrations/misc/credential.json' -bucket='sourcegraph-cloud-data-migration-7ec798d67ff4'
```

See uploaded data

### DIFF
--- a/internal/pgdump/extensions.go
+++ b/internal/pgdump/extensions.go
@@ -4,6 +4,8 @@ import (
 	"bufio"
 	"bytes"
 	"io"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
 // PartialCopyWithoutExtensions will perform a partial copy of a SQL database dump from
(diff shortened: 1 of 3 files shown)


internal/pgdump/extensions.go
```diff
@@ -0,0 +1,63 @@
...
+		noMoreExtensions bool
+	)
+
+	for !noMoreExtensions {
+		// Read up to a line, keeping track of our position in src
+		var line []byte
+		line, err = reader.ReadBytes('\n')
+		consumed += int64(len(line))
+		if err != nil {
+			return
```
daxmc99, line 36: Style nit: I was surprised to not see an `err` here or in the rest of the function

internal/pgdump/extensions.go
```diff
@@ -0,0 +1,63 @@
...
+		}
+
+		// Once we start seeing table creations, we are definitely done with extensions,
+		// so we can hand off the rest to the superior io.Copy implementation.
```
daxmc99, line 40: Are we sure about this? I'm lacking some context if this is being used by end users or not but this is a large assumption. 
The official postgres docs use things like `split` https://www.postgresql.org/docs/12/backup-dump.html#BACKUP-DUMP-LARGE

daxmc99 (approved):

internal/pgdump/extensions.go
bobheadxi: If you're referring to table creations being before extensions, then yes - extensions are typically a prerequisite

As for handing off to io.Copy, yes - the io.Copy implementation is more robust and efficient, and the GCS bucket handler will handle chunking for us. I don't think Cloud SQL will be happy with split dumps, and to piece it together we need to download it somewhere, which I'd like to avoid (right now the process can just have us import directly from GCS)

internal/pgdump/extensions.go
unknwon: +1 nit: not a fan of naked return... 😛

unknwon (approved):

internal/pgdump/extensions.go
bobheadxi: Updated to use explicit returns! [8020472](https://github.com/sourcegraph/src-cli/pull/893/commits/8020472d04b28ceb8f74b25241eb1179437ae6fb)

PR #896, opened 2022-12-02
add CRUD support for repo key-value pairs

tbliu98:
This PR adds commands to src-cli to provide an alternative way for users to manage repo key-value pairs without having to interact directly with GraphQL.

- Add field KeyValuePairs to the Repository type so it will be part of the response object for repo read requests
- Add commands `src repos add-kvp`, `src repos update-kvp`, and `src repos delete-kvp` which correspond to the GraphQL endpoints `addRepoKeyValuePair`, `updateRepoKeyValuePair`, and `deleteRepoKeyValuePair`, respectively
    - These commands will exit with an error if all required flags are not provided

### Test plan
Manually tested against local dev instance


cmd/src/repos_add_kvp.go
```diff
@@ -0,0 +1,99 @@
...
+import (
+	"context"
+	"flag"
+	"fmt"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/src-cli/internal/api"
+)
+
+func init() {
```
camdencheek, line 12: IMO, this is a lot of code to put in an `init()`. Style-wise, I think it would be best to move most of this stuff outside the `init()` function and just put the "register the command" step inside. Then it's clear what actually _runs_ on startup.

cmd/src/repos_add_kvp.go
```diff
@@ -0,0 +1,99 @@
...
+			if f.Name == "key" {
+				key := f.Value.String()
+				keyFlag = &key
+			}
+
+			if f.Name == "value" {
+				value := f.Value.String()
+				valueFlag = &value
+			}
+		})
```
camdencheek, line 56: Can't you just use `keyFlag` and `valueFlag` above?

cmd/src/repos_add_kvp.go
```diff
@@ -0,0 +1,99 @@
...
+  $repo: ID!,
+  $key: String!,
+  $value: String,
+) {
+  addRepoKeyValuePair(
+    repo: $repo,
+    key: $key,
+    value: $value,
+  ) {
+    alwaysNil
```
camdencheek, line 73: didn't you add errors [here](https://github.com/sourcegraph/sourcegraph/pull/45023#pullrequestreview-1201722757)?

cmd/src/repos_add_kvp.go
```diff
@@ -0,0 +1,99 @@
...
+  }
+}`
+
+		if ok, err := client.NewRequest(query, map[string]interface{}{
+			"repo":  *repoFlag,
+			"key":   *keyFlag,
+			"value": valueFlag,
+		}).Do(context.Background(), nil); err != nil || !ok {
```
camdencheek, line 81: This should probably be `context.WithTimeout`

camdencheek (approved):
Nice! This will definitely make this feature a bit more accessible

cmd/src/repos_add_kvp.go
tbliu98: Good point, but I was just following the convention that's used for several other commands I looked at in this repo, and I feel like it's preferable to keep the style consistent

cmd/src/repos_add_kvp.go
tbliu98: I thought I could, but it turns out that when you define flags using `flagset.String()` ([these lines](https://github.com/sourcegraph/src-cli/pull/896/files#diff-ae7d3d85e9e477ad6a3b6fe781d1637f29d2aae7bee90f3671c8be8679cb196eR29-R34)), it requires you to provide a default value for the flag, and that can't be `nil` for strings. So I'm setting the empty string `""` as the default value for the flags, which seems to be the convention elsewhere in this repo. But this means that if you do `src repos add-kvp -repo=myrepo` (omitting `-key` and `-value`) it'll create a KVP of the form `"":""` which is probably not what the user wants. This also means that without this call to `flagset.Visit()` there's no way for a user to create a key with a null value.

So using `flagset.Visit()` this way helps avoid surprises for the user by erroring out if they don't explicitly set `-key`. And if they want to create a tag they can just omit `-value` and it'll be created with a null value. If they _do_ want to create a key that's an empty string then they can do that by explicitly setting that flag's value to an empty string (like `-key=`)

cmd/src/repos_add_kvp.go
tbliu98: Yeah, the errors do get surfaced to the user:
```
% go run ./cmd/src repos add-kvp -repo=fakerepo -key=mykey
GraphQL errors: {
  "message": "invalid graphql.ID",
  "path": [
    "addRepoKeyValuePair"
  ]
}
exit status 1
```
But `alwaysNil` needs to be here because we need to specify a selection of fields from the return type of the GraphQL endpoint (which is `EmptyResponse` which only has one field, `alwaysNil`)

Screen Shot 2022-12-05 at 4 11 02 PM

If I remove the `alwaysNil` selection from the query string I get:
```
% go run ./cmd/src repos add-kvp -repo=fakerepo -key=mykey
GraphQL errors: {
  "locations": [
    {
      "column": 3,
      "line": 6
    }
  ],
  "message": "Field \"addRepoKeyValuePair\" of type \"EmptyResponse!\" must have a selection of subfields. Did you mean \"addRepoKeyValuePair { ... }\"?"
}
exit status 1
```

cmd/src/repos_add_kvp.go
tbliu98: I don't think any of the other commands use `context.WithTimeout`, so I kind of assumed it was fine to use `context.Background()` as well here

cmd/src/repos_add_kvp.go
camdencheek: Aah, gotcha. That makes sense. Thanks for explaning!

cmd/src/repos_add_kvp.go
camdencheek: Oh duh. Forgot that `alwaysNil` could still have an error returned

cmd/src/repos_add_kvp.go
camdencheek: Huh. That's...not great. But no worse than the current state, so I guess it's fine for now. We should really be inheriting the context from something that respects an interrupt but that's way out of scope of this PR.

PR #897, opened 2022-12-06
[language-platform] update src-cli to handle scip file upload

Numbers88s:
Fixes https://github.com/sourcegraph/sourcegraph/issues/45078

Before this PR src codeintel upload would translate a SCIP file to LSIF prior to upload. With this PR it now instead checks with the instance to see if it can handle SCIP uploads. This is done with a HEAD request to the upload endpoint it then updates the headers to reflect the SCIP metadata.

### Test plan

To test properly you will need the changes from this PR https://github.com/sourcegraph/sourcegraph/pull/45269

1. Update `go.mod` in this repo to point to your local environment with the changes above.
Example:

```yaml
replace github.com/sourcegraph/sourcegraph/lib => /Users/<YOUR NAME>/go/src/sourcegraph/lib
```
2. Build src-cli to have the current build. i.e `go build` inside the src-cli/cmd/src directory
3. Navigate to any typescript repo and use https://github.com/sourcegraph/scip-typescript to create an index to upload. I tend to index `sourcegraph/sourcegraph/client/web`
4. Use your newly built src-cli (from step 2) to upload your newly created index (from step 3)
```terminal
➜ SRC_ENDPOINT=https://sourcegraph.test:3443 ../../../src-cli/cmd/src/src lsif upload -indexer=scip-typescript -indexerVersion=v0.3.3
```
5. Success you can hit the new endpoint.

At the moment, the instance cannot handle SCIP. But this PR (https://github.com/sourcegraph/sourcegraph/pull/45104) should handle that part.


efritz (approved):
LGTM. I'd like to run it by hand though once we have some of the changes to sg/sg landed before this lands as well.

cmd/src/code_intel_upload.go
```diff
@@ -132,12 +137,19 @@ func handleCodeIntelUpload(args []string) error {
...
+func codeintelUploadOptions(out *output.Output, isSCIPAvailable bool) upload.UploadOptions {
 	var associatedIndexID *int
 	if codeintelUploadFlags.associatedIndexID != -1 {
 		associatedIndexID = &codeintelUploadFlags.associatedIndexID
 	}
 
+	cfg.AdditionalHeaders["Content-Type"] = "application/x-ndjson+lsif"
+	path := codeintelUploadFlags.uploadRoute
+	if isSCIPAvailable {
+		cfg.AdditionalHeaders["Content-Type"] = "application/x-protobuf+scip"
```
efritz, line 149: Because https://github.com/sourcegraph/sourcegraph/pull/45269 has landed, you'll need to update your `sg/sg/lib` version in go.mod.

efritz (approved):

efritz:
@Numbers88s There's a point where we try to read the metadata vertex from the LSIF file to get the indexer name/version. We should also support reading this from the Metadata message in a SCIP proto file. Can you find out where that is and modify it as well?

efritz:
Please review and merge #907 into this branch before landing. We also need a changelog entry.

efritz:
Fixes https://github.com/sourcegraph/sourcegraph/issues/45078.

efritz:
@Numbers88s I made a few updates in #911 that target this branch. This change makes the default file value include SCIP files if there's no ambiguity. I tested all four scenarios (instance does/doesn't accept scip vs. scip/lsif upload file) and everything seems good.

@Numbers88s Could you add an item to the CHANGELOG and then I think we can get read to merge this after we've updated the codeintel-qa to use SCIP directly (check for any last minute holes we can plug before we start rewriting data in production).

efritz (approved):

CHANGELOG.md
```diff
@@ -9,7 +9,19 @@
 
 All notable changes to `src-cli` are documented in this file.
 
-## Unreleased
+## Unreleased (4.5)
```
efritz, line 12: - [ ] Merge after 4.4 is released, update changelog.

PR #898, opened 2022-12-07
add watchdog goroutine for docker

BolajiOlajide:
Part 2 [#44432](https://github.com/sourcegraph/sourcegraph/issues/44432)
Closes [#44432](https://github.com/sourcegraph/sourcegraph/issues/44432)

### Test plan

Manually tested that the banner shows up when there's an error from getting the current context for docker. 

CleanShot 2022-12-07 at 19 26 03@2x

### DIFF
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -32,9 +32,12 @@ import (
 	"github.com/sourcegraph/src-cli/internal/batches/repozip"
 	"github.com/sourcegraph/src-cli/internal/batches/service"
 	"github.com/sourcegraph/src-cli/internal/batches/ui"
+	"github.com/sourcegraph/src-cli/internal/batches/watchdog"
 	"github.com/sourcegraph/src-cli/internal/batches/workspace"
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
 )
+
+const dockerWatchDuration = 5 * time.Second
 
 // batchExecutionFlags are common to batch changes that are executed both
 // locally and remotely.
@@ -251,6 +254,15 @@ type executeBatchSpecOpts struct {
 	client api.Client
 }
 
+func createDockerWatchdog(ctx context.Context, execUI ui.ExecUI) watchdog.Watcher {
+	return watchdog.New(dockerWatchDuration, func() {
+		_, err := docker.CurrentContext(ctx)
+		if err != nil {
+			execUI.DockerWatchDogWarning(errors.Wrap(err, "docker watchdog"))
+		}
+	})
+}
+
 // executeBatchSpec performs all the steps required to upload the batch spec to
 // Sourcegraph, including execution as needed and applying the resulting batch
 // spec if specified.
@@ -263,7 +275,11 @@ func executeBatchSpec(ctx context.Context, opts executeBatchSpecOpts) (err error
 		execUI = &ui.TUI{Out: out}
 	}
 
+	w := createDockerWatchdog(ctx, execUI)
+	go w.Start()
+
 	defer func() {
+		w.Stop()
 		if err != nil {
 			execUI.ExecutionError(err)
 		}
--- a/cmd/src/batch_common.go
+++ b/cmd/src/batch_common.go
@@ -37,7 +37,8 @@ import (
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
 )
 
-const dockerWatchDuration = 5 * time.Second
+// We check for docker responsiveness every minute
+const dockerWatchDuration = 1 * time.Minute
 
 // batchExecutionFlags are common to batch changes that are executed both
 // locally and remotely.
@@ -254,7 +255,7 @@ type executeBatchSpecOpts struct {
 	client api.Client
 }
 
-func createDockerWatchdog(ctx context.Context, execUI ui.ExecUI) watchdog.Watcher {
+func createDockerWatchdog(ctx context.Context, execUI ui.ExecUI) *watchdog.WatchDog {
 	return watchdog.New(dockerWatchDuration, func() {
 		_, err := docker.CurrentContext(ctx)
 		if err != nil {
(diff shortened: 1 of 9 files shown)


internal/batches/watchdog/watchdog.go
```diff
@@ -0,0 +1,39 @@
+package watchdog
+
+import (
+	"time"
+)
+
+type WatchDog struct {
+	interval time.Duration
+	ticker   *time.Ticker
+	done     chan struct{}
```
LawnGnome, line 10: Could this be simplified to a single `*time.Ticker`, with `Stop` calling `w.ticker.Stop()` and `Start()` having something like:

```go
for range <-w.ticker.C { go w.callback() }
```

(This isn't a rhetorical question; I've mostly been thinking about Rust channels recently, and they have different semantics.)

internal/batches/ui/tui.go
```diff
@@ -449,3 +453,10 @@ func batchCompletePending(p output.Pending, message string) {
 func batchCompleteWarning(p output.Pending, message string) {
 	p.Complete(output.Line(batchWarningEmoji, batchWarningColor, message))
 }
+
+func dockerWatchDogWarning(out *output.Output, err error) {
+	block := out.Block(output.Line("🐳", output.StyleWarning, "Hey there, Your docker engine might be frozen."))
+	block.WriteLine(output.Line("", output.StyleWarning, "If there's no progress in the next couple minutes, you can try rebooting docker and running the command again."))
```
LawnGnome, line 459: Very slight wording tweaks (more input encouraged from other Batchers!):

```suggestion
	block := out.Block(output.Line("🐳", output.StyleWarning, "Hey there, your Docker engine might be frozen."))
	block.WriteLine(output.Line("", output.StyleWarning, "If there's no progress in the next couple minutes, you may want to try restarting Docker and running the command again."))
```

cmd/src/batch_common.go
```diff
@@ -251,6 +255,15 @@ type executeBatchSpecOpts struct {
 	client api.Client
 }
 
+func createDockerWatchdog(ctx context.Context, execUI ui.ExecUI) *watchdog.WatchDog {
+	return watchdog.New(dockerWatchDuration, func() {
+		_, err := docker.CurrentContext(ctx)
```
LawnGnome, line 260: I'm not sure this actually goes back to the Docker Engine in all cases — in theory, the client can answer this from its own configuration. (I briefly tested this by stopping Docker Desktop and running `docker context ls` on the command line, but I haven't put much more effort in than that, admittedly.)

Should we call `NCPU` to be on the safe side? That one _definitely_ has to go to the server.

LawnGnome (commented):
Generally LGTM. I'm surprised at how little code this actually was!

cmd/src/batch_common.go
BolajiOlajide: Ha! Nice catch - I definitely didn't know that.

internal/batches/ui/tui.go
BolajiOlajide: This looks good.

@danielmarquespt Do you have any thoughts?

danielmarquespt: Personal feeling, I would remove the `"Hey there"`, and just leave it as `"It seems like your docker engine might be frozen..."`

BolajiOlajide: Got it. Updated.

internal/batches/watchdog/watchdog.go
BolajiOlajide: Oh nice! Didn't know one could do that.
Admittedly, (I shared with @Piszmog earlier), channels are one part of Go I'm still not very familiar with but this ticket helped me gain a bit of understanding there.

internal/batches/watchdog/watchdog.go
```diff
@@ -0,0 +1,39 @@
...
+	close(w.done)
+}
+
+func (w *WatchDog) Start() {
+	for {
+		select {
+		case <-w.ticker.C:
+			go w.callback()
+		case <-w.done:
+			w.ticker.Stop()
```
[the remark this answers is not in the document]
BolajiOlajide: Reverted the implementation so no need for this anymore.

internal/batches/watchdog/watchdog.go
```diff
@@ -0,0 +1,30 @@
...
+		callback: callback,
+	}
+}
+
+func (w *WatchDog) Stop() {
+	w.ticker.Stop()
+}
+
+func (w *WatchDog) Start() {
+	for _ = range w.ticker.Chan() {
```
LawnGnome, line 27: I realised while re-reviewing this that we can't do it quite this simply — I vaguely remembered that there was some weirdness around something to do with channels, and it was that [stopped Tickers are not actually closed](https://pkg.go.dev/time#Ticker.Stop), because… well, you can read the docs. (I honestly interpret it mostly as "because fuck you, that's why".)

This would have been OK if it was a "normal" channel that was closed when it was no longer needed. Bah.

So, two options:

1. We reinstate the previous, two channel behaviour, just with a `return` on the `done` case, or
2. We decide we don't care about leaking a goroutine, because `src` is likely about to exit anyway.

I'm not super fussed either way. I'm sorry for forgetting the nuance until after you'd done the work!

BolajiOlajide: There's no need to apologize. This ticket has been a big learning experience with channels for me - learnt a lot from pairing with @Piszmog earlier.

I actually read the docs and saw the comment about the channel not being closed after calling Stop() but I figured it wasn't going to be an issue since src will edit anyway.

However now that I think about it, if src ever tried doing some background work like we do for sg (to check for an update whenever you run a command), then this could lead to an issue.

So I'll reinstate the former way I did it tomorrow, to avoid this. Thanks for taking a closer look @LawnGnome

PR #901, opened 2022-12-15
Rename users clean or prune -- use aggregated user statistics

### LINKED ISSUE #848: Improve performance of `src users clean` in instances with many users
In instances with many users the graphQL request to get all users and compute their `lastActiveUsage` can timeout going over the network -- 

In the example below the command is used on an instance with 29 users, then pointed at an instance with 186, this shows an example of the failure.
```bash
warrengifford@Warrens-MacBook-Pro ~ % src users clean
Users to remove from instance at https://cse-k8s.sgdev.org
╭───────────────────┬──────────────────────────────────┬────────────────────────╮
│ USERNAME          │ EMAIL                            │ DAYS SINCE LAST ACTIVE │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ sourcegraph       │ cs-integrations@sourcegraph.com  │                    302 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ g-bot             │ warrenbruceg@gmail.com           │                    282 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ testAtSourcegraph │ testAtSourcegraph@mailinator.com │                    253 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ Malo              │ malo@sourcegraph.com             │                    221 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ ErikSeliger       │ erik@sourcegraph.com             │                    191 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ support           │                                  │                    132 │
├───────────────────┼──────────────────────────────────┼────────────────────────┤
│ gab2              │                                  │                     64 │
╰───────────────────┴──────────────────────────────────┴────────────────────────╯
Do you  wish to proceed with user removal [y/N]: n
Aborting removal
warrengifford@Warrens-MacBook-Pro ~ % vim ~/.zshrc
warrengifford@Warrens-MacBook-Pro ~ % source ~/.zshrc
warrengifford@Warrens-MacBook-Pro ~ % src login

✔️  Authenticated as warren on https://k8s.sgdev.org

warrengifford@Warrens-MacBook-Pro ~ % src users clean
error: 504 Gateway Timeout

<html>
<head><title>504 Gateway Time-out</title></head>
<body>
<center><h1>504 Gateway Time-out</h1></center>
<hr><center>nginx</center>
</body>
</html>
``` 

This problem was [foreseen](https://github.com/sourcegraph/src-cli/pull/826#discussion_r960392615) in the original release PR by @mrnugget 

Since then @mrnugget has added a new graphQL field that should improve performance here, and `users clean` needs to be rewritten to incorperate this more efficient query https://github.com/sourcegraph/sourcegraph/pull/42352/files

DaedalusG:
This PR refactors use of the `usageStatistics` `lastActiveUsage` table which caused the command to timeout during execution on instances with a sufficiently large number of users. See #848 

Instead the `aggregated_user_statistics` [table](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/internal/database/schema.md?subtree=true) is used via the `site` graphQL endpoint
```graphql
query getInactiveUsers {
  site {
    users {
      nodes {
        id
        username
        email
        siteAdmin
        lastActiveAt
      }
    }
  }
}
```

This has two benefits 
1. The backend to query `lastActiveAt` from the `site` graphQL endpoint is much more efficient than `usageStatistics`, and eliminates timeout concerns on many users instances
2. `aggregated_user_statistics` persists the date `lastActiveAt` eliminating the concern of `null` `lastActiveUsage` values which are misleading. `lastActiveUsage` is computed by reference to the `event_logs` table which only persists event entries for 93 days, when 93 days pass if a user had no entries in `event_logs` `lastActiveUsage` was set to `null` -- previously `src users clean` interpreted a `null` value as a user never having used the instance. _A flag `-remove-null-users` is provided to remove `null` lastActiveAt` users_

This PR also changes the name of the `src users clean` command renaming it to `src users prune` which follows the `docker` and `kubectl` cleanup command convention

Closes #848

### Test plan
Tested on local machine

### DIFF
--- a/cmd/src/users_prune.go
+++ b/cmd/src/users_prune.go
@@ -21,7 +21,7 @@ Examples:
 
 	$ src users prune -days 182
 	
-	$ src users prune -remove-admin -remove-never-active 
+	$ src users prune -remove-admin -remove-null-users
 `
 
 	flagSet := flag.NewFlagSet("prune", flag.ExitOnError)
@@ -33,7 +33,7 @@ Examples:
 	var (
 		daysToDelete       = flagSet.Int("days", 60, "Days threshold on which to remove users, must be 60 days or greater and defaults to this value ")
 		removeAdmin        = flagSet.Bool("remove-admin", false, "prune admin accounts")
-		removeNoLastActive = flagSet.Bool("remove-never-active", false, "removes users with no events registered in the database")
+		removeNoLastActive = flagSet.Bool("remove-null-users", false, "removes users with no last active value")
 		skipConfirmation   = flagSet.Bool("force", false, "skips user confirmation step allowing programmatic use")
 		apiFlags           = api.NewFlags(flagSet)
 	)
--- a/cmd/src/users_prune.go
+++ b/cmd/src/users_prune.go
@@ -19,12 +19,12 @@ This command removes users from a Sourcegraph instance who have been inactive fo
 	
 Examples:
 
-	$ src users clean -days 182
+	$ src users prune -days 182
 	
-	$ src users clean -remove-admin -remove-never-active 
+	$ src users prune -remove-admin -remove-never-active 
 `
 
-	flagSet := flag.NewFlagSet("clean", flag.ExitOnError)
+	flagSet := flag.NewFlagSet("prune", flag.ExitOnError)
 	usageFunc := func() {
 		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
 		flagSet.PrintDefaults()
@@ -32,7 +32,7 @@ Examples:
 	}
 	var (
 		daysToDelete       = flagSet.Int("days", 60, "Days threshold on which to remove users, must be 60 days or greater and defaults to this value ")
-		removeAdmin        = flagSet.Bool("remove-admin", false, "clean admin accounts")
+		removeAdmin        = flagSet.Bool("remove-admin", false, "prune admin accounts")
 		removeNoLastActive = flagSet.Bool("remove-never-active", false, "removes users with no events registered in the database")
 		skipConfirmation   = flagSet.Bool("force", false, "skips user confirmation step allowing programmatic use")
 		apiFlags           = api.NewFlags(flagSet)
--- a/cmd/src/users_prune.go
+++ b/cmd/src/users_prune.go
@@ -85,9 +85,9 @@ query getInactiveUsers {
 
 		var usersResult struct {
 			Site struct {
-			Users struct {
-				Nodes []SiteUser 
-			}
+				Users struct {
+					Nodes []SiteUser
+				}
 			}
 		}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,7 +15,11 @@ All notable changes to `src-cli` are documented in this file.
 
 ### Changed
 
+- Renamed `src users clean` command to `src users prune` [#901](https://github.com/sourcegraph/src-cli/pull/901)
+
 ### Fixed
+
+- Fix network timeout in `src users clean` occuring in instances with many users  [#901](https://github.com/sourcegraph/src-cli/pull/901)
 
 ### Removed
 
--- a/cmd/src/users.go
+++ b/cmd/src/users.go
@@ -90,3 +90,12 @@ type UserUsageStatistics struct {
 	LastActiveTime                    string
 	LastActiveCodeHostIntegrationTime string
 }
+
+type SiteUser struct {
+	ID string
+	Username string
+	Email string
+	SiteAdmin bool
+	LastActiveAt string
+}
+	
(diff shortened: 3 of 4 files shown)


cmd/src/users_prune.go
```diff
@@ -19,21 +19,21 @@ This command removes users from a Sourcegraph instance who have been inactive fo
 	
 Examples:
 
-	$ src users clean -days 182
+	$ src users prune -days 182
```
erzhtor, line 22: How are such command/API changes usually made in `src` cli? Should this be a major version increase or should we keep backward compatibility leaving both `clean` and `prune` commands and marking one deprecated?

erzhtor (approved):
Besides, a comment regarding command (and flag) backward compatibility, PR looks good!

cmd/src/users_prune.go
DaedalusG: Almost all the functionality changes in terms of the command run locally, the old `clean` command had some big timeout issues see https://github.com/sourcegraph/src-cli/issues/848. The name change from `clean` to `prune` was mostly just a style change. I don't think this command is being widely used so I think the change is alright. To my knowledge no one has put this on a cronjob or anything like that yet.

cmd/src/users_prune.go
```diff
@@ -68,43 +68,50 @@ query {
...
+	  site {
+    users {
+      nodes {
+        username
+        email
+        siteAdmin
+        lastActiveAt
+      }
+    }
+  }
```
mrnugget, line 82: Intendation is off

mrnugget (commented):
How is this different than before in its performance characteristics? You're still fetching _all_ users in the table, without doing any filtering. I don't see any improvement here?

cmd/src/users_prune.go
DaedalusG: This is hitting the `aggregated_user_statistics` table, and the [value](https://sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/cmd/frontend/graphqlbackend/site_users.go?L91-97&subtree=true) of `lastActiveAt` doesn't need to access the `event_logs` table the same way `usageStatistics` used to. Try it out [here](https://k8s.sgdev.org/api/console#%7B%22query%22%3A%22query%20InactiveUsersShort%20%7B%5Cn%20%20site%20%7B%5Cn%20%20%20%20users%20%7B%5Cn%20%20%20%20%20%20nodes%20%7B%5Cn%20%20%20%20%20%20%20%20id%5Cn%20%20%20%20%20%20%20%20username%5Cn%20%20%20%20%20%20%20%20email%5Cn%20%20%20%20%20%20%20%20siteAdmin%5Cn%20%20%20%20%20%20%20%20lastActiveAt%5Cn%20%20%20%20%20%20%7D%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%7D), its much more performant.

`aggregated_user_statistics` table is [updating in the background](https://sourcegraph.com/github.com/sourcegraph/sourcegraph@84726dc70827092c17547b6d7a35c505d8834895/-/blob/internal/users/update_aggregated_stats_job.go), so this isn't accessing the event logs table at the time of the request.

DaedalusG: https://github.com/sourcegraph/src-cli/pull/903/files Indentation fix

mrnugget:
Okay, so we're removing the N+1 error, got it.

That's only 1 of 2 possible performance issues, the other is that we transfer all users over the wire. (I don't think we want that on an instance with 10k+ users).

But I just checked and it turns out that the code here doesn't work: `site { users }` is _paginated_ and only returns the first 100 users by default. You need to paginate through the results.

Here, [run the query](https://k8s.sgdev.org/api/console#%7B%22query%22%3A%22query%20InactiveUsersShort%20%7B%5Cn%20%20site%20%7B%5Cn%20%20%20%20users%20%7B%5Cn%20%20%20%20%20%20nodes%20%7B%5Cn%20%20%20%20%20%20%20%20id%5Cn%20%20%20%20%20%20%20%20username%5Cn%20%20%20%20%20%20%20%20email%5Cn%20%20%20%20%20%20%20%20siteAdmin%5Cn%20%20%20%20%20%20%20%20lastActiveAt%5Cn%20%20%20%20%20%20%7D%5Cn%20%20%20%20%20%20totalCount%5Cn%20%20%20%20%20%20%5Cn%20%20%20%20%7D%5Cn%20%20%7D%5Cn%7D%22%2C%22operationName%22%3A%22InactiveUsersShort%22%7D) (note that I added `totalCount`), copy the output into `result.json` and then check with `jq`:

```
$ cat result.json | jq '.data.site.users.nodes | length'
100
$ cat result.json | jq '.data.site.users.totalCount'
208
```

that means you're only cleaning up the first 100 users.

DaedalusG:
Ah, good catch @mrnugget will open a follow up on this to paginate and maybe implement a filter similar to what we were doing here: https://github.com/sourcegraph/sourcegraph/pull/43370

PR #907, opened 2023-01-03
codeintel: Only upload SCIP when accepted

efritz:
This PR ensures that we only upload SCIP if the instance accepts it. This stops us from uploading LSIF which then has to be converted to SCIP in the background (breaking some implicit assumptions that oobmigrations should be non-decreasing, and not display as incompletely after hitting 100%).

Essentially we are cutting out an entire release from our backcompat guarantees by doing this in one step. We could have just removed LSIF support entirely later, but this has the same effect over time.

### Test plan

- [ ] Test by hand on a local instance (all variations; old/new instance by lsif/scip index)


go.mod
```diff
@@ -134,3 +128,5 @@ require (
 
 // See: https://github.com/ghodss/yaml/pull/65
 replace github.com/ghodss/yaml => github.com/sourcegraph/yaml v1.0.1-0.20200714132230-56936252f152
+
+replace github.com/sourcegraph/sourcegraph/lib => /Users/efritz/dev/sourcegraph/sourcegraph/lib
```
efritz, line 132: NO YOU BOZO

Numbers88s (approved):

PR #922, opened 2023-01-12
Teams CLI interface

eseliger:
This implements a CLI interface for managing/ingesting teams into Sourcegraph.

This still requires https://github.com/sourcegraph/sourcegraph/pull/47143 on the backend but otherwise is functional, so opening up for review to get some feedback on it and I will make sure to re-test once that PR landed. (Also removing team members is not yet implemented in the backend so that also has to wait for a bit to test)

Closes https://github.com/sourcegraph/sourcegraph/issues/48228

### Test plan

Manually verified things work with an ingestion script I'm still working on to shove over teams from GitHub into our instance.


cmd/src/users_get.go
```diff
@@ -26,6 +28,7 @@ Examples:
 	}
 	var (
 		usernameFlag = flagSet.String("username", "", `Look up user by username. (e.g. "alice")`)
+		emailFlag    = flagSet.String("email", "", `Look up user by email. (e.g. "alice@sourcegraph.com")`)
```
eseliger, line 31: extended to also allow getting by email, which is useful for ingestion scripts.

cmd/src/teams_members_add.go
```diff
@@ -0,0 +1,115 @@
+package main
```
eseliger, line 1: this requires https://github.com/sourcegraph/sourcegraph/pull/47143 to work which I'm intending to work on tomorrow.

cmd/src/teams_members_add.go
```diff
@@ -0,0 +1,115 @@
...
+	"github.com/sourcegraph/src-cli/internal/cmderrors"
+)
+
+func init() {
+	usage := `
+Examples:
+
+  Add a team member:
+
+    	$ src teams members add -team-name='engineering' [-email='alice@sourcegraph.com'] [-username='alice'] [-id='VXNlcjox'] [-external-account-service-id='https://github.com/' -external-account-service-type='github' [-external-account-account-id='123123123'] [-external-account-login='alice']]
```
eseliger, line 19: TODO: I would love if this supported adding multiple users at once, like the API already supports, but I didn't find a nice way in command flags to specify multiple users, as all of these flags might relate to the same user, but also might now. If someone has ideas, I'm all ears :)

leonore (approved):
thanks for doing this, lots of painful boilerplate!! logic looks good to me overall

PR #923, opened 2023-01-12
batches: skip empty log output

BolajiOlajide:
### Test plan

I added a unit test for this.

### DIFF
--- a/internal/batches/ui/interval_writer.go
+++ b/internal/batches/ui/interval_writer.go
@@ -3,6 +3,7 @@ package ui
 import (
 	"bytes"
 	"context"
+	"fmt"
 	"io"
 	"time"
 
@@ -118,10 +119,15 @@ type prefixedWriter struct {
 
 func (w *prefixedWriter) Write(p []byte) (int, error) {
 	var prefixedLines []byte
+	fmt.Println("=====================")
+	fmt.Println("'", string(p), "'")
+	fmt.Println("=====================")
 	for _, line := range bytes.Split(p, []byte("\n")) {
-		// We run into a weird situation where the output from a docker container
-		// contains an extra new line with spaces. This results in n*2 times the log
-		// output expected from an execution. We use this check to remove such blank lines
+		// When we split a byte slice on every new line, we get atleast two sub-slice.
+		// I'll use a string to demonstrate, if I have a string "Peppermint\n", splitting
+		// this string on a new line will return a slice with two items: ["Peppermint", ""].
+		// Using this logic, we can then skip empty lines from being added to the log output.
+		// SideNote: This means when one logs an actual empty line, we'll also discard that.
 		if len(bytes.TrimSpace(line)) == 0 {
 			continue
 		}
--- a/internal/batches/ui/interval_writer.go
+++ b/internal/batches/ui/interval_writer.go
@@ -116,23 +116,17 @@ type prefixedWriter struct {
 	prefix     string
 }
 
-func (w *prefixedWriter) Write(p []byte) (int, error) {
-	var prefixedLines []byte
-	for _, line := range bytes.Split(p, []byte("\n")) {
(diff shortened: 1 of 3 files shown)


internal/batches/ui/interval_writer.go
```diff
@@ -119,6 +119,14 @@ type prefixedWriter struct {
 func (w *prefixedWriter) Write(p []byte) (int, error) {
```
eseliger, line 119: I just checked the source of `p` and I think we will only ever stream 1 line at a time, either with the newline at the end, or not if the stream has ended and there is no newline (for example `echo -n`).
So that makes me thing that we shouldn't even split the `p` in here at all - and would as a side-effect get that for free, while retaining _actual_ empty log lines. How does that feel for you?

BolajiOlajide: Oh! Yeah true - that's better. I can just remove the loop then. It makes sense to me.

eseliger: Should just re-validate that - but from reading code it seems like our scanner on the other end of this stream would write in line-chunks :)

BolajiOlajide: Yeah, I just tested it, and it works. For both SSBC and local executions - the log output is consistent now.
Pushed a refactor.

internal/batches/ui/interval_writer.go
```diff
@@ -116,15 +116,17 @@ type prefixedWriter struct {
...
-	w.writes <- prefixedLines
+func (w *prefixedWriter) Write(p []byte) (int, error) {
+	var prefixedLine []byte
+	// We remove new lines appended to the log output so all of the stream can have
+	// only one new line separating them. This ensures the previous behavior for the log
+	// structure is preserved.
+	p = bytes.TrimRight(p, "\n")
+	prefixedLine = append([]byte(w.prefix), p...)
+	prefixedLine = append(prefixedLine, newLineByteSlice...)
+	w.writes <- prefixedLine
```
eseliger, line 129: This could be simplified

```suggestion
	// We remove new lines appended to the log output so all of the stream can have
	// only one new line separating them. This ensures the previous behavior for the log
	// structure is preserved.
	p = bytes.TrimRight(p, "\n")
	prefixedLine := append([]byte(w.prefix), p...)
	prefixedLine = append(prefixedLine, newLineByteSlice...)
	w.writes <- prefixedLine
```

eseliger: Maybe this gives a small perf-boost but non-blocking:

```suggestion
	prefixedLine := append([]byte(w.prefix), p...)
	if prefixedLine[len(prefixedLine)-1] != newLineByteSlice {
		prefixedLine = append(prefixedLine, newLineByteSlice...)
    }
	w.writes <- prefixedLine
```

internal/batches/ui/interval_writer.go
```diff
@@ -116,15 +116,17 @@ type prefixedWriter struct {
...
-	var prefixedLines []byte
-	for _, line := range bytes.Split(p, []byte("\n")) {
-		prefixedLine := append([]byte(w.prefix), line...)
-		prefixedLine = append(prefixedLine, []byte("\n")...)
+var newLineByteSlice = []byte("\n")
 
-		prefixedLines = append(prefixedLines, prefixedLine...)
-	}
-	w.writes <- prefixedLines
+func (w *prefixedWriter) Write(p []byte) (int, error) {
```
eseliger, line 121:
```suggestion
// Write is only ever called with a single line. That line may or may not end with a newline character.
// It then writes the content as a single line to the inner writer, regardless of if the provided line
// had a newline or not. That is, because our encoding requires exactly one line per formatted line.
func (w *prefixedWriter) Write(p []byte) (int, error) {
```

eseliger (approved):
Good stuff!

PR #929, opened 2023-01-26
batches: query max unlicensed changesets count

### LINKED ISSUE #913: batches: remove hardcoded max unlicensed changesets from src-cli TUI output
Currently, the limit of 10 changesets per batch change for unlicensed/free tier Sourcegraph instances is [hardcoded](https://sourcegraph.com/github.com/sourcegraph/src-cli/-/blob/internal/batches/ui/tui.go?L331) into the TUI. There is no reason to assume this value will be static moving forward, and hardcoding it here means we are maintaining multiple sources of truth for the current static limit.

There are two ways we could address this:

- See how hard it is to just return a clear, legible error that includes the license limit when we can't create a batch spec due to hitting this limit. This would be the preferred method and would enable us to delete that extra TUI method entirely, but I'm not sure how easy it would be.
- Query `maxUnlicensedChangeset` in the mutation where we create the batch spec and pass that to the TUI method, too. If we do still require special error handling here to make it clear it's a licensing error, this way the TUI method could use the value from the backend instead of its hardcoded one.

adeola-ak:
Closes #913

Change the 10 changeset limit for batch changes for unlicensed/free tier Sourcegraph instances from being hardcoded, to querying `maxUnlicensedChangeset` in the mutation where we create the batch spec so that error codes`ErrBatchChangesUnlicensed` and `ErrBatchChangesOverLimit` returns value

CleanShot 2023-01-26 at 19 46 52@2x

### Test plan

Test creating a batch change with more changesets than the current `maxUnlicensedChangesets` values as an unlicensed user


internal/batches/service/service.go
```diff
@@ -43,10 +43,12 @@ func New(opts *Opts) *Service {
 // The reason we ask for batchChanges here is to surface errors about trying to use batch
 // changes in an unsupported environment sooner, since the version check is typically the
 // first thing we do.
-const sourcegraphVersionQuery = `query SourcegraphVersion {
+
+const getInstanceInfo = `query SourcegraphVersion {
```
courier-new, line 47: Should probably match the query name:

```suggestion
const getInstanceInfo = `query InstanceInfo {
```

internal/batches/service/service.go
```diff
@@ -43,10 +43,12 @@ func New(opts *Opts) *Service {
...
 	site {
 		productVersion
 	}
+		maxUnlicensedChangesets
```
courier-new, line 51:
```suggestion
	maxUnlicensedChangesets
```

internal/batches/service/service.go
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
 
 // getSourcegraphVersion queries the Sourcegraph GraphQL API to get the
 // current version of the Sourcegraph instance.
```
courier-new, line 61: Can we update this docstring?

internal/batches/service/service.go
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
-func (svc *Service) getSourcegraphVersion(ctx context.Context) (string, error) {
+
+func (svc *Service) getSourcegraphVersionAndMaxChangesetsCount(ctx context.Context) (string, int, error) {
 	var result struct {
-		Site struct {
+		maxUnlicensedChangesets int
```
courier-new, line 65: Convention is usually to capitalize the start of property names, too:

```suggestion
		MaxUnlicensedChangesets int
```

internal/batches/service/service.go
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
 
-	ok, err := svc.client.NewQuery(sourcegraphVersionQuery).Do(ctx, &result)
+	ok, err := svc.client.NewQuery(getInstanceInfo).Do(ctx, &result)
 	if err != nil || !ok {
-		return "", err
+		return "", 0, err
 	}
 
-	return result.Site.ProductVersion, err
+	//add to return
```
courier-new, line 76:
```suggestion
```

internal/batches/service/service.go
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
+	return result.Site.ProductVersion, result.maxUnlicensedChangesets, err
```
courier-new, line 77:
```suggestion
	return result.Site.ProductVersion, result.MaxUnlicensedChangesets, err
```

internal/batches/service/service.go
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
 }
 
-// DetermineFeatureFlags fetches the version of the configured Sourcegraph and
+// DetermineFeatureFlagsAndLicense fetches the version of the configured Sourcegraph and
 // returns the enabled features.
```
courier-new, line 81: Maybe:

```suggestion
// DetermineFeatureFlagsAndLicense returns the enabled features and license restrictions
// configured for the Sourcegraph instance.
```
```

File: internal/batches/service/service.go [hunk 8]
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
-func (svc *Service) DetermineFeatureFlags(ctx context.Context) (*batches.FeatureFlags, error) {
-	version, err := svc.getSourcegraphVersion(ctx)
+
+func (svc *Service) DetermineFeatureFlagsAndLicense(ctx context.Context) (*batches.LicenseRestrictions, *batches.FeatureFlags, error) {
```
courier-new, line 83: Nit: We should probably either call the function `DetermineLicenseAndFeatureFlags` or flip the order of the returned items: `(*batches.FeatureFlags, *batches.LicenseRestrictions, error)` so that they're self-consistent.

internal/batches/service/service.go
```diff
@@ -57,30 +59,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
+
+	lr := &batches.LicenseRestrictions{
+		MaxUnlicensedChangesets: mc,
 	}
+
 	ffs := &batches.FeatureFlags{}
-	return ffs, ffs.SetFromVersion(version)
+
+	return lr, ffs, ffs.SetFromVersion(version)
+
```
courier-new, line 96: It's so spacious in here lol 😜

```suggestion
	return lr, ffs, ffs.SetFromVersion(version)
```

internal/batches/service/service.go
```diff
@@ -120,6 +133,7 @@ mutation CreateBatchSpec(
     ) {
         id
         applyURL
+
```
courier-new, line 136: 👻
```suggestion
```

internal/batches/ui/exec_ui.go
```diff
@@ -40,7 +40,7 @@ type ExecUI interface {
 
 	CreatingBatchSpec()
 	CreatingBatchSpecSuccess(previewURL string)
-	CreatingBatchSpecError(err error) error
+	CreatingBatchSpecError(lr int, err error) error
```
courier-new, line 43: Since here it's the raw int, maybe:

```suggestion
	CreatingBatchSpecError(maxUnlicensedCS int, err error) error
```

internal/batches/ui/json_lines.go
```diff
@@ -149,7 +149,7 @@ func (ui *JSONLines) CreatingBatchSpecSuccess(batchSpecURL string) {
 	})
 }
 
-func (ui *JSONLines) CreatingBatchSpecError(err error) error {
+func (ui *JSONLines) CreatingBatchSpecError(lr int, err error) error {
```
courier-new, line 152:
```suggestion
func (ui *JSONLines) CreatingBatchSpecError(maxUnlicensedCS int, err error) error {
```

internal/batches/ui/tui.go
```diff
@@ -217,8 +217,8 @@ func (ui *TUI) CreatingBatchSpecSuccess(previewURL string) {
 	batchCompletePending(ui.pending, "Creating batch spec on Sourcegraph")
 }
 
-func (ui *TUI) CreatingBatchSpecError(err error) error {
-	return prettyPrintBatchUnlicensedError(ui.Out, err)
+func (ui *TUI) CreatingBatchSpecError(lr int, err error) error {
+	return prettyPrintBatchUnlicensedError(ui.Out, lr, err)
```
courier-new, line 221:
```suggestion
func (ui *TUI) CreatingBatchSpecError(maxUnlicensedCS int, err error) error {
	return prettyPrintBatchUnlicensedError(ui.Out, maxUnlicensedCS, err)
```

internal/batches/ui/tui.go
```diff
@@ -307,7 +307,7 @@ func (ui *TUI) RemoteSuccess(url string) {
 // is, then a better message is output. Regardless, the return value of this
 // function should be used to replace the original error passed in to ensure
 // that the displayed output is sensible.
-func prettyPrintBatchUnlicensedError(out *output.Output, err error) error {
+func prettyPrintBatchUnlicensedError(out *output.Output, lr int, err error) error {
```
courier-new, line 310:
```suggestion
func prettyPrintBatchUnlicensedError(out *output.Output, maxUnlicensedCS int, err error) error {
```

internal/batches/ui/tui.go
```diff
@@ -321,19 +321,19 @@ func prettyPrintBatchUnlicensedError(out *output.Output, err error) error {
...
-			} else if code == "ErrBatchChangesUnlicensed" {
+			} else if code == "ErrBatchChangesUnlicensed" || code == "ErrBatchChangesOverLimit" {
 				// OK, let's print a better message, then return an
 				// exitCodeError to suppress the normal automatic error block.
 				// Note that we have hand wrapped the output at 80 (printable)
 				// characters: having automatic wrapping some day would be nice,
 				// but this should be sufficient for now.
 				block := out.Block(output.Line("🪙", output.StyleWarning, "Batch Changes is a paid feature of Sourcegraph. All users can create sample"))
-				block.WriteLine(output.Linef("", output.StyleWarning, "batch changes with up to 10 changesets without a license. Contact Sourcegraph"))
+				block.WriteLine(output.Linef("", output.StyleWarning, "batch changes with up to %s changesets without a license. Contact Sourcegraph", lr))
```
courier-new, line 331:
```suggestion
				block.WriteLine(output.Linef("", output.StyleWarning, "batch changes with up to %s changesets without a license. Contact Sourcegraph", maxUnlicensedCS))
```

internal/batches/ui/tui.go
```diff
@@ -321,19 +321,19 @@ func prettyPrintBatchUnlicensedError(out *output.Output, err error) error {
...
 				block.WriteLine(output.Linef("", output.StyleWarning, "sales at %shttps://about.sourcegraph.com/contact/sales/%s to obtain a trial", output.StyleSearchLink, output.StyleWarning))
 				block.WriteLine(output.Linef("", output.StyleWarning, "license."))
 				block.Write("")
-				block.WriteLine(output.Linef("", output.StyleWarning, "To proceed with this batch change, you will need to create 5 or fewer"))
-				block.WriteLine(output.Linef("", output.StyleWarning, "changesets. To do so, you could try adding %scount:5%s to your", output.StyleSearchAlertProposedQuery, output.StyleWarning))
+				block.WriteLine(output.Linef("", output.StyleWarning, "To proceed with this batch change, you will need to create %s or fewer", lr))
+				block.WriteLine(output.Linef("", output.StyleWarning, "changesets. To do so, you could try adding %scount:%s%s to your", output.StyleSearchAlertProposedQuery, lr, output.StyleWarning))
```
courier-new, line 336:
```suggestion
				block.WriteLine(output.Linef("", output.StyleWarning, "To proceed with this batch change, you will need to create %s or fewer", maxUnlicensedCS))
				block.WriteLine(output.Linef("", output.StyleWarning, "changesets. To do so, you could try adding %scount:%s%s to your", output.StyleSearchAlertProposedQuery, maxUnlicensedCS, output.StyleWarning))
```

courier-new (commented):
Approach looks really solid to me! Sorry my review is basically a million nits to improve code quality a bit, but hopefully they make it easier for you to update things. 🙂

Would you mind... (I bet you can guess what I'm about to ask! 😜) ...sharing a screenshot of the new error, too?

adeola-ak:
@courier-new i asked for a review and then i removed it when i realized i had some mistakes but thank you for your prompt and helpful review!

internal/batches/ui/json_lines.go
```diff
@@ -149,7 +149,7 @@ func (ui *JSONLines) CreatingBatchSpecSuccess(batchSpecURL string) {
 	})
 }
 
-func (ui *JSONLines) CreatingBatchSpecError(lr int, err error) error {
+func (ui *JSONLines) CreatingBatchSpecError(maxUnlicensedCS int, err error) error {
```
BolajiOlajide, line 152:
```suggestion
func (ui *JSONLines) CreatingBatchSpecError(_ int, err error) error {
```

Since aren't making use of this variable and we are only passing it here to satisfy the interface we can safely ignore it using the underscore. That way the linter doesn't complain about this.

internal/batches/service/service.go
```diff
@@ -43,10 +43,12 @@ func New(opts *Opts) *Service {
...
 // changes in an unsupported environment sooner, since the version check is typically the
 // first thing we do.
-const sourcegraphVersionQuery = `query SourcegraphVersion {
+
+const getInstanceInfo = `query InstanceInfo {
 	site {
 		productVersion
 	}
+	maxUnlicensedChangesets
 	batchChanges(first: 1) {
```
BolajiOlajide, line 52: Nit: It seems we don't use this fields so we should be able to remove it.

BolajiOlajide: I know its not related to your work here but its not used so I thought to call it out.

internal/batches/service/service.go
```diff
@@ -55,32 +57,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
-func (svc *Service) DetermineFeatureFlags(ctx context.Context) (*batches.FeatureFlags, error) {
-	version, err := svc.getSourcegraphVersion(ctx)
+// DetermineFeatureFlagsAndLicense returns the enabled features and license restrictions
+// configured for the Sourcegraph instance.
+
+func (svc *Service) DetermineFeatureFlagsAndLicense(ctx context.Context) (*batches.LicenseRestrictions, *batches.FeatureFlags, error) {
+	version, mc, err := svc.getSourcegraphVersionAndMaxChangesetsCount(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to query Sourcegraph version to check for available features")
+		return nil, nil, errors.Wrap(err, "failed to query Sourcegraph version and license info for instance capabilities")
```
BolajiOlajide, line 85:
```suggestion
		return nil, nil, errors.Wrap(err, "failed to query Sourcegraph version and license info for instance")
```

I think this reads better. Just a suggestion.

internal/batches/service/service.go
courier-new: It's mentioned in the docstring up above, I think we use this as a hack to get better error feedback if you try to run a batch change on dotcom or another unsupported environment.

BolajiOlajide: Oh! I missed that. Thanks

BolajiOlajide (approved):

internal/batches/service/service.go
```diff
@@ -43,10 +43,12 @@ func New(opts *Opts) *Service {
 // The reason we ask for batchChanges here is to surface errors about trying to use batch
 // changes in an unsupported environment sooner, since the version check is typically the
 // first thing we do.
-const sourcegraphVersionQuery = `query SourcegraphVersion {
+
```
courier-new, line 46:
```suggestion

```

I promise this is actually relevant. 😜 The extra line between the docstring and the variable definition breaks inline documentation on hover for this variable. 😱

Without the extra new line:
image

With the extra new line:
image

internal/batches/service/service.go
```diff
@@ -55,32 +57,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
 }
 `
 
-// getSourcegraphVersion queries the Sourcegraph GraphQL API to get the
-// current version of the Sourcegraph instance.
-func (svc *Service) getSourcegraphVersion(ctx context.Context) (string, error) {
+// getSourcegraphVersionAndMaxChangesetsCount queries the Sourcegraph GraphQL API to get the
+// current version and max unlicensed changesets count for the Sourcegraph instance.
+
```
courier-new, line 62:
```suggestion

```

Same deal here.

internal/batches/service/service.go
```diff
@@ -55,32 +57,41 @@ const sourcegraphVersionQuery = `query SourcegraphVersion {
...
+	return result.Site.ProductVersion, result.MaxUnlicensedChangesets, err
 }
 
-// DetermineFeatureFlags fetches the version of the configured Sourcegraph and
-// returns the enabled features.
-func (svc *Service) DetermineFeatureFlags(ctx context.Context) (*batches.FeatureFlags, error) {
-	version, err := svc.getSourcegraphVersion(ctx)
+// DetermineFeatureFlagsAndLicense returns the enabled features and license restrictions
+// configured for the Sourcegraph instance.
+
```
courier-new, line 81:
```suggestion

```

Once more!

courier-new (approved):
Just made a couple suggestions to preserve inline function documentation, otherwise 👍

internal/batches/service/service.go
courier-new: Ugh, okay, the diff view really makes it difficult to understand what my suggestion is, lol.

Currently you have this:
```go
// The reason we ask for batchChanges here is to surface errors about trying to use batch
// changes in an unsupported environment sooner, since the version check is typically the
// first thing we do.

const sourcegraphVersionQuery = `query SourcegraphVersion {
...
```

My suggestion here and in the following two cases is to remove the extra line between the comment and the variable, which preserve the inline documentation on hover:

```go
// The reason we ask for batchChanges here is to surface errors about trying to use batch
// changes in an unsupported environment sooner, since the version check is typically the
// first thing we do.
const sourcegraphVersionQuery = `query SourcegraphVersion {
...
```

PR #938, opened 2023-02-16
servegit: use more performant WalkDir

keegancsmith:
WalkDir was introduced in go1.16 suits our use case while being more efficient.

Test Plan: go test

### DIFF
--- a/internal/servegit/serve.go
+++ b/internal/servegit/serve.go
@@ -5,10 +5,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"html/template"
+	"io/fs"
 	"log"
 	"net"
 	"net/http"
-	"os"
 	"os/exec"
 	pathpkg "path"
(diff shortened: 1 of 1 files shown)


internal/servegit/serve.go
```diff
@@ -170,7 +170,7 @@ func (s *Serve) Repos() ([]Repo, error) {
 		return nil, nil
 	}
 
-	err = filepath.Walk(root, func(path string, fi os.FileInfo, fileErr error) error {
+	err = filepath.WalkDir(root, func(path string, fi fs.DirEntry, fileErr error) error {
```
mrnugget, line 173: Do you still need the `if !fi.IsDir()` below then?

keegancsmith: yes walkdir still goes over files.

mrnugget (approved):

PR #939, opened 2023-02-16
src validation for ebs-csi drivers and VPCs

jasonhawkharris:
This commit adds 
- initial support for validating Sourcegraph on EKS clusters
- validation for `ebs-csi` drivers on AWS EKS clusters and
- validation for configured VPCs for AWS
 
### Test plan
Shipped with test suite


internal/validate/kube/utils.go
```diff
@@ -0,0 +1,49 @@
+package kube
```
jdpleiness, line 1: I would rename this file to `eks.go`. Using `utils` or `common` in go is considered a[ code smell ](https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common) (more so for package names) and these functions are used for EKS only anyhow.

internal/validate/kube/kube.go
```diff
@@ -50,41 +57,81 @@ func Quiet() Option {
...
-		{Services, "validating services", "services validated", "validating services failed"},
-		{PVCs, "validating pvcs", "pvcs validated", "validating pvcs failed"},
-		{Connections, "validating connections", "connections validated", "validating connections failed"},
+	var validations []validationGroup
+	validations = append(validations, validationGroup{Pods, "validating pods", "pods validated", "validating pods failed"})
+	validations = append(validations, validationGroup{Services, "validating services", "services validated", "validating services failed"})
+	validations = append(validations, validationGroup{PVCs, "validating pvcs", "pvcs validated", "validating pvcs failed"})
+	validations = append(validations, validationGroup{Connections, "validating connections", "connections validated", "validating connections failed"})
+
+	if cfg.eks == true {
```
jdpleiness, line 110:
```suggestion
	if cfg.eks {
```

internal/validate/kube/kube.go
```diff
@@ -387,3 +434,106 @@ func Connections(ctx context.Context, config *Config) ([]validate.Result, error)
 
 	return results, nil
 }
+
```
jdpleiness, line 437: EKS functions from here down should be moved to `eks.go` file as well

internal/validate/kube/kube.go
```diff
@@ -387,3 +434,106 @@ func Connections(ctx context.Context, config *Config) ([]validate.Result, error)
...
+		return results, err
+	}
+
+	r := validateEbsCsiDrivers(&outputs.Addons)
+	results = append(results, r...)
+
+	return results, nil
+}
+
+func validateEbsCsiDrivers(addons *[]string) (result []validate.Result) {
```
jdpleiness, line 469: Is a pointer to a slice needed here?

internal/validate/kube/utils.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"strings"
+
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/client-go/util/homedir"
+)
+
+func Contains(sl *[]string, t string) bool {
```
jdpleiness, line 13: This probably doesn't need to a be a function as it's only used once. I would move the logic inline to where it is called.

internal/validate/kube/kube.go
```diff
@@ -50,41 +57,81 @@ func Quiet() Option {
...
-		{Connections, "validating connections", "connections validated", "validating connections failed"},
+	var validations []validationGroup
+	validations = append(validations, validationGroup{Pods, "validating pods", "pods validated", "validating pods failed"})
+	validations = append(validations, validationGroup{Services, "validating services", "services validated", "validating services failed"})
+	validations = append(validations, validationGroup{PVCs, "validating pvcs", "pvcs validated", "validating pvcs failed"})
+	validations = append(validations, validationGroup{Connections, "validating connections", "connections validated", "validating connections failed"})
+
+	if cfg.eks == true {
+		if !CurrentContextSetToEKSCluster() {
+			return fmt.Errorf("🛑 set current context to EKS cluster to use --eks flag")
```
jdpleiness, line 112: Use [Sourcegraph errors lib ](https://docs.sourcegraph.com/dev/background-information/languages/go_errors)here instead of `fmt.Errorf`

```suggestion
return errors.Newf("%s no EKS cluster detected in current context", validate.FailureEmoji)
```

internal/validate/kube/utils.go
```diff
@@ -0,0 +1,49 @@
...
+	for _, s := range *sl {
+		if s == t {
+			return true
+		}
+	}
+	return false
+}
+
+// checks if current context is set to EKS cluster
```
jdpleiness, line 22: Comment should start with function name in this case, but this likely doesn't need to be exported anyhow

internal/validate/kube/utils.go
```diff
@@ -0,0 +1,49 @@
...
+func CurrentContextSetToEKSCluster() bool {
```
jdpleiness, line 23: I would change this func sig to return a bool or an error

internal/validate/kube/utils.go
```diff
@@ -0,0 +1,49 @@
...
+	pathToKubeConfig := filepath.Join(home, ".kube", "config")
+
+	config, err := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(
+		&clientcmd.ClientConfigLoadingRules{ExplicitPath: pathToKubeConfig},
+		&clientcmd.ConfigOverrides{
+			CurrentContext: "",
+		}).RawConfig()
+
+	if err != nil {
+		fmt.Printf("error while checking current context: %s\n", err)
```
jdpleiness, line 34: With the new func sig, you can now return this error and handle it in the calling function instead of printing it out.

internal/validate/kube/kube.go
```diff
@@ -50,41 +57,81 @@ func Quiet() Option {
...
+	}
+
+	return func(config *Config) {
+		config.eks = true
+		config.eksClient = eks.NewFromConfig(eksConfig)
+		config.ec2Client = ec2.NewFromConfig(ec2Config)
+	}
+}
+
 type validation func(ctx context.Context, config *Config) ([]validate.Result, error)
```
jdpleiness, line 78: Essentially here the validation type has changed. `validationGroup` isn't so much a group of validations here, as it is the new type needed. Therefor I would change `validation` to something like the following:

```go
type validation struct {
	Function   func(ctx context.Context, config *Config) ([]validate.Result, error)
	WaitMsg    string
	SuccessMsg string
	ErrMsg     string
}
```

jdpleiness (changes requested):
This looks great overall! I think it's going to be really helpful as well 🚀

internal/validate/kube/kube.go
jasonhawkharris: Moved!

internal/validate/kube/utils.go
jasonhawkharris: Done!

internal/validate/kube/kube.go
```diff
@@ -1,389 +1,441 @@
...
+        opt(cfg)
+    }
+
+    log.SetOutput(cfg.output)
+
+    var validations []validation
+    validations = append(validations, validation{Pods, "validating pods", "pods validated", "validating pods failed"})
+    validations = append(validations, validation{Services, "validating services", "services validated", "validating services failed"})
+    validations = append(validations, validation{PVCs, "validating pvcs", "pvcs validated", "validating pvcs failed"})
+    validations = append(validations, validation{Connections, "validating connections", "connections validated", "validating connections failed"})
```
jac, line 113: nit: I think this would be more idiomatic Go
```suggestion
 validations := []validation{
    {Pods, "validating pods", "pods validated", "validating pods failed"},
    {Services, "validating services", "services validated", "validating services failed"},
    {PVCs, "validating pvcs", "pvcs validated", "validating pvcs failed"},
    {Connections, "validating connections", "connections validated", "validating connections failed"},
}
```

jac (approved):
LGTM!

internal/validate/kube/eks.go
```diff
@@ -0,0 +1,351 @@
...
+	want := []string{"arn", "aws", "eks"}
+
+	if len(got) >= 3 {
+		got = got[:3]
+		if reflect.DeepEqual(got, want) {
+			return nil
+		}
+	}
+
+	return errors.Newf("%s no eks cluster configured", validate.FailureEmoji)
```
jdpleiness, line 254:
```suggestion
return errors.New("no eks cluster configured")
```
Take out emoji here as it's added by the calling function when returning an error

internal/validate/kube/kube.go
```diff
@@ -50,41 +58,86 @@ func Quiet() Option {
...
+	ec2Config, err := config.LoadDefaultConfig(context.TODO())
+	if err != nil {
+		log.Printf("error while loading config: %s\n", err)
+	}
+
+	iamConfig, err := config.LoadDefaultConfig(context.TODO())
+	if err != nil {
+		log.Printf("error while loading config: %s\n", err)
+	}
+
```
jdpleiness, line 76: I would move these to a separate function in `eks.go`, then call that function in the nested `if cfg.eks` block below. This way you can use the context already passed to `Validate` instead of the `TODO` context. Since `config` is passed around as a reference, you should still have no issue storing the configs for eks, ec2, and iam in the `Config` struct.

jdpleiness (approved):
This is looking really nice! Just a few suggestions, otherwise LGTM 🚀

internal/validate/kube/kube.go
jasonhawkharris: Done!

PR #942, opened 2023-02-21
permissions-center: use new `permissionSyncJobs` query.

sashaostrikov:
Test plan:
`src snapshot test` runs successfully and fetches perms sync information from a local sg instance.

Follow-up to https://github.com/sourcegraph/sourcegraph/pull/47933

### DIFF
--- a/internal/instancehealth/checks.go
+++ b/internal/instancehealth/checks.go
@@ -139,7 +139,7 @@ func checkPermissionsSyncing(
 			continue
 		}
 		syncCount += 1
-		if sync.State == "ERROR" {
+		if sync.State == "ERROR" || sync.State == "FAILED" {
 			syncErrors = append(syncErrors, sync.FailureMessage)
 		}
 		for _, p := range sync.CodeHostStates {
@@ -149,9 +149,7 @@ func checkPermissionsSyncing(
 			}
 			// Just track one message per state for reference
 			seenProviders[key][p.Status] = p.Message
-			out.WriteLine(output.Linef(output.EmojiInfo, output.StyleBold, "%s %s %s %s", p.ProviderID, p.ProviderType, p.Message, p.Status))
 		}
-		out.WriteLine(output.Linef(output.EmojiInfo, output.StyleBold, "%s %s %s", sync.State, sync.FailureMessage, sync.FinishedAt.String()))
 	}
 
 	if syncCount == 0 {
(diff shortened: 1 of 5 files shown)


internal/instancehealth/checks.go
```diff
@@ -134,22 +134,24 @@ func checkPermissionsSyncing(
...
-	for _, sync := range instanceHealth.PermissionsSyncJobs.Nodes {
-		if sync.CompletedAt.Before(time.Now().Add(-since)) {
+	for _, sync := range instanceHealth.PermissionSyncJobs.Nodes {
+		if sync.FinishedAt.Before(time.Now().Add(-since)) {
 			continue
 		}
 		syncCount += 1
-		if sync.Status == "ERROR" {
-			syncErrors = append(syncErrors, sync.Message)
+		if sync.State == "ERROR" {
```
thenamankumar, line 142: sync.State == "ERROR" || sync.State == "FAILURE"

internal/instancehealth/checks.go
```diff
@@ -134,22 +134,24 @@ func checkPermissionsSyncing(
...
-		for _, p := range sync.Providers {
-			key := fmt.Sprintf("%s - %s", p.Type, p.ID)
+		for _, p := range sync.CodeHostStates {
+			key := fmt.Sprintf("%s - %s", p.ProviderType, p.ProviderID)
 			if _, ok := seenProviders[key]; !ok {
 				seenProviders[key] = make(map[string]string)
 			}
 			// Just track one message per state for reference
 			seenProviders[key][p.Status] = p.Message
+			out.WriteLine(output.Linef(output.EmojiInfo, output.StyleBold, "%s %s %s %s", p.ProviderID, p.ProviderType, p.Message, p.Status))
```
thenamankumar, line 152: Intentional?

thenamankumar (approved):

internal/instancehealth/checks.go
sashaostrikov: it was not the part of original logic, but I agree.

@bobheadxi ping, just in case :)

internal/instancehealth/checks.go
sashaostrikov: whoops, leftover debug stuff, thanks for catching!

internal/instancehealth/checks.go
bobheadxi: +1, the `FAILURE` state is new IIRC so let's include it :)

bobheadxi (approved):
thank you!!

PR #945, opened 2023-03-01
batch_exec: inject docker container registry as a prefix

danieldides:
Part of https://github.com/sourcegraph/customer/issues/1872

This checks for the presence of an optional environment variable passed to the `src batch exec` directive that contains a docker registry URL. If it's present, it prefixes the container name with the registry information to create a fully-qualified domain name for all subsequent docker commands.

### Test plan
WIP

### DIFF
--- a/cmd/src/batch_exec.go
+++ b/cmd/src/batch_exec.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/sourcegraph/src-cli/internal/batches/docker"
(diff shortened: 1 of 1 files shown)


cmd/src/batch_exec.go
```diff
@@ -174,6 +175,28 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
...
+
+	// If the docker prefix registry is set, inject it into each step's container
+	// to fully qualify the container name
+	environ := map[string]string{}
+	for _, e := range globalEnv {
+		pair := strings.SplitN(e, "=", 2)
+		environ[pair[0]] = pair[1]
+	}
+
+	dockerPrefix, ok := environ["DOCKER_PREFIX_REGISTRY_URL"]
```
eseliger, line 192: could we instead make this a flag that the server passes? Or do we want to configure this per-executor?
That way we wouldn't need to parse the env here.

cmd/src/batch_exec.go
```diff
@@ -174,6 +175,28 @@ func executeBatchSpecInWorkspaces(ctx context.Context, flags *executorModeFlags)
...
+	if ok {
+		for i := 0; i < len(task.Steps); i++ {
+			step := &task.Steps[i]
+			step.Container = fmt.Sprintf("%s/%s", dockerPrefix, step.Container)
```
eseliger, line 196: Changing the step might change the cache key, which which would case a mismatch between backend and src-cli 🤔 I'd need to validate that, but if that's the case we should only change it when we actually use the images, not on the step element itself.

eseliger: If you want to validate that yourself, you'd want to run a spec with 2 steps, execute, change the 2nd step command or whatever, and see if a re-execution would hit a cache result for step1 and then if src-cli correctly sees it and skips step 1 in execution.

PR #947, opened 2023-03-02
Add Snapshot Restore command

jac:
Snapshot Restore generates commands to restore dumped snapshots to a Sourcegraph instance

Usage:
After generating a dump the restore commands can be generated with
```
src snapshot restore <docker|kubectlpg_dump>
```
The `--run` command can be used to attempt to automatically run these commands. Useful e.g. in the AMIs e.g.
```
src snapshot restore --run <docker|kubectlpg_dump>
```

- Changed snapshot args ordering
  - To get the --run flag to work it was necessary to change arg ordering as Go's flag package requires all flags come before non flag args
```
src snapshot databases docker --run
# changes to
src snapshot databases --run docker
```
- Removed the -t flags from <docker|kubectl> exec
  - afaik only -i should be necessary for input/output redirection. Without -t we get better compatibility (prevents `device is not a TTY error`) 

### Test plan

- Generated snapshots using the snapshot tool and then restored them to another instance (running the same Sourcegraph version)

### DIFF
--- a/cmd/src/snapshot_databases.go
+++ b/cmd/src/snapshot_databases.go
@@ -113,7 +113,7 @@ TARGETS FILES
 					}
 				}
 
-				out.WriteLine(output.Emoji(output.EmojiSuccess, "Sucessfully completed dump commands"))
+				out.WriteLine(output.Emoji(output.EmojiSuccess, "Successfully completed dump commands"))
 			} else {
 				b := out.Block(output.Emoji(output.EmojiSuccess, "Run these commands to generate the required database dumps:"))
 				b.Write("\n" + strings.Join(commands, "\n"))
--- a/cmd/src/snapshot_databases.go
+++ b/cmd/src/snapshot_databases.go
@@ -49,35 +49,13 @@ TARGETS FILES
 			}
 			out := output.NewOutput(flagSet.Output(), output.OutputOpts{Verbose: *verbose})
 
-			var builder string
-			if len(args) > 0 {
-				builder = args[len(args)-1]
-			}
+			builder := flagSet.Arg(0)
 
-			targetKey := "docker"
-			var commandBuilder pgdump.CommandBuilder
-			switch builder {
-			case "pg_dump", "":
-				targetKey = "local"
-				commandBuilder = func(t pgdump.Target) (string, error) {
-					cmd := pgdump.DumpCommand(t)
-					if t.Target != "" {
-						return fmt.Sprintf("%s --host=%s", cmd, t.Target), nil
-					}
-					return cmd, nil
-				}
-			case "docker":
-				commandBuilder = func(t pgdump.Target) (string, error) {
-					return fmt.Sprintf("docker exec -i %s sh -c '%s'", t.Target, pgdump.DumpCommand(t)), nil
-				}
-			case "kubectl":
-				targetKey = "k8s"
-				commandBuilder = func(t pgdump.Target) (string, error) {
-					return fmt.Sprintf("kubectl exec -i %s -- bash -c '%s'", t.Target, pgdump.DumpCommand(t)), nil
-				}
-			default:
+			commandBuilder, targetKey := pgdump.Builder(builder, pgdump.DumpCommand)
+			if targetKey == "" {
 				return errors.Newf("unknown or invalid template type %q", builder)
 			}
+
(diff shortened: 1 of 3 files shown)


internal/pgdump/pgdump.go
```diff
@@ -30,9 +30,19 @@ type Target struct {
 	Password string `yaml:"password"`
 }
 
-// Command generates a pg_dump command that can be used for on-prem-to-Cloud migrations.
-func Command(t Target) string {
-	dump := fmt.Sprintf("pg_dump --no-owner --format=p --no-acl --username=%s --dbname=%s",
+// RestoreCommand generates a psql command that can be used for migrations.
+func RestoreCommad(t Target) string {
```
jdpleiness, line 34:
```suggestion
func RestoreCommand(t Target) string {
```

cmd/src/snapshot_restore.go
```diff
@@ -0,0 +1,129 @@
...
+				for _, c := range commands {
+					out.WriteLine(output.Emojif(output.EmojiInfo, "Running command: %q", c))
+					command := exec.Command("bash", "-c", c)
+					out, err := command.CombinedOutput()
+					if err != nil {
+						return errors.Wrapf(err, "failed to run command: %q\n%q", c, string(out))
+					}
+				}
+
+				out.WriteLine(output.Emoji(output.EmojiSuccess, "Sucessfully completed restore commands"))
```
jdpleiness, line 114:
```suggestion
out.WriteLine(output.Emoji(output.EmojiSuccess, "Successfully completed restore commands"))
```

jdpleiness (approved):
Looks good! 🚀

cmd/src/snapshot_databases.go
```diff
@@ -94,19 +96,31 @@ TARGETS FILES
...
 			_ = os.MkdirAll(srcSnapshotDir, os.ModePerm)
 
-			b := out.Block(output.Emoji(output.EmojiSuccess, "Run these commands to generate the required database dumps:"))
-			b.Write("\n" + strings.Join(commands, "\n"))
-			b.Close()
+			if *run {
+				for _, c := range commands {
+					out.WriteLine(output.Emojif(output.EmojiInfo, "Running command: %q", c))
+					command := exec.Command("bash", "-c", c)
+					out, err := command.CombinedOutput()
```
bobheadxi, line 110: I forget, how much output does pgdump and friends generate? If it gives some progress info we might want to send output to stdout/stderr instead of collecting them with `CombinedOutput`

bobheadxi (commented):
nice work! a few comments

cmd/src/snapshot_restore.go
```diff
@@ -0,0 +1,129 @@
...
+					return fmt.Sprintf("docker exec -i %s sh -c '%s'", t.Target, pgdump.RestoreCommand(t)), nil
+				}
+			case "kubectl":
+				targetKey = "k8s"
+				commandBuilder = func(t pgdump.Target) (string, error) {
+					return fmt.Sprintf("kubectl exec -i %s -- bash -c '%s'", t.Target, pgdump.RestoreCommand(t)), nil
+				}
+			default:
+				return errors.Newf("unknown or invalid template type %q", builder)
+			}
```
bobheadxi, line 81: this is somewhat repeated from `shaptshot_databases`, do we want to generalize it such that the caller can provide the pg command builder?

bobheadxi: the only change seems to be `pgdump.DumpCommand` -> `pgdump.RestoreCommand`

cmd/src/snapshot_databases.go
```diff
@@ -49,7 +51,7 @@ TARGETS FILES
 
 			var builder string
 			if len(args) > 0 {
-				builder = args[0]
+				builder = args[len(args)-1]
 			}
```
bobheadxi, line 55: this was my mistake, I think we want to use `flagSet.Args()` to get the non-flag arguments: https://sourcegraph.com/github.com/golang/go@1fffeddfe9c977510d855277da57e0564700d6c3/-/blob/src/flag/flag.go?L626-627

cmd/src/snapshot_databases.go
jac: Restore was pretty noisy so now restore stdout is redirected to /dev/null
Only stderr is captured

PR #948, opened 2023-03-02
Adds validation of CSI drivers for GKE

jasonhawkharris:
### Test plan
Manual testing. The functionality is, in itself, a test. Manually tested failing case and passing case.


internal/validate/kube/gke.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+type ClusterInfo struct {
+	ServiceType string
+	ProjectId   string
+	Region      string
+	ClusterName string
+}
+
+func Gke(ctx context.Context) Option {
```
jdpleiness, line 17: Does this need a context?

jdpleiness (approved):
LGTM, just a few cleanup items 🚀

PR #951, opened 2023-03-06
Adds conditions for clusters that do not have Sourcegraph installed.

jasonhawkharris:
- Before, clusters/namespaces would pass all validations even if no Sourcegraph instance was installed.
- Now: pods, services, connections and PVC validations fail with a warning if none exist.

### Test plan
Manually tested on AKS, GKE clusters with and without a Sourcegraph instance.


jdpleiness (approved):
Looks good!

PR #957, opened 2023-03-14
Admin: add `src admin create` command

jdpleiness:
Add `src admin create` command to setup an initial admin account on a new deployment of Sourcegraph. This is mainly useful for CI pipelines and integration testing purposes in which a Sourcegraph deployment will be deployed in an automated way.

This functionality was previously tied to the experimental `src validate` command but was removed.

### Test plan

Locally tested changes using latest Sourcegraph Helm install.

ran go test ./... - all tests passing
ran staticcheck ./... - no issues found in new code

### DIFF
--- a/cmd/src/admin_create.go
+++ b/cmd/src/admin_create.go
@@ -1,0 +1,74 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+
+	"github.com/sourcegraph/src-cli/internal/users"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+func init() {
+	usage := `
+Examples:
+
+	Create an initial admin user on a new Sourcegraph deployment:
+
+		$ src admin create -url https://your-sourcegraph-url -username admin -email admin@yourcompany.com -password p@55w0rd -with-token
+`
+
+	flagSet := flag.NewFlagSet("create", flag.ExitOnError)
+	usageFunc := func() {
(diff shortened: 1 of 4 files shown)


cmd/src/admin_create.go
```diff
@@ -0,0 +1,74 @@
...
+		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src users %s':\n", flagSet.Name())
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+
+	var (
+		urlFlag      = flagSet.String("url", "", "The base URL for the Sourcegraph instance.")
+		usernameFlag = flagSet.String("username", "", "The new admin user's username.")
+		emailFlag    = flagSet.String("email", "", "The new admin user's email address.")
+		passwordFlag = flagSet.String("password", "", "The new admin user's password.")
```
eseliger, line 32: Should we read this from stdin instead? Otherwise it appears in the bash history.

jdpleiness: Added env var option to set password and a warning in usage examples to prefer using env var over the `-password` flag. The `-password` flag can still be useful in certain temporary situations however (CI pipelines, integration tests, etc...) so it will remain for now. See [commit](https://github.com/sourcegraph/src-cli/pull/957/commits/5ddd6cce48e03be76b3bf3dee02b72061976d09a)

BolajiOlajide (approved):
Approving to unblock.

PR #958, opened 2023-03-15
Validate: no longer require SRC_GITHUB_TOKEN to be set

### LINKED ISSUE #956: `src validate install` requires GH token no matter what
# Overview

In the `src validate install` command, you are able to define a custom spec to run for your validations. This spec can do a number of different things, including search across repos that already exist, create an insight, or add an external service. In this process, we validate that the user has the variable `SRC_GITHUB_TOKEN` set in their shell, but that token not be required depending on the spec.

## Proposed Behavior

* Fail gracefully if an action requiring a GH token (Ex: adding an external service) is needed and no token is set.
* Allow for searching across repos that already exist on the instance
* Do not attempt to clone or validate anything related to external services if there's no spec to create one.

jdpleiness:
Previously `src validate install` required the env var `SRC_GITHUB_TOKEN` to be set, even if no external services were being added. This also blocked validating queries with repositories that already existed on a Sourcegraph deployment.

This fix allows `src validate install` to be run without requiring the `SRC_GITHUB_TOKEN` env var to be set, as well as validating queries with repositories that already exist on a deployment.

Fixes #956 
Fixes https://github.com/sourcegraph/sourcegraph/issues/47576

### Test plan
Locally tested changes using latest Sourcegraph Helm install.

ran go test ./... - all tests passing
ran staticcheck ./... - no issues found in new code

### DIFF
--- a/cmd/src/validate_install.go
+++ b/cmd/src/validate_install.go
@@ -11,7 +11,6 @@ import (
 	"github.com/mattn/go-isatty"
 
 	"github.com/sourcegraph/src-cli/internal/api"
-	"github.com/sourcegraph/src-cli/internal/validate"
 	"github.com/sourcegraph/src-cli/internal/validate/install"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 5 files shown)


scjohns (commented):
LGTM, but I don't write Go code so would be great to have someone more familiar with the code give the approval

eseliger (approved):
did a quick glance, nothing stood out so let's go :)

PR #959, opened 2023-03-16
Bump sourcegraph dependency to fix version parsing of insiders

eseliger:
For untagged versions, we added the current Sourcegraph version as well, so the parsing didn't work anymore. This fixes it.

### Test plan

Verified the error no longer shows when targeting S2.


cmd/src/snapshot_upload.go
```diff
@@ -113,7 +113,7 @@ BUCKET
 			progress := out.Progress(progressBars, nil)
 			progress.WriteLine(output.Emoji(output.EmojiHourglass, "Starting uploads..."))
 			bucket := c.Bucket(*bucketName)
-			g := group.New().WithErrors().WithContext(ctx)
+			g := pool.New().WithErrors().WithContext(ctx)
```
BolajiOlajide, line 116: Shiny!

BolajiOlajide (approved):

cmd/src/snapshot_upload.go
eseliger: I had to, the `lib/group` package was deleted :lolsob: 😬

eseliger:
oh boi this is a big mess. conc requires go 1.19, but then golangci-lint fails because it doesn't understand 1.19 in that version, and when I upgrade it it starts complaining about new stuff - really annoying.

eseliger:
@BolajiOlajide @Piszmog could I get a quick re-review on this, I had to do a bunch of changes to get it to compile with conc as a dependency.

internal/batches/docker/context.go
```diff
@@ -67,9 +67,11 @@ func (e *fastCommandTimeoutError) Error() string {
 
 func (*fastCommandTimeoutError) Timeout() bool { return true }
 
+type fastCommandTimeoutKey string
```
BolajiOlajide, line 70: Why do we need this type?

internal/batches/service/remote.go
```diff
@@ -185,6 +185,8 @@ func (svc *Service) uploadFile(ctx context.Context, workingDir, filePath, batchS
 		// Errors passed to pipeWriter.CloseWithError come through here.
 		return err
 	}
+	defer resp.Body.Close()
```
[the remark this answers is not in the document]
BolajiOlajide: Thanks for catching.

BolajiOlajide (approved):

internal/batches/docker/context.go
eseliger: golangci-lint reminded me that using a custom type the string is ensured to always be globally unique, if it's a plain string it might have clashes with other ctx keys - pretty interesting magic honestly

internal/batches/service/remote.go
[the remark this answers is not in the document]
eseliger: golangci-lint did, I know nothing 😆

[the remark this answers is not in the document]
BolajiOlajide: Oh, nice. I like.

internal/batches/docker/context.go
BolajiOlajide: Oh interesting.

PR #973, opened 2023-04-21
Scj/update/smtp validate

scjohns:
### Test plan

- [X] Validate that adding smtp configs to a validation.yaml file will attempt to send a test 
- [X] Validate that not including smtp in validation.yaml does not trigger an attempt to send a test


jasonhawkharris (approved):
LGTM!

PR #974, opened 2023-04-21
Scj/update/executor validate

scjohns:
### Test plan

- [X] Confirm that adding `executor.enabled: true` to vaildate.yaml will check live executor connections
- [X] Confirm that omitting will skip executor connection validation

### DIFF
--- a/internal/validate/install/install.go
+++ b/internal/validate/install/install.go
@@ -114,10 +114,10 @@ func checkExecutors(ctx context.Context, client api.Client, query string, variab
 
 	ok, err := client.NewRequest(q.query, q.variables).Do(ctx, &result)
 	if err != nil {
-		return 0, errors.Wrap(err, "checkExecutors failed")
+		return -1, errors.Wrap(err, "checkExecutors failed")
 	}
 	if !ok {
-		return 0, errors.New("checkExecutors failed, no data to unmarshal")
+		return -1, errors.New("checkExecutors failed, no data to unmarshal")
 	}
 
 	return result.Executor.TotalCount, nil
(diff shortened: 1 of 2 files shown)


internal/validate/install/install.go
```diff
@@ -71,6 +99,30 @@ func Validate(ctx context.Context, client api.Client, config *ValidationSpec) er
...
+
+	var result struct {
+		Executor struct {
+			TotalCount int `json:"totalCount"`
+		} `json:"executors"`
+	}
+
+	ok, err := client.NewRequest(q.query, q.variables).Do(ctx, &result)
+	if err != nil {
+		return 0, errors.Wrap(err, "checkExecutors failed")
```
jasonhawkharris, line 117: Consider using `-1` as the int value when returning an error. See note [here](https://chortle.ccsu.edu/qbasic/chapter09/bc09_15.html#:~:text=A%20sentinel%20value%20should%20be,it%20without%20changing%20the%20sum.) on sentinel values. Just a preference, but there's lots of precedence for this convention. Also, you do, in fact, validly return `0` at times. -1 is a value you would never expect to receive if the function is behaving as expected.

internal/validate/install/install.go
```diff
@@ -71,6 +99,30 @@ func Validate(ctx context.Context, client api.Client, config *ValidationSpec) er
...
+	}
+	if !ok {
+		return 0, errors.New("checkExecutors failed, no data to unmarshal")
```
jasonhawkharris, line 120: Same as above. Consider using -1 instead of 0.

jasonhawkharris (approved):
Looks good to me. I made one note about preferences but I think the note generally lines up with convention.

PR #978, opened 2023-05-02
Adds the `src scout` utility to the CLI

jasonhawkharris:
The `src scout` utility is a command that allows users manage resources. This PR:
- adds the `scout` utility to the CLI
- adds the `resources` subcommand, which prints resource allocations for all Sourcegraph services
- adds unit tests for ResourcesK8s and ResourcesDocker

### Test plan
Shipped with unit tests.


internal/scout/scout.go
```diff
@@ -0,0 +1,10 @@
+package scout
+
+type Status string
+
+var (
+	WarningSign           = "⚠️ "
+	EmojiFingerPointRight = "👉"
+)
+
+const Warning Status = "Warning"
```
jdpleiness, line 10: It doesn't look like these are used anywhere

internal/scout/resources/resources.go
```diff
@@ -0,0 +1,191 @@
...
+				pvc, err := cfg.k8sClient.CoreV1().PersistentVolumeClaims(cfg.namespace).Get(
+					ctx,
+					volume.PersistentVolumeClaim.ClaimName,
+					metav1.GetOptions{},
+				)
+
+				if err != nil {
+					return "", errors.Wrapf(
+						err,
+						"Error getting PVC %s",
```
jdpleiness, line 105:
```suggestion
						"error getting PVC %s",
```

internal/scout/resources/resources.go
```diff
@@ -0,0 +1,191 @@
...
+	}
+
+	return listPodResources(ctx, cfg)
+}
+
+func listPodResources(ctx context.Context, cfg *Config) error {
+	podInterface := cfg.k8sClient.CoreV1().Pods(cfg.namespace)
+	podList, err := podInterface.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "Error listing pods: ")
```
jdpleiness, line 54:
```suggestion
		return errors.Wrap(err, "error listing pods: ")
```

jdpleiness (commented):
On first pass this looks good overall. I would revisit the testing setup and whether or not to use mocking or if another approach would work.

internal/scout/README.md
jdpleiness, line 1: If this is just empty, we should probably get rid of it

internal/scout/resources/resources_test.go
```diff
@@ -0,0 +1,168 @@
...
+	if err != nil {
+		t.Fatalf("ResourcesDocker returned an error: %v", err)
+	}
+
+	err = dockerClient.Close()
+	if err != nil {
+		t.Fatalf("Error closing docker client: %v", err)
+	}
+
+	w.Close()
```
jdpleiness, line 155: Even though this is just a test, we probably still want to use a `defer` with this

internal/scout/resources/resources.go
```diff
@@ -0,0 +1,194 @@
...
+				capacity = pvc.Status.Capacity.Storage().String()
+				break
+			}
+		}
+	}
+	return capacity, nil
+}
+
+// DockerClientInterface defines the interface for interacting with the Docker API.
+type DockerClientInterface interface {
```
jdpleiness, line 121: I don't think this needs to be here if we break some of the code up and change how the tests work. Lets pair on this though since it's a bit more than a PR suggestion 💪

jasonhawkharris: Sounds good to me. I'll put some time on the calendar.

internal/scout/resources/resources_test.go
jasonhawkharris: Hmm... the test totally freaked out when I deferred this for some reason.

jasonhawkharris: Keeping it the same for now.

jdpleiness (approved):
Looks like everything as been addressed 🚀 

LGTM!

PR #979, opened 2023-05-05
build src-cli w/ bazel

davejrt:
Allows us to build src-cli with bazel and subsequently use bazel build in other projects where we use src-cli ie: codeintelQA 

### Test plan

```➜  src-cli git:(dt/bazel) bazel build //...
INFO: Analyzed 45 targets (0 packages loaded, 0 targets configured).
INFO: Found 45 targets...
INFO: Elapsed time: 0.188s, Critical Path: 0.00s
INFO: 1 process: 1 internal.
INFO: Build completed successfu


jhchabran (changes requested):
Can't we instead use the binaries as they are released for each platform? https://github.com/sourcegraph/src-cli/releases

davejrt:
> Can't we instead use the binaries as they are released for each platform? https://github.com/sourcegraph/src-cli/releases

One of the tests builds it from a specific commit, not a release https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/dev/ci/integration/code-intel/install-src.sh

eseliger (approved):
this won't build the released versions with bazel yet would it? we use goreleaser for that. Just curious, not a blocker

jhchabran (approved):

davejrt:
> this won't build the released versions with bazel yet would it? we use goreleaser for that. Just curious, not a blocker

Not yet, we're using this so we can build src-cli for other images/tests inside the sourcegraph/sourcegraph code base with bazel instead of go run/go build

PR #981, opened 2023-05-16
add ability to dump all resource data to text file

jasonhawkharris:
### Test plan
Tested manually for kubernetes and docker deployments of sourcegraph


internal/scout/resource/style/Table.go
jdpleiness, line 1: Should be `table.go` instead of `Table.go`

jdpleiness (approved):

PR #983, opened 2023-05-18
add usage command to 'src scout'

jasonhawkharris:
### Test plan
- Shipped with unit tests.
- Tested manually with healthy, and unhealthy k8s instances of sourcegraph


internal/scout/usage/k8s.go
```diff
@@ -0,0 +1,384 @@
...
+//   - k8sClient: A Kubernetes clientset for accessing the API server
+//   - namespace: The namespace of the pods
+//
+// It returns:
+// - Any error that occurred while rendering the table
+func renderUsageTable(ctx context.Context, cfg *Config) error {
+	podInterface := cfg.k8sClient.CoreV1().Pods(cfg.namespace)
+	podList, err := podInterface.List(ctx, metav1.ListOptions{})
+	if err != nil {
+		return errors.Wrap(err, "error listing pods: ")
```
jdpleiness, line 81: Nit: At this point, we know it's an error. So we don't really need to list out `error` in the error message itself. Otherwise when looking over log output we get messages like "error: error listing pods: error no such file found"... Something like "could not list pods: \<wrapped error\>" would be more concise. Same for all these messages.

internal/scout/usage/k8s.go
```diff
@@ -0,0 +1,384 @@
...
+// - containerName: The name of the container
+//
+// It returns:
+// - The storage usage of the container (or "-" if stateless)
+// - Any error that occurred while executing the command
+//
+// When the kubernetes API has a better way of getting storage usage,
+// this should be refactored to use the API.
+func getStorageUsage(podName, namespace, containerName string) (string, error) {
+	cmd := exec.Command("kubectl", "exec", "-it", podName, "-n", namespace, "--", "df", "-h")
```
jdpleiness, line 361: This works, but it also introduces an assumed dependency that the user has `kubectl` installed and in their path. We can use the `k8s.io/client-go/tools/remotecommand` pkg to do the same thing without assuming the user has `kubectl` at all.

jdpleiness (commented):
Looking good 😎 

A few suggestions here and there.

cmd/src/scout_usage.go
```diff
@@ -0,0 +1,122 @@
...
+		fmt.Println(cmdUsage)
+	}
+
+	var (
+		kubeConfig *string
+		namespace  = flagSet.String("namespace", "", "(optional) specify the kubernetes namespace to use")
+		pod        = flagSet.String("pod", "", "(optional) specify a single pod")
+		docker     = flagSet.Bool("docker", false, "(optional) using docker deployment")
+		spy        = flagSet.Bool("spy", false, "(optional) see resource usage in real time")
+		// TODO: option for getting resource allocation of the Node
```
jdpleiness, line 52: Take out TODO

internal/scout/usage/k8s.go
```diff
@@ -0,0 +1,399 @@
...
+	"strings"
+
+	"github.com/charmbracelet/bubbles/table"
+	"github.com/charmbracelet/lipgloss"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/sourcegraph/src-cli/internal/scout/style"
+
+	"gopkg.in/inf.v0"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
jdpleiness, line 19: Duplicate?

jdpleiness (approved):
Two minor fixes, but overall looks good!

PR #984, opened 2023-05-24
Adds the `--pod` flag to `scout usage`

jasonhawkharris:
This commit: 
- Adds the `--pod` flag so that users can target resources on a single pod instead of the entire cluster
- Makes improvements to the architecture

### Test plan
Shipped with unit tests. 
Also manually tested on healthy and unhealthy k8s deployments.


internal/scout/usage/k8s.go
```diff
@@ -362,7 +453,7 @@ func getStorageUsage(
 
 	exec, err := remotecommand.NewSPDYExecutor(cfg.restConfig, "POST", req.URL())
 	if err != nil {
-		return "", err
+		return -1, -1, err
```
jdpleiness, line 456: I think these should return the zero value for `float64` by convention which would be `0`

jdpleiness (approved):
LGTM! 👍

PR #985, opened 2023-05-25
scout/usage: add `--docker` flag to `scout usage` tool

jasonhawkharris:
### Test plan
Tested manually with healthy and unhealthy docker sourcegraph instance.

### DIFF
--- a/internal/scout/usage/docker.go
+++ b/internal/scout/usage/docker.go
@@ -7,6 +7,7 @@ import (
 	"github.com/charmbracelet/bubbles/table"
 	"github.com/docker/docker/api/types"
 	"github.com/docker/docker/client"
+	"github.com/jasonhawkharris/dockerstats"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/src-cli/internal/scout/style"
 )
@@ -32,11 +33,16 @@ func Docker(ctx context.Context, client client.Client, opts ...Option) error {
 	if err != nil {
 		return errors.Wrap(err, "could not get list of containers")
 	}
-    
+
 	return renderDockerUsageTable(ctx, cfg, containers)
 }
 
 func renderDockerUsageTable(ctx context.Context, cfg *Config, containers []types.Container) error {
+	stats, err := dockerstats.Current()
+	if err != nil {
+		return errors.Wrap(err, "could not get docker stats")
+	}
+
 	columns := []table.Column{
 		{Title: "Container", Width: 20},
 		{Title: "Cores", Width: 10},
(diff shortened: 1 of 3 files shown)


internal/scout/usage/docker.go
```diff
@@ -4,25 +4,89 @@ import (
...
+	style.ResourceTable(columns, rows)
 	return nil
 }
+
+// makeDockerUsageRow generates a table row displaying CPU and memory usage for a Docker container.
+// It takes a ContainerJSON struct containing info about the container and a Stats struct containing usage stats.
+// It calculates the number of CPU cores, CPU usage percentage, memory limit in GB, and memory usage percentage.
+// It then returns a table.Row containing this info, to be displayed in the usage table.
+func makeDockerUsageRow(containerInfo types.ContainerJSON, usage dockerstats.Stats) table.Row {
+	cpuCores := containerInfo.HostConfig.NanoCPUs / 1000000000
```
eseliger, line 81: Go syntax sugar, makes it easier to read :) 
```suggestion
	cpuCores := containerInfo.HostConfig.NanoCPUs / 1_000_000_000
```

eseliger: since these two are of type int, the result will always be in full cores, should we use float and format it as 2 decimals or so? Entire CPUs might be a bit too coarse.

eseliger: Same for memory below.

internal/scout/usage/docker.go
```diff
@@ -4,25 +4,89 @@ import (
...
+	rows := []table.Row{}
+
+	for _, container := range containers {
+		containerInfo, err := cfg.dockerClient.ContainerInspect(ctx, container.ID)
+		if err != nil {
+			return errors.Wrap(err, "could not get container info")
+		}
+
+		for _, s := range stats {
+			if s.Container == container.ID[0:12] {
```
eseliger, line 64: are these IDs always guaranteed to be 12 characters long?

eseliger (commented):
I'm missing a bit of context here, what is the relation of this tool and the `docker stats` command? 
This is what it would give:
```
 ~ docker stats --no-stream --no-trunc --all
CONTAINER ID                                                       NAME                 CPU %     MEM USAGE / LIMIT     MEM %     NET I/O       BLOCK I/O   PIDS
4844cb8c6fd16f57f69290197f8eaa6b8f4e7a340148a8ba3bc3ad1c4139b972   syntax-highlighter   0.00%     34.92MiB / 7.765GiB   0.44%     1.51kB / 0B   0B / 0B     14
6696bf320826e206b910284e2f4b8b0bf8d2d45105a343d347383cdbbaf2c89e   lago-front           0.00%     0B / 0B               0.00%     0B / 0B       0B / 0B     0
```

internal/scout/usage/docker.go
jasonhawkharris: Yes. The dockerstats library truncates the ID to the first 12 characters every time.

jasonhawkharris:
@eseliger long and short of it is that I just haven't found a convenient way to get cpu/memory usage from the goclient for the docker API. The `dockerstats` library solves this issue, and does so quicker than if I were to just wrap the command in an `exec` function.

internal/scout/usage/docker.go
```diff
@@ -2,27 +2,88 @@ package usage
...
+	}
+	rows := []table.Row{}
+
+	for _, container := range containers {
+		containerInfo, err := cfg.dockerClient.ContainerInspect(ctx, container.ID)
+		if err != nil {
+			return errors.Wrap(err, "failed to get container info")
+		}
+
+		stats, err := cfg.dockerClient.ContainerStats(context.Background(), container.ID, false)
```
eseliger, line 54: Probably want to use the same context as above here? 
```suggestion
		stats, err := cfg.dockerClient.ContainerStats(ctx, container.ID, false)
```

internal/scout/usage/docker.go
```diff
@@ -2,27 +2,88 @@ package usage
...
+		if err != nil {
+			return errors.Wrap(err, "could not get container stats")
+		}
+		defer func() {
+			_ = stats.Body.Close()
+		}()
```
eseliger, line 60: Go style nit: this should be equivalent
```suggestion
		defer stats.Body.Close()
```

eseliger (approved):
Code LGTM, just to clarify - what is this going to be used for that `docker stats` could not do? Is this for envs where docker cli isn't available? Is this meant to generate a more elaborate report?

jasonhawkharris:
`src scout usage` works on a Kubernetes deployment and returns current usages. The `usage` command, in particular, is more useful for Kubernetes deployments.

However, adding the `--docker` flag will allow it to work on docker deployments as well. It's less needed because of the `docker stats` command, but I wanted to have it as an option for our support/TA/CE folks because this command takes you to a TUI that allows the user to easily export this information and send it back to us. 

Happy to hop on a call if you'd like to see this in real time.

eseliger:
Thanks, that's the bit of context I was missing :)

PR #1012, opened 2023-07-25
codeintel: handle -insecure-skip-verify in SCIP endpoint detection

Strum355:
Previously, we were doing SCIP detection before parsing cli flags. As SCIP detection makes a HTTP request, this meant it wasnt configured with the (global :grimacing:) API client flags

### Test plan

Ran manually to confirm flag is picked up at the correct time with :sparkles: print statements :sparkles:

### DIFF
--- a/cmd/src/code_intel_upload_flags.go
+++ b/cmd/src/code_intel_upload_flags.go
@@ -100,9 +100,9 @@ func init() {
 //
 // On success, the global codeintelUploadFlags object will be populated with valid values. An
 // error is returned on failure.
-func parseAndValidateCodeIntelUploadFlags(args []string, isSCIPAvailable bool) (*output.Output, error) {
+func parseAndValidateCodeIntelUploadFlags(args []string) (*output.Output, bool, error) {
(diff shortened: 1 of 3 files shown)


efritz (approved):

cmd/src/code_intel_upload_flags.go
```diff
@@ -100,9 +100,9 @@ func init() {
 //
 // On success, the global codeintelUploadFlags object will be populated with valid values. An
 // error is returned on failure.
-func parseAndValidateCodeIntelUploadFlags(args []string, isSCIPAvailable bool) (*output.Output, error) {
+func parseAndValidateCodeIntelUploadFlags(args []string) (*output.Output, bool, error) {
 	if err := codeintelUploadFlagSet.Parse(args); err != nil {
-		return nil, err
+		return nil, false, err
```
olafurpg, line 105: Non-blocking: we default to "scip is not available", but it might make sense to flip the boolean and rename it into `isLSIFRequired` or something like that

olafurpg (approved):

PR #1015, opened 2023-08-07
batches: File upload error handling

Piszmog:
Two recent issues were found.
* When a file exceeded the size limit, the file name was not in the error. This made it hard to tell which file is causing the issue
* The error being returned was still set after logging as a warn. This made it so the program still errored out

[Thread](https://sourcegraph.slack.com/archives/C04M9JE73E1/p1691434360532449) of context.

### Test plan

Existing tests cover changes.

Screenshot 2023-08-07 at 13 41 54


peterguy (approved):
Good catch on `err` being set

BolajiOlajide (approved):

PR #1021, opened 2023-08-17
Clean up volumes if workspace preparation fails

lowjoel:
If workspace preparation fails, the temporary Docker volume is not cleaned up.

### Test plan

I'll need help filling in this bit. Apart from manually inducing an error with the workspace process, how else? I could possibly build locally a broken batch-change container image that doesn't have Git nor unzip?

### DIFF
--- a/internal/batches/workspace/volume_workspace.go
+++ b/internal/batches/workspace/volume_workspace.go
@@ -30,16 +30,22 @@ type dockerVolumeWorkspaceCreator struct {
 
 var _ Creator = &dockerVolumeWorkspaceCreator{}
 
-func (wc *dockerVolumeWorkspaceCreator) Create(ctx context.Context, repo *graphql.Repository, steps []batcheslib.Step, archive repozip.Archive) (Workspace, error) {
+func (wc *dockerVolumeWorkspaceCreator) Create(ctx context.Context, repo *graphql.Repository,
+	steps []batcheslib.Step, archive repozip.Archive) (ws Workspace, err error) {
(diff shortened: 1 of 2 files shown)


internal/batches/workspace/volume_workspace.go
```diff
@@ -219,7 +229,8 @@ var _ Workspace = &dockerVolumeWorkspace{}
 
 func (w *dockerVolumeWorkspace) Close(ctx context.Context) error {
 	// Cleanup here is easy: we just get rid of the Docker volume.
-	return exec.CommandContext(ctx, "docker", "volume", "rm", w.volume).Run()
+	_ = deleteVolume(ctx, w.volume)
+	return nil
```
eseliger, line 233:
```suggestion
	return deleteVolume(ctx, w.volume)
```

To keep the error reporting

eseliger (commented):
Left one comment about preserving error reporting from the `Close` method. Thanks for the fix!

internal/batches/workspace/volume_workspace.go
lowjoel: D'oh, not sure how I misread the code. I'll fix.

lowjoel: Fixed, thanks for pointing out!

lowjoel:
I also fixed the tests this round. Let me know what you think.

BolajiOlajide (approved):
LGTM.

Thanks for the fix.

eseliger (approved):

PR #1022, opened 2023-08-17
switch to attoto/clipboard package

jasonhawkharris:
Change clipboard library to a cross platform package to eliminate the hard x11 dependency. 

### Test plan
1. Manual test with `usage` command: passed
2. Manual test with `resource` command: passed


mrnugget (approved):
Sweet! Love the manual testing in the PR description.

Changelog entry?

PR #1037, opened 2023-09-14
servegit: use error hook

keegancsmith:
Before this if the git command failed we would nil panic since the Logger on gitservice.Handler was not set. This commit updates us to the latest version of lib which instead takes in an error hook so we can use the attached stdlib logger.

Corresponding PR at https://github.com/sourcegraph/sourcegraph/pull/56611

Test Plan: go test

### DIFF
--- a/internal/servegit/serve.go
+++ b/internal/servegit/serve.go
@@ -106,6 +106,9 @@ func (s *Serve) handler() http.Handler {
 		Dir: func(name string) string {
 			return filepath.Join(s.Root, filepath.FromSlash(name))
 		},
+		ErrorHook: func(err error, stderr string) {
+			s.Info.Printf("git-service error: %s\nstderr:\n%s", err.Error(), stderr)
+		},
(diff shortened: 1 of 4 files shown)


internal/servegit/serve.go
```diff
@@ -106,6 +106,9 @@ func (s *Serve) handler() http.Handler {
 		Dir: func(name string) string {
 			return filepath.Join(s.Root, filepath.FromSlash(name))
 		},
+		ErrorHook: func(err error, stderr string) {
+			s.Info.Printf("git-service error: %s\nstderr:\n%s", err.Error(), stderr)
```
mrnugget, line 110: `Info`? Why not `Error`?

keegancsmith: There is no error in this case.

mrnugget: `:sad-trombone:`

mrnugget (approved):

keegancsmith:
@DaedalusG fix for the nil panic is now in main :)

PR #1048, opened 2023-11-08
goreleaser: quote go version number

burmudar:
YAML sees '1.20' as a number and so it cuts off the trailing 0, leading to the action setting up go 1.2. So we add a '.x' to make it a string.

[Action seems to be checking out 1.20 ](https://github.com/sourcegraph/src-cli/actions/runs/6803034582)

### Test plan
CI


BolajiOlajide (approved):

camdencheek:
🙌 YAML 🙌

PR #1049, opened 2023-11-23
 Add fail-fast flag to src batch

gabtorre:
This adds a new `--fail-fast` flag to the `src batch` commands. When enabled, any errors encountered while executing steps in a repository will stop the execution of the whole batch spec.  The flag is off by default, so existing behavior is preserved.

- A new CancelableContext type was added to support cancellation with errors. The batch commands now wrap the context with this to enable fail-fast behavior.
- Updated contextCancelOnInterrupt to support cancellation with error causes.
- Added error check in executeBatchSpec to surface fail-fast cancellation errors.
- Errors now propagate up by calling the cancel context's `Cause()`
- Minor code reorganizations around log file handling to return logs if -fail-fast and -keep-logs are used together.


BolajiOlajide:
Hey @gabtorre 

Apologies for the long delay. I'll review this soon.
Thank you so much.

internal/batches/executor/executor.go
```diff
@@ -92,12 +92,14 @@ func NewExecutor(opts NewExecutorOpts) *executor {
...
+var ErrFastFail = errors.New("Execution stopped due to fast-fail mode.")
+
 // Start starts the execution of the given Tasks in goroutines, calling the
 // given taskStatusHandler to update the progress of the tasks.
-func (x *executor) Start(ctx context.Context, tasks []*Task, ui TaskExecutionUI) {
+func (x *executor) Start(ctx CancelableContext, tasks []*Task, ui TaskExecutionUI) {
 	defer func() { close(x.doneEnqueuing) }()
-
 	for _, task := range tasks {
+		fmt.Println(task.Repository)
```
BolajiOlajide, line 102: Left over debug statement

cmd/src/batch_common.go
```diff
@@ -628,22 +637,22 @@ func checkExecutable(cmd string, args ...string) error {
...
+func contextCancelOnInterrupt(parent context.Context) (context.Context, func(error)) {
+	ctx, ctxCancel := context.WithCancelCause(parent)
 	c := make(chan os.Signal, 1)
 	signal.Notify(c, os.Interrupt)
 
 	go func() {
 		select {
 		case <-c:
-			ctxCancel()
+			ctxCancel(errors.New("Ctrl-C hit"))
```
BolajiOlajide, line 648: This is awesome. TIL I learned about the `.WithCancelCause` method.

cmd/src/batch_exec.go
```diff
@@ -101,7 +101,7 @@ Examples:
 		}
 
 		ctx, cancel := contextCancelOnInterrupt(context.Background())
-		defer cancel()
+		defer cancel(nil)
```
BolajiOlajide, line 104: I wonder if we should be passing in nil directly here, an error could occur at any point in the execution of a batch spec.

Is there a reason why the `handler` function doesn't return a named return value and we can then pass that into `defer cancel`?

```go
handler := func(args []string) (err error) {
	...
	ctx, cancel := contextCancelOnInterrupt(context.Background())
	defer cancel(err)
	...
}
```

This way we are guaranteed to always have access to whatever error occurs in the lifecycle of an execution.

cmd/src/batch_preview.go
```diff
@@ -40,9 +41,20 @@ Examples:
...
+
+		failFastCancel := func(error) {}
+		if flags.failFast {
+			failFastCancel = cancel
+		}
+
+		cctx := executor.CancelableContext{
+			Context: ctx,
+			Cancel:  failFastCancel,
+		}
```
BolajiOlajide, line 55: Is there a need for this struct? Is there a reason we can't pass in the context and `failFastCancel` as separate arguments?

Or better still have the `failFastCancel` as one of the fields in `executeBatchSpecOpts`, that way we are certain `ctx` is always going to be of type `context.Context` and wouldn't be confusing to anyone in the future;

internal/batches/executor/coordinator.go
```diff
@@ -177,7 +182,7 @@ func (c *Coordinator) buildSpecs(ctx context.Context, batchSpec *batcheslib.Batc
 
 // ExecuteAndBuildSpecs executes the given tasks and builds changeset specs for the results.
 // It calls the ui on updates.
-func (c *Coordinator) ExecuteAndBuildSpecs(ctx context.Context, batchSpec *batcheslib.BatchSpec, tasks []*Task, ui TaskExecutionUI) ([]*batcheslib.ChangesetSpec, []string, error) {
+func (c *Coordinator) ExecuteAndBuildSpecs(ctx CancelableContext, batchSpec *batcheslib.BatchSpec, tasks []*Task, ui TaskExecutionUI) ([]*batcheslib.ChangesetSpec, []string, error) {
```
BolajiOlajide, line 185: Since the arguments to this method is getting bulky, we can move it into a struct and pass in the cancel function explicitly.

```suggestion
type ExecuteAndBuildSpecsArgs struct {
	batchSpec *batcheslib.BatchSpec
	tasks []*Task
	ui TaskExecutionUI
	cancelFunc context.CancelCauseFunc
}

func (c *Coordinator) ExecuteAndBuildSpecs(ctx context.Context, args ExecuteAndBuildSpecsArgs) ([]*batcheslib.ChangesetSpec, []string, error) {
```

BolajiOlajide: The cancelFunc isn't being used here, I wonder if we need to pass it in here as an argument?

PR #1085, opened 2024-05-15
fix go-generate command in goreleaser config

craigfurman:
https://github.com/golang/go/issues/60079 changed the behavior of go-generate when the target package does not exist. It used to silently succeeed, now it errors. The schema package was removed in 2021.

### Test plan

Manually ran `go generate`.


BolajiOlajide (approved):
:ty:

PR #1100, opened 2024-07-09
batches: support new version field

stefanhengl:
This bumps sourcegraph/sourcegraph/lib to pull in the latest changes from https://github.com/sourcegraph/sourcegraph/pull/63613 to support the new `version` field for batch specs.

Notes:
 - To resolve repos, batches calls the GraphQL endpoint `resolveWorkspacesForBatchSpec`, which takes the serialized spec. This means all the logic to resolve repos is outsourced to Sourcegraph. This is great news for this PR because we don't have to worry how `on.RepositoriesMatchingQuery` is interpreted. The code path in which src-cli itself resolves the workspaces was removed in #819.
 
### Test plan
- New unit tests
- Manual testing. I tested the following workflow
  - `src batch new -f test.spec.yaml` (-> spec contains `version: 2`)
  - `src batch apply -f test.spec.yaml`   (-> validation passes,  batch spec shows up in Sourcegraph db)
  - `src batch validate -f test.spec.yaml` (-> validation passes or fails if I change the version to an unsupported value)
  - `src batch repos -f test.spec.yaml` (-> checked that returned repos match with expected pattern type) 

Closes SPLF-126


internal/batches/docker/image.go
```diff
@@ -103,7 +103,7 @@ func (image *image) Ensure(ctx context.Context) error {
 			// docker image inspect will return a non-zero exit code if the image and
 			// tag don't exist locally, regardless of the format.
 			var digest string
-			if digest, err = inspectDigest(); errors.HasType(err, &fastCommandTimeoutError{}) {
+			if digest, err = inspectDigest(); errors.HasType[*fastCommandTimeoutError](err) {
```
stefanhengl, line 106: `errors.HasType` was changed in Sourcegraph so I had to update the call here. (See https://github.com/sourcegraph/sourcegraph/pull/63024)

internal/batches/service/service.go
```diff
@@ -506,7 +506,9 @@ func validateMount(batchSpecDir string, spec *batcheslib.BatchSpec) error {
 	return nil
 }
 
-const exampleSpecTmpl = `name: NAME-OF-YOUR-BATCH-CHANGE
+const exampleSpecTmpl = `# "version" determines how the batch spec is interpreted.
+version: 2
+name: NAME-OF-YOUR-BATCH-CHANGE
```
stefanhengl, line 511: This is returned by `src batch new -f <file>`

jtibshirani: Tiny comment, as a user I would find `# "version" determines how the batch spec is interpreted` to be a bit mysterious. Maybe we could say
```
version: 2 # Use the latest schema version
name: ...
```

jtibshirani (approved):
Looks great!

camdencheek (approved):
Thanks!

Issue #1055, opened 2024-01-26
Allow mounting files/directories as RW instead of always RO


chrisdruta:
## Use Case
Sometimes to speed up batch changes, its useful to use a cache directory (e.g. package manager repository) in-order to avoid hammering package repositories over and over again.

I accomplished this by recompiling the program and removing the ":ro" option for constructing the docker mount commands in [run_steps.go](https://github.com/sourcegraph/src-cli/blob/0621a6386f97c78991f71f879e3e9ba4300d98c7/internal/batches/executor/run_steps.go#L348) which fixed issues I was having.

## Security Implications
Looking at previous PR/discussions, I understand this feature was requested to work for remote environments where this ask might add a security risk.

To resolve any potential concerns, maybe only allow local src-cli to mount files in read-write?

## Related
* https://github.com/sourcegraph/src-cli/pull/816

atishpatel:
👍  to enabling rw mount. 

I have an use case where i want to output some analysis files when running locally per repo so than i can summarize the output for all repos. Forcing a read-only mount drastically reduces the flexibility of the Batch Changes tool.

PR #1103, opened 2024-07-17
Remove ro for locally mounted folders

atishpatel:
https://github.com/sourcegraph/src-cli/issues/1055

Remove read-only config from locally mounted folders. This gives more flexibility to locally run Batch Changes since more files can be shared as output for any step where the folder is mounted.

### DIFF
--- a/internal/batches/executor/run_steps.go
+++ b/internal/batches/executor/run_steps.go
@@ -345,7 +345,7 @@ func executeSingleStep(
 		if err != nil {
 			return bytes.Buffer{}, bytes.Buffer{}, err
 		}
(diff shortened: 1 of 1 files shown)


peterguy:
Thank you for the contribution, @atishpatel!

As was mentioned in #1055, there could be some security concerns mounting volumes read-write, so we should pull in @sourcegraph/security-code-review and some domain experts like @eseliger and @BolajiOlajide to get their input.

atishpatel:
Thanks Peter! 

I am curious what security vector the team is concerned about. I would think security should be fine since mount already has to be a sub director of the batch spec file and the container + code being run is also determined by the user. What are the the sourcegraph security team's concerns?

atishpatel:
Any updates on this PR or something like this PR? My company is using rw local mounting as a core part of tooling we're building around sourcegraph batch changes

Issue #689, opened 2022-02-01
Should --help text be consistently written to stdout?


varungandhi-src:
For `src` CLI, it looks like help text with `--help` is output to either stderr or a mix of stdout and stderr.

For example, `src --help` seems to output everything to stderr. `src lsif upload --help` seems to output some text to stdout and some to stderr.

In either case, you need an extra `2&>1` if you want to pipe the output to a pager. My understanding is outputting help text to stderr is intentionally the default behavior for the `flag` package. (https://github.com/golang/go/issues/17284, https://cs.opensource.google/go/go/+/refs/tags/go1.17.5:src/flag/flag.go;l=369;drc=refs%2Ftags%2Fgo1.17.5)

Since `src` is likely to be used by more non-Go developers compared to Go developers, would it make sense to consistently output the help text to stdout instead?

chrispine:
@varungandhi-src Could you ask around in Slack to see what behavior people expect with this? Once we have that, we'll prioritize a fix. Thanks!

varungandhi-src:
[Poll in Slack](https://sourcegraph.slack.com/archives/C07KZF47K/p1645548986703179) didn't gather a lot of data points:
- +4 for stdout (including my own vote)
- +2 for stderr

mrnugget:
I think [the answers to this SO question](https://stackoverflow.com/questions/1068020/app-help-should-go-to-stdout-or-stderr) are quite good and make sense:

* if the user explicitly asked for `--help` output by providing `--help` then the help text is not an error and should be displayed on `stdout`
* if the user can a command the wrong way and we need to print help output to guide them, it should be printed on `stderr`
* when in doubt or if it's hard to implement: use `stdout` for everything

PR #1105, opened 2024-08-29
Consistently redirect usage examples to stdout

tinvaan:
Addresses https://github.com/sourcegraph/src-cli/issues/689

### Test plan

Tested manually. Re-directed `src` command and sub-commands outputs for default usage (no args) and with `help`/`--help` args and checked all output was captured.

### DIFF
--- a/cmd/src/cmd.go
+++ b/cmd/src/cmd.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"log"
 	"os"
+	"slices"
 	"strings"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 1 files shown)


peterguy:
Sending error and usage output to `stderr` is common practice for command-line tools because historically many of them were designed for chaining, where the `stdout` of one could be sent to another, so printing help/usage to `stdout` could break that chaining.

That's probably why the default channel for `FlagSet` is `stderr`.

I don't think that `src` is designed for chaining, so changing help/usage output to `stdout` should be ok, but I'll check internally.

peterguy:
@camdencheek pointed out that even if `src` is used in chaining (might be for things like `src search`), sending help/usage to `stdout` doesn't break that, and errors are still sent to `stderr`

peterguy (approved):
Confirmed that errors are still sent to `stderr`. Help/usage output is now sent to `stdout`.

tinvaan:
Thanks for the review @peterguy! Can you help merge this changeset? I don't have write access.

peterguy:
@tinvaan done!

PR #1120, opened 2024-10-21
Add proxy support

peterguy:
This PR adds support for explicit HTTP(S), SOCKS5 and UNIX Domain Socket proxies that can be set using the `SRC_PROXY` environment variable:
```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
SRC_PROXY=https://localhost:8080 \
src login --insecure-skip-verify=true

✔️  Authenticated as peter.guy on https://sourcegraph.sourcegraph.com
```

### Test plan

CI tests all pass.

Added more config tests.

#### Manual tests

*Option 1:*

Use `dev/test-proxies.sh`. 
Follow the instructions in comments at the beginning, then run the script.

*Option 2:*

0. Install `socat` and `mitmproxy`
```
brew install socat mitmproxy
```
3. In two separate terminals, run:
```
socat -d -d unix-listen:${HOME}/src-proxy.sock,fork tcp:localhost:8080
```
and
```
mitmproxy -v
```
4. Check out this branch and build `src-cli`:
```
go build -o ~/go/bin/src ./cmd/src
```
5. Run `src`, setting the endpoint, PAT, and UDS path in the environment variables:
```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
SRC_PROXY=~/src-proxy.sock \
~/go/bin/src login --insecure-skip-verify=true
```
- You need to use the `--insecure-skip-verify=true` flag because `mitmproxy` uses a self-signed certificate. (You can get rid of that by [importing the `mitmproxy` CA cert into your local cert store](https://docs.mitmproxy.org/stable/concepts-certificates/))
- You should get a successful "Authenticated as ..." message. If the PAT is wrong, you'll see "Problem: Invalid access token."
- Look in the terminal where `socat` is running. You'll see output showing connections being made.
- Look in the terminal running `mitmproxy`. You'll see output showing connections being made.

Repeat those tests using the `mitmproxy` HTTP proxy directly:
```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
SRC_PROXY=https://localhost:8080 \
~/go/bin/src login --insecure-skip-verify=true
```

To test SOCKS5 proxies, you can use a Docker-based SOCKS5 proxy.
Run this command in a terminal window:
```
docker run --rm -p 1080:1080 serjs/go-socks5-proxy
```
Then in another terminal window, run `src` pointing at that proxy:
```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
SRC_PROXY=socks5://localhost:1080 \
~/go/bin/src login
```
You should see the same success messages as before, and in the terminal running the `docker` command, you should see connections being made.

We support proxy authentication with HTTP(S) and SOCKS5 proxies.

To test that, run `mitmproxy` and `serjs/go-socks-proxy` with authentication specified:
```
mitmproxy -v --proxyauth user:pass
```
```
docker run --rm -p 1080:1080 -e PROXY_USER=user -e PROXY_PASSWORD=pass serjs/go-socks5-proxy
```
Then add those credentials to the `SRC_PROXY` URL: either `https://user:pass@localhost:8080` or `socks5://user:pass@localhost:1080`

Also confirm that non-proxied connections can be made by omitting the `SRC_PROXY` environment variable setting.

### DIFF
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -7,7 +7,6 @@ import (
 	"log"
 	"net"
 	"os"
-	"os/user"
 	"path/filepath"
 	"strings"
 
@@ -229,11 +228,11 @@ func expandHomeDir(filePath string) (string, error) {
 		if testHomeDir != "" {
 			homeDir = testHomeDir
 		} else {
-			u, err := user.Current()
+			hd, err := os.UserHomeDir()
 			if err != nil {
 				return "", err
 			}
-			homeDir = u.HomeDir
+			homeDir = hd
 		}
 
 		if strings.HasPrefix(filePath, "~/") {
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"log"
 	"net"
+	"net/url"
 	"os"
 	"path/filepath"
 	"strings"
@@ -23,11 +24,22 @@ Usage:
 	src [options] command [command options]
 
 Environment variables
-	SRC_ACCESS_TOKEN  Sourcegraph access token
-	SRC_ENDPOINT      endpoint to use, if unset will default to "https://sourcegraph.com"
-	SRC_PROXY_SOCKET  path to a unix domain socket to use for proxying requests to the
-                      Sourcegraph instance. Can be either a full path, or contain the prefix
-                      ~/ or %USERPROFILE%\ to indicate a path in the user's home directory.
+	SRC_ACCESS_TOKEN    Sourcegraph access token
+	SRC_ENDPOINT        endpoint to use, if unset will default to "https://sourcegraph.com"
+	SRC_PROXY_ENDPOINT  A proxy to use for proxying requests to the Sourcegraph endpoint.
+	                    Supports HTTP(S), SOCKS5/5h, and UNIX Domain Socket proxies.
+						If a UNIX Domain Socket, the path can be either an absolute path, 
+						or can start with ~/ or %USERPROFILE%\ for a path in the user's home directory.
+					   Examples:
+					     - https://localhost:3080
+					     - https://<user>:<password>localhost:8080
+						 - socks5h://localhost:1080
+						 - socks5://<username>:<password>@localhost:1080
+						 - unix://~/src-proxy.sock
+						 - unix://%USERPROFILE%\src-proxy.sock
+						 - ~/src-proxy.sock
+						 - %USERPROFILE%\src-proxy.sock
+						 - C:\some\path\src-proxy.sock
 
 The options are:
 
@@ -86,9 +98,10 @@ type config struct {
 	Endpoint          string            `json:"endpoint"`
 	AccessToken       string            `json:"accessToken"`
 	AdditionalHeaders map[string]string `json:"additionalHeaders"`
-	ProxySocket       string            `json:"proxySocket"`
-
-	ConfigFilePath string
+	ProxyEndpoint     string            `json:"proxyEndpoint"`
+	ProxyEndpointURL  *url.URL
+	ProxyEndpointPath string
+	ConfigFilePath    string
 }
 
 // apiClient returns an api.Client built from the configuration.
@@ -99,6 +112,8 @@ func (c *config) apiClient(flags *api.Flags, out io.Writer) api.Client {
 		AdditionalHeaders: c.AdditionalHeaders,
 		Flags:             flags,
 		Out:               out,
+		ProxyEndpointURL:  c.ProxyEndpointURL,
+		ProxyEndpointPath: c.ProxyEndpointPath,
 	})
 }
 
@@ -130,7 +145,7 @@ func readConfig() (*config, error) {
 
 	envToken := os.Getenv("SRC_ACCESS_TOKEN")
 	envEndpoint := os.Getenv("SRC_ENDPOINT")
-	envProxySocket := os.Getenv("SRC_PROXY_SOCKET")
+	envProxyEndpoint := os.Getenv("SRC_PROXY_ENDPOINT")
 
 	if userSpecified {
 		// If a config file is present, either zero or both environment variables must be present.
@@ -153,21 +168,63 @@ func readConfig() (*config, error) {
 	if cfg.Endpoint == "" {
 		cfg.Endpoint = "https://sourcegraph.com"
 	}
-	if envProxySocket != "" {
-		cfg.ProxySocket = envProxySocket
+	if envProxyEndpoint != "" {
+		cfg.ProxyEndpoint = envProxyEndpoint
 	}
 
-	if cfg.ProxySocket != "" {
-		cfg.ProxySocket, err = expandHomeDir(cfg.ProxySocket)
-		if err != nil {
-			return nil, err
+	if cfg.ProxyEndpoint != "" {
+
+		parseEndpoint := func(endpoint string) (scheme string, address string) {
+			parts := strings.SplitN(endpoint, "://", 2)
+			if len(parts) == 2 {
+				return parts[0], parts[1]
+			}
+			return "", endpoint
 		}
-		isValidUDS, err := isValidUnixSocket(cfg.ProxySocket)
-		if err != nil {
-			return nil, err
+
+		schemeInfo := func(scheme string) (isValid bool, isFile bool) {
+			urlSchemes := []string{"http", "https", "socks", "socks5", "socks5h"}
+			for _, s := range urlSchemes {
+				if scheme == s {
+					return true, false
+				}
+			}
+			if scheme == "unix" {
+				return true, true
+			}
+			return false, false
 		}
-		if !isValidUDS {
-			return nil, errors.Newf("invalid proxy socket: %s", cfg.ProxySocket)
+
+		scheme, address := parseEndpoint(cfg.ProxyEndpoint)
+
+		isValidScheme, isFileScheme := schemeInfo(scheme)
+
+		if !isValidScheme {
+			return nil, errors.Newf("invalid proxy endpoint: %s", cfg.ProxyEndpoint)
+		}
+
+		if !isFileScheme {
+			// assume socks means socks5, because that's all we support
+			if scheme == "socks" {
+				cfg.ProxyEndpoint = "socks5://" + address
+			}
+			cfg.ProxyEndpointURL, err = url.Parse(cfg.ProxyEndpoint)
+			if err != nil {
+				return nil, err
+			}
+		} else {
+			path, err := expandHomeDir(address)
+			if err != nil {
+				return nil, err
+			}
+			isValidUDS, err := isValidUnixSocket(path)
+			if err != nil {
+				return nil, err
+			}
+			if !isValidUDS {
+				return nil, errors.Newf("invalid proxy socket: %s", path)
+			}
+			cfg.ProxyEndpointPath = path
 		}
 	}
 
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -101,29 +101,20 @@ func (c *config) apiClient(flags *api.Flags, out io.Writer) api.Client {
 	})
 }
 
-var testHomeDir string // used by tests to mock the user's $HOME
-
 // readConfig reads the config file from the given path.
 func readConfig() (*config, error) {
-	cfgPath := *configPath
+	cfgFile := *configPath
 	userSpecified := *configPath != ""
 
-	var homeDir string
-	if testHomeDir != "" {
-		homeDir = testHomeDir
-	} else {
-		u, err := user.Current()
-		if err != nil {
-			return nil, err
-		}
-		homeDir = u.HomeDir
+	if !userSpecified {
+		cfgFile = "~/src-config.json"
 	}
 
-	if !userSpecified {
-		cfgPath = filepath.Join(homeDir, "src-config.json")
-	} else if strings.HasPrefix(cfgPath, "~/") {
-		cfgPath = filepath.Join(homeDir, cfgPath[2:])
+	cfgPath, err := expandHomeDir(cfgFile)
+	if err != nil {
+		return nil, err
 	}
+
 	data, err := os.ReadFile(os.ExpandEnv(cfgPath))
 	if err != nil && (!os.IsNotExist(err) || userSpecified) {
 		return nil, err
@@ -167,6 +158,20 @@ func readConfig() (*config, error) {
 	}
 	if envProxyURL != "" {
 		cfg.ProxyURL = envProxyURL
+	}
+
+	if cfg.ProxySocket != "" {
+		cfg.ProxySocket, err = expandHomeDir(cfg.ProxySocket)
+		if err != nil {
+			return nil, err
+		}
+		isValidUDS, err := isValidUnixSocket(cfg.ProxySocket)
+		if err != nil {
+			return nil, err
+		}
+		if !isValidUDS {
+			return nil, errors.Newf("invalid proxy socket: %s", cfg.ProxySocket)
+		}
 	}
 
 	cfg.AdditionalHeaders = parseAdditionalHeaders()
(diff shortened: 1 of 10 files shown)


internal/api/proxy.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		// Be sure to clone TLS-specific config settings from the transport,
+		// like InsecureSkipVerify.
+		tlsConn := tls.Client(conn, &tls.Config{
+			ServerName:         host,
+			InsecureSkipVerify: transport.TLSClientConfig.InsecureSkipVerify,
+		})
+		if err := tlsConn.Handshake(); err != nil {
+			return nil, err
+		}
```
camdencheek, line 35: We should use the `HandshakeContext` variant here

internal/api/proxy.go
```diff
@@ -0,0 +1,139 @@
...
+		return tlsConn, nil
+	}
+
+	var dial func(ctx context.Context, network, addr string) (net.Conn, error)
+	var dialTLS func(ctx context.Context, network, addr string) (net.Conn, error)
+
+	if proxyEndpointPath != "" {
+		dial = func(_ context.Context, _, _ string) (net.Conn, error) {
+			return net.Dial("unix", proxyEndpointPath)
```
camdencheek, line 44: Also `net.DialContext` here (and everywhere else we call `Dial`)

internal/api/proxy.go
```diff
@@ -0,0 +1,139 @@
...
+			}
+			dialTLS = func(ctx context.Context, network, addr string) (net.Conn, error) {
+				// Dial the underlying connection through the proxy
+				conn, err := dial(ctx, network, addr)
+				if err != nil {
+					return nil, err
+				}
+				return handshakeTLS(conn, addr)
+			}
+		} else if proxyEndpointURL.Scheme == "http" || proxyEndpointURL.Scheme == "https" {
```
camdencheek, line 78: Why are we manually implementing this? It looks to me like is all supported by the standard library by setting `(*http.Transport).Proxy`. 

In fact, I think everything but sockets is likely already supported with no code changes by setting `HTTP_PROXY` and `HTTPS_PROXY`. Take a look at `DefaultTransport`, which defaults to respecting [httpproxy.FromEnvironment](https://pkg.go.dev/golang.org/x/net/http/httpproxy#FromEnvironment).

peterguy: Prompted by your question, I tested socks proxies and they do work out of the box.

I hadn't tested them - I started with testing https proxy, which didn't work, so I just dove into implementing them all.

The error I get when using https proxies OOTB (I've used `mitmproxy` locally and https://devproxy.sgdev.org/) is:
```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
https_proxy=https://localhost:8080 \
src login --insecure-skip-verify=true 
❌ Problem: Error communicating with https://sourcegraph.sourcegraph.com: Post "https://sourcegraph.sourcegraph.com/.api/graphql": malformed HTTP status code "protocol"
```
When using `devproxy.sgdev.org`, I see the error `proxyconnect tcp: tls: first record does not look like a TLS handshake`.

When using `mitmproxy` locally, there's a corresponding error `HTTP/2 protocol error: Invalid HTTP/2 preamble.` in its output.

Note that this is the current release of `src` that I'm using here, not a local build (although I get the same behavior with a local build).

Manually dialing and doing the CONNECT for the https proxy fixes the protocol errors. There might be a cleaner way to fix them.

internal/api/proxy.go
peterguy: Those dial functions are applied to `transport.DialContext` and `transport.DialTLSContext`. They're just named without "Context" - poor name choice perhaps.

internal/api/proxy.go
peterguy: Yep, good catch!

internal/api/api.go
```diff
@@ -95,10 +99,19 @@ func NewClient(opts ClientOpts) Client {
...
-			Transport: &http.Transport{TLSClientConfig: &tls.Config{InsecureSkipVerify: true}},
-		}
+		customTransport = true
+		transport.TLSClientConfig = &tls.Config{InsecureSkipVerify: true}
+	}
+
+	proxyApplied := applyProxy(transport, opts.ProxyEndpointURL, opts.ProxyEndpointPath)
+
+	if customTransport || proxyApplied {
+		httpClient.Transport = transport
```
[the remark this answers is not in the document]
peterguy: Hmmm, no, that was an oversight. I'm not sure it matters in the case of a short-lived CLI process, but I've restored the  original behavior of creating a new client if we're modifying the transport.

cmd/src/main.go
```diff
@@ -157,6 +169,60 @@ func readConfig() (*config, error) {
...
+			if err != nil {
+				return nil, err
+			}
+			isValidUDS, err := isValidUnixSocket(path)
+			if err != nil {
+				return nil, err
+			}
+			if !isValidUDS {
+				return nil, errors.Newf("invalid proxy socket: %s", path)
+			}
```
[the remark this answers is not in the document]
peterguy: The intent was to provide a better error message.

Dial failures result in errors like, "dial unix /foo/bar/path: connect: no such file or directory", or "connect: socket operation on non-socket" and I wanted to provide something better up front.

I didn't completely deliver on that because the error messages during config validation aren't great either. I can make the config validation error messages better, but also, I'm not stuck on doing this validation up front.

cmd/src/main_test.go
```diff
@@ -2,24 +2,39 @@ package main
...
 	"os"
 	"path/filepath"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/sourcegraph/src-cli/internal/api"
 )
 
 func TestReadConfig(t *testing.T) {
+	socketPath, err := api.CreateTempFile(os.TempDir(), "TestReadConfig_", ".sock")
```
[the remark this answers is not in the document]
peterguy: I was hoping to use `t.TempDir()`, but the config validator could never find the file when I did. I didn't put too much effort into figuring out why; I can give it a second pass.

internal/api/proxy.go
camdencheek: Right! But the `net.Dial` call inside the function doesn't use the passed-in context, so will not respect any timeouts

cmd/src/main_test.go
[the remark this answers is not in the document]
camdencheek: Huh, interesting! Not a big deal either way, but if you do take a second pass I'd be interested to know what was going no

cmd/src/main.go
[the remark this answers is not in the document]
camdencheek: Ah, got it. Makes sense! No real preference here, feel free to ignore

internal/api/proxy.go
peterguy: Oh, I see. Interestingly enough, there is no `net.DialContext` function - it's a method on `Dialer`, so I have to define a `Dialer`, then call `DialContext` on it. FWIW, `net.Dial` does the same thing.

camdencheek: Oh interesting! That's surprising to me. TIL

internal/api/proxy.go
peterguy: Hmmm, legit point about `--insecure-skip-verify`, but I am not able to fix the errors the way you describe, and I get similar errors even when I use a proxy that doesn't require `--insecure-skip-verify`, like https://devproxy.sgdev.org:3128, which has a public CA cert, or `tinyproxy`, which is a tunneling proxy that does not require its own CA cert:
```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
https_proxy=https://devproxy.sgdev.org:3128 \
~/go/bin/src login

❌ Problem: Error communicating with https://sourcegraph.sourcegraph.com: Post "https://sourcegraph.sourcegraph.com/.api/graphql": proxyconnect tcp: tls: first record does not look like a TLS handshake
```

```
SRC_ENDPOINT=https://sourcegraph.sourcegraph.com \
SRC_ACCESS_TOKEN=deadbeef \
https_proxy=https://localhost:8888 \         
~/go/bin/src login

❌ Problem: Error communicating with https://sourcegraph.sourcegraph.com: Post "https://sourcegraph.sourcegraph.com/.api/graphql": proxyconnect tcp: net/http: TLS handshake timeout
```
The connection with `tinyproxy` actually times out.

All of these errors are solved  when I use manual dialing, by changing from `https_proxy` to `SRC_PROXY_ENDPOINT`.

I don't want to be pedantic about this, but I am interested in understanding the actual behavior and resolution, if possible.

camdencheek: > I don't want to be pedantic about this

Proxy behavior is exactly the kind of thing that is worth being pedantic about IMO 😄

peterguy: Fair point. :-)

I tried importing the cert into the login keychain (and marking it as "Always trust"), and while that affects the behavior when using `SRC_PROXY_ENDPOINT`, it does not when using `https_proxy`.

camdencheek: I'm able to get `http://devproxy.sgdev.org:3128` working, but both `src` and `curl` fail when trying to use `https://devproxy.sgdev.org:3129` for me with an "expired certificate" error

camdencheek: Looks to me like that's correct, and that the cert is, in fact, expired. Do you know who runs the devproxy?

camdencheek: I confirmed that `src` works against `https://devproxy.sgdev.org:3129` if I set `--insecure-skip-verify` (with the fix for the insecure transport), as does `curl` if I also set `--proxy-insecure`

cmd/src/main_test.go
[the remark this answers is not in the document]
peterguy: Turns out it was the length of the file path. UNIX Domain Socket path lengths are limited. It varies, but it's around 100 characters. The ones that failed were about 107 characters (example: `/var/folders/20/1r11xzwn4xs3xyw9bncsqq3c0000gn/T/TestReadConfig124057903/001/TestReadConfig_3992591027.sock`). I removed the prefix and suffix from the temp file creation and switched to using `t.TempDir()`

internal/api/proxy.go
peterguy: You were turning off HTTP/2 by using `GODEBUG=http2client=0`. The upshot of our long discussion and experimentation about this is that we're manually dialing in order to use the HTTP/1 protocol instead of HTTP/2.

I added a bunch of comments to the code to help our future selves remember this.

cmd/src/main.go
```diff
@@ -178,3 +244,65 @@ func readConfig() (*config, error) {
...
+// If the connection succeeds, it's considered a valid Unix socket.
+// If the file doesn't exist, it returns false without an error.
+// For any other errors, it returns false and the encountered error.
+func isValidUnixSocket(path string) (bool, error) {
+	conn, err := net.Dial("unix", path)
+	if err != nil {
+		if os.IsNotExist(err) {
+			return false, nil
+		}
+		return false, errors.Newf("Not a UNIX Domain Socket: %v: %v", path, err)
```
camdencheek, line 267:
```suggestion
		return false, errors.Newf("Not a UNIX Domain Socket: %v: %w", path, err)
```

dev/test-proxies.sh
camdencheek, line 1: Thanks for committing and documenting your test script!

internal/api/proxy.go
```diff
@@ -0,0 +1,138 @@
...
+		host, _, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, err
+		}
+		// Be sure to clone TLS-specific config settings from the transport,
+		// like InsecureSkipVerify.
+		tlsConn := tls.Client(conn, &tls.Config{
+			ServerName:         host,
+			InsecureSkipVerify: transport.TLSClientConfig.InsecureSkipVerify,
+		})
```
camdencheek, line 30: Are there any other TLS configs we need to copy? Woudl it be safer to do something like this?

```suggestion
		tlsConfig := *transport.TLSClientConfig
		tlsConfig.ServerName = host
		tlsConn := tls.Client(conn, &tlsConfig)
```

cmd/src/main.go
```diff
@@ -25,6 +26,20 @@ Usage:
...
 	SRC_ENDPOINT      endpoint to use, if unset will default to "https://sourcegraph.com"
+	SRC_PROXY         A proxy to use for proxying requests to the Sourcegraph endpoint.
+	                  Supports HTTP(S), SOCKS5/5h, and UNIX Domain Socket proxies.
+					  If a UNIX Domain Socket, the path can be either an absolute path, 
+					  or can start with ~/ or %USERPROFILE%\ for a path in the user's home directory.
+					  Examples:
+					    - https://localhost:3080
+					    - https://<user>:<password>localhost:8080
+						- socks5h://localhost:1080
+						- socks5://<username>:<password>@localhost:1080
```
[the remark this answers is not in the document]
peterguy: Whoops, yep: there was a weird mixture of tabs and spaces. Fixed it.

cmd/src/main.go
```diff
@@ -83,8 +98,10 @@ type config struct {
 	Endpoint          string            `json:"endpoint"`
 	AccessToken       string            `json:"accessToken"`
 	AdditionalHeaders map[string]string `json:"additionalHeaders"`
-
-	ConfigFilePath string
+	Proxy             string            `json:"proxy"`
+	ProxyURL          *url.URL
+	ProxyPath         string
+	ConfigFilePath    string
```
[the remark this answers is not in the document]
peterguy: Correct: not exposed to the config file - just there as a way to pass validated input on to `api.NewClient`. TBH, I just copied the `ConfigFilePath` pattern for `ProxyURL` and `ProxyPath` - it feels kind of weird to me, but it was easier and simpler than coming up with another struct, or using global variables, or similar.

cmd/src/main.go
```diff
@@ -157,6 +169,60 @@ func readConfig() (*config, error) {
...
+				return nil, err
+			}
+		} else if scheme == "" || scheme == "unix" {
+			path, err := expandHomeDir(address)
+			if err != nil {
+				return nil, err
+			}
+			isValidUDS, err := isValidUnixSocket(path)
+			if err != nil {
+				return nil, errors.Newf("Invalid proxy configuration: %v", err)
```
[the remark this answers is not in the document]
peterguy: Whoops; meant to use `w` instead of `v`! Thanks for catching that.

cmd/src/main.go
peterguy: Whoops; meant to use `w` instead of `v`! Thanks for catching that.

cmd/src/main_test.go
```diff
@@ -2,21 +2,39 @@ package main
...
 
 	"github.com/google/go-cmp/cmp"
+
+	"github.com/sourcegraph/src-cli/internal/api"
 )
 
 func TestReadConfig(t *testing.T) {
+	// UNIX Domain Sockets have a max path length: 104 on BSD/macOS, 108 on Linux.
+	// Including a prefix and suffix was causing the path to be too long
+	// with t.TempDir() (os.TempDir() is a shorter path) so we don't use them.
```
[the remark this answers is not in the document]
peterguy: Yep. Lost a lot of hair over that one because the error message is something like, "bind: not a valid operation".

cmd/src/main.go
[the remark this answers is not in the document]
camdencheek: Got it, sounds good. Just making sure it was intentional to not make these configurable with the config file

internal/api/proxy.go
peterguy: I don't know that we really want to share things like client certs and session tickets. Let's stick with `InsecureSkipVerify` for now and I'll update the comment.

peterguy:
Thanks @camdencheek for the thorough review and extra-mile pairing!

PR #1124, opened 2024-11-11
feat/gateway: Add `src gateway benchmark` command

vdavid:
Adds the `src gateway benchmark` command to the tool.

https://github.com/sourcegraph/sourcegraph/pull/1526 adds some new endpoints that this benchmarking script uses.

The benchmarking tool compares ~three~ two routes using the endpoints added in above PR:
1. Sourcegraph HTTP → Cody Gateway HTTP
2. Sourcegraph HTTP → Cody Gateway WebSocket
3. (There should be a third one: Sourcegraph WebSocket → Cody Gateway WebSocket, but I haven't been able to make that endpoint work so it's currently not added to this script.)

### Testing the benchmarking tool locally

(Description WIP)

- `brew install ngrok/ngrok/ngrok` if you don't have ngrok
- Run `ngrok http http://localhost:9992`
- Copy the `*.ngrok-free.app` URL from the "Forwarding" section of ngrok's output to your `dev-private/enterprise/dev/site-config.json` as the "modelConfiguration" / "sourcegraph" / "endpoint" value.
- **Important note:** Actually, when using ngrok, the protocol when calling Cody Gateway in the new endpoint handler has to be `wss://`, not `ws://`! (Just a one-char change but it's manual for now.) But when running locally, it needs to be `ws://`. Something to be mindful of when running this and in the future.
- Run Sourcegraph: `run sg dotcom`
- Run: `src gateway benchmark`

The output should look like this:

CleanShot 2024-11-07 at 15 01 49@2x

### Next steps

Missing work to be done:
- Add a direct call to Cody Gateway. Endpoints are at:
  - `{codyGatewayUrl}/v2/http` and
  - `{codyGatewayUrl}/v2/websocket`.
- Docs?

### Test plan

Tested manually:

1. ✅ `go run ./cmd/src` lists the new `gateway` command correctly.
2. ✅ `go run ./cmd/src gateway` lists the `benchmark` subcommand correctly.
3. ✅ `go run ./cmd/src gateway benchmark` runs the benchmarks against the endpoint set in the `SRC_ENDPOINT` env var correctly.
4. ✅ `go run ./cmd/src gateway benchmark --requests 2` sets the number of requests correctly.
5. ✅ `SRC_ENDPOINT=http://localhost:3080 go run ./cmd/src gateway benchmark --requests 2` sets the endpoint to my localhost and works correctly.

### DIFF
--- a/cmd/src/gateway.go
+++ b/cmd/src/gateway.go
@@ -8,7 +8,7 @@ import (
 var gatewayCommands commander
 
 func init() {
-	usage := `'src gateway' manages Cody Gateway operations on a Sourcegraph instance.
+	usage := `'src gateway' interacts with Cody Gateway (directly or through a Sourcegraph instance).
 
 Usage:
 
@@ -30,9 +30,9 @@ Use "src gateway [command] -h" for more information about a command.
 
 	// Register the command.
 	commands = append(commands, &command{
-		flagSet: flagSet,
-		aliases: []string{},  // No aliases for gateway command
-		handler: handler,
+		flagSet:   flagSet,
+		aliases:   []string{}, // No aliases for gateway command
+		handler:   handler,
 		usageFunc: func() { fmt.Println(usage) },
 	})
 }
--- a/cmd/src/gateway.go
+++ b/cmd/src/gateway.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+)
+
+var gatewayCommands commander
+
+func init() {
+	usage := `'src gateway' manages Cody Gateway operations on a Sourcegraph instance.
+
+Usage:
+
+	src gateway command [command options]
+
+The commands are:
+
+	benchmark           runs benchmarks against Cody Gateway
+
+Use "src gateway [command] -h" for more information about a command.
+
+`
+
+	flagSet := flag.NewFlagSet("gateway", flag.ExitOnError)
+	handler := func(args []string) error {
+		gatewayCommands.run(flagSet, "src gateway", usage, args)
+		return nil
+	}
+
+	// Register the command.
+	commands = append(commands, &command{
+		flagSet: flagSet,
+		aliases: []string{},  // No aliases for gateway command
+		handler: handler,
+		usageFunc: func() { fmt.Println(usage) },
+	})
+}
--- a/cmd/src/gateway_benchmark.go
+++ b/cmd/src/gateway_benchmark.go
@@ -13,15 +13,6 @@ import (
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
 )
 
-const (
-	defaultRequestCount = 100
-	blue                = "\033[34m"
-	green       = "\033[32m"
-	yellow      = "\033[33m"
-	red         = "\033[31m"
-	reset       = "\033[0m"
-)
-
 func init() {
 	usage := `
 'src gateway benchmark' runs performance benchmarks against Cody Gateway endpoints.
@@ -39,7 +30,7 @@ Examples:
 	flagSet := flag.NewFlagSet("benchmark", flag.ExitOnError)
 
 	var (
-		requestCount = flagSet.Int("requests", defaultRequestCount, "Number of requests to make per endpoint")
+		requestCount = flagSet.Int("requests", 1000, "Number of requests to make per endpoint")
 	)
 
 	handler := func(args []string) error {
@@ -59,8 +50,8 @@ Examples:
 		}
 
 		endpoints := map[string]string{
-			"HTTP":                fmt.Sprintf("%s/cody-gateway-call-http", cfg.Endpoint),
-			"HTTP then WebSocket": fmt.Sprintf("%s/cody-gateway-call-http-then-websocket", cfg.Endpoint),
+			"HTTP":                fmt.Sprintf("%s/gateway", cfg.Endpoint),
+			"HTTP then WebSocket": fmt.Sprintf("%s/gateway/http-then-websocket", cfg.Endpoint),
 		}
 
 		fmt.Printf("Starting benchmark with %d requests per endpoint...\n", *requestCount)
@@ -76,7 +67,6 @@ Examples:
 				if duration > 0 {
 					durations = append(durations, duration)
 				}
-				fmt.Printf("\rTesting %s: %d/%d", name, i+1, *requestCount)
 			}
 			fmt.Println()
 
@@ -136,12 +126,11 @@ func benchmarkEndpoint(client *http.Client, url string) time.Duration {
 		}
 	}(resp.Body)
 
-	body, err := io.ReadAll(resp.Body)
+	_, err = io.ReadAll(resp.Body)
 	if err != nil {
 		fmt.Printf("Error reading response body: %v\n", err)
 		return 0
 	}
-	fmt.Printf("Response from %s: %s\n", url, body)
 
 	return time.Since(start)
 }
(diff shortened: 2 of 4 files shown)


cmd/src/gateway.go
```diff
@@ -0,0 +1,38 @@
...
+
+import (
+	"flag"
+	"fmt"
+)
+
+var gatewayCommands commander
+
+func init() {
+	usage := `'src gateway' manages Cody Gateway operations on a Sourcegraph instance.
```
slimsag, line 11: Since Cody Gateway doesn't run as part of a Sourcegraph instance, I think this phrasing is maybe a bit confusing?

Perhaps "interacts with Cody Gateway directly or through a Sourcegraph instance" may be better?

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+	"github.com/sourcegraph/src-cli/internal/cmderrors"
+)
+
+const (
+	defaultRequestCount = 100
+	blue                = "\033[34m"
+	green       = "\033[32m"
+	yellow      = "\033[33m"
+	red         = "\033[31m"
+	reset       = "\033[0m"
```
slimsag, line 22: I think we should be using the stuff in color.go here: https://sourcegraph.sourcegraph.com/github.com/sourcegraph/src-cli/-/blob/cmd/src/colors.go

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+Examples:
+
+    $ src gateway benchmark
+    $ src gateway benchmark --requests 50
+`
+
+	flagSet := flag.NewFlagSet("benchmark", flag.ExitOnError)
+
+	var (
+		requestCount = flagSet.Int("requests", defaultRequestCount, "Number of requests to make per endpoint")
```
slimsag, line 42: Maybe just write `defaultRequestCount` inline here instead of a constant (easier to see what the defautl is that way, less likely to conflict with other constants / files in the folder)

cmd/src/main.go
```diff
@@ -53,6 +53,7 @@ The commands are:
 	config          manages global, org, and user settings
 	extensions,ext  manages extensions (experimental)
 	extsvc          manages external services
+	gateway         manages Cody Gateway
```
slimsag, line 56:
```suggestion
	gateway         interact with Cody Gateway
```

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+		if len(flagSet.Args()) != 0 {
+			return cmderrors.Usage("additional arguments not allowed")
+		}
+
+		// Create HTTP client with TLS skip verify
+		client := &http.Client{
+			Transport: &http.Transport{
+				TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+			},
+		}
```
slimsag, line 59: I think it's fine if we have a CLI option for `--insecure` to disable TLS for testing against local dev, but this wouldn't match a real prod environment so let's make sure we have TLS verification enabled by default here

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+
+		endpoints := map[string]string{
+			"HTTP":                fmt.Sprintf("%s/cody-gateway-call-http", cfg.Endpoint),
+			"HTTP then WebSocket": fmt.Sprintf("%s/cody-gateway-call-http-then-websocket", cfg.Endpoint),
```
slimsag, line 63: let's work torwards naming these endpoints whatever the real V2 API endpoint would be named

cmd/src/gateway_benchmark.go
slimsag: Also may want to bump this to something statistically significant (e.x. 1000)

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+		for name, url := range endpoints {
+			durations := make([]time.Duration, 0, *requestCount)
+			fmt.Printf("\nTesting %s...", name)
+
+			for i := 0; i < *requestCount; i++ {
+				duration := benchmarkEndpoint(client, url)
+				if duration > 0 {
+					durations = append(durations, duration)
+				}
+				fmt.Printf("\rTesting %s: %d/%d", name, i+1, *requestCount)
```
slimsag, line 79: let's drop this print statement, or make it only print on every 1/10th of the requestCount - to reduce the change that printing each request adds non-deterministic latency.

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+			fmt.Println()
+
+			avg, p5, p95, median, total := calculateStats(durations)
```
slimsag, line 83: let's make calculateStats return a struct with these as fields, so there's no chance we get the order wrong here and look at the wrong thing by accident.

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+			fmt.Printf("Error closing response body: %v\n", err)
+		}
+	}(resp.Body)
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		fmt.Printf("Error reading response body: %v\n", err)
+		return 0
+	}
+	fmt.Printf("Response from %s: %s\n", url, body)
```
slimsag, line 144: similarly, let's drop this print statement in the success path

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+	return time.Since(start)
+}
+
+func calculateStats(durations []time.Duration) (time.Duration, time.Duration, time.Duration, time.Duration, time.Duration) {
+	if len(durations) == 0 {
+		return 0, 0, 0, 0, 0
+	}
+
+	sort.Slice(durations, func(i, j int) bool {
+		return durations[i] < durations[j]
```
slimsag, line 155: adding a note here for me to triple check we got the order right here; just don't want any chance we've got p95 and p5 inverted.

Actually; could you add p80 and p75 as well? that'd make it clear and might be useful to collect anyway.

cmd/src/gateway_benchmark.go
```diff
@@ -0,0 +1,258 @@
...
+	if best {
+		return green + value + reset
+	}
+	if worst {
+		return red + value + reset
+	}
+	return yellow + value + reset
+}
+
+func printResults(results []endpointResult, requestCount *int) {
```
slimsag, line 193: Can we add a CLI flag to dump the results as e.g. a CSV as well? When we hand this out to ~10 people to collect results it might be nicer when we try to import them into e.g. a google sheet

slimsag:
Cool, great starting point! can you get this client talking to / benchmarking 4 endpoints?

e.g. with one test `src gateway benchmark -endpoint=$CODY_GATEWAY_URL -sourcegraph=$SG_INSTANCE_URL` I'd love for it to benchmark:

1. `{CODY_GATEWAY_URL}/v2` (HTTP)
2. `{CODY_GATEWAY_URL}/v2/websocket` (websocket)
3. `{SG_INSTANCE_URL}/.api/gateway` (HTTP)
4. `{SG_INSTANCE_URL}/.api/gateway/websocket` (websocket)

i.e. if the user passes `-endpoint` then run (1) and (2), if the user passes `-sourcegraph` then run (3) and (4), if they pass both then run all 4.

Then the default if you pass nothing can be to run against Cody Gateway prod and Sourcegraph.com

For (2) and (4) you'll need to add a websocket client to this PR, you should be able to do that testing against any websocket server until I have a chance to debug the issue you brought up with doing it in the SG instance.

cmd/src/gateway_benchmark.go
vdavid: Not a fan of inlineing magic values, but we only used it once at the relative top of the file so I guess it's okay. Inlined it and bumped it to 1000.

cmd/src/main.go
vdavid: Fixed to "interacts with"

cmd/src/gateway_benchmark.go
vdavid: Done in https://github.com/sourcegraph/sourcegraph/pull/1526, updated the URLs here to match the new paths

cmd/src/gateway_benchmark.go
vdavid: Dropped it.

cmd/src/gateway_benchmark.go
vdavid: I think we don't need it anyway, the script works without it locally. Removed.

cmd/src/gateway_benchmark.go
vdavid: Great idea, added!

vdavid:
Update:
Fixed all asks in individual messages.

In response to the main message:
- The `-sourcegraph=$SG_INSTANCE_URL` param is missing
- The `-endpoint=$CODY_GATEWAY_URL` param is missing
- The direct calls to Cody Gateway are missing.

It's like 20 min to add these but I haven't had the time. Later.

slimsag (approved):
Cool, this looks good, I'm going to merge as-is and we can continue iterating in `main` together

PR #1125, opened 2024-11-15
update Go version; add websocket / correct endpoints to 'src gateway benchmark'

slimsag:
Continues development of #1124

* Updates Go version used in this repository (I had to do this to develop, see https://sourcegraph.slack.com/archives/C07KZF47K/p1731627487087909)
* Checks the status code returned by HTTP endpoints.
* Updates Sourcegraph HTTP endpoints to `/.api/gateway` and `/.api/gateway/websocket`.
* Updates gateway HTTP endpoints to `/v2` and `/v2/websocket`
* Adds websocket client.
* Respects `-gateway` parameter and `SRC_ENDPOINT`/`SRC_ACCESS_TOKEN`

This should be sufficient for testing against the actual production endpoints once they are added / working / live in prod.

### Test plan

Not tested yet since endpoints are not live yet. Command not in use yet, so no risk.

### DIFF
--- a/cmd/src/gateway_benchmark.go
+++ b/cmd/src/gateway_benchmark.go
@@ -152,6 +152,10 @@ func benchmarkEndpoint(client *http.Client, url string) time.Duration {
 		fmt.Printf("Error reading response body: %v\n", err)
 		return 0
 	}
+	if resp.StatusCode != http.StatusOK {
+		fmt.Printf("non-200 response: %v\n", resp.Status)
+		return 0
+	}
 
 	return time.Since(start)
 }
--- a/cmd/src/gateway_benchmark.go
+++ b/cmd/src/gateway_benchmark.go
@@ -62,7 +62,7 @@ Examples:
 			gatewayWebsocket, sourcegraphWebsocket *websocket.Conn
 			err                                    error
 			httpClient                             = &http.Client{}
-			endpoints                              = map[string]any{}
+			endpoints                              = map[string]any{} // Values: URL `string`s or `*websocket.Conn`s
 		)
 		if *gatewayEndpoint != "" {
 			wsURL := strings.Replace(fmt.Sprint(*gatewayEndpoint, "/v2/websocket"), "http", "ws", 1)
@@ -70,8 +70,8 @@ Examples:
 			if err != nil {
 				return fmt.Errorf("WebSocket dial(%s): %v", wsURL, err)
 			}
-			endpoints["wss: gateway"] = gatewayWebsocket
-			endpoints["https: gateway"] = fmt.Sprint(*gatewayEndpoint, "/v2")
+			endpoints["ws(s): gateway"] = gatewayWebsocket
+			endpoints["http(s): gateway"] = fmt.Sprint(*gatewayEndpoint, "/v2")
 		}
 		if cfg.Endpoint != "" {
 			wsURL := strings.Replace(fmt.Sprint(cfg.Endpoint, "/.api/gateway/websocket"), "http", "ws", 1)
@@ -79,8 +79,8 @@ Examples:
 			if err != nil {
 				return fmt.Errorf("WebSocket dial(%s): %v", wsURL, err)
 			}
-			endpoints["wss: sourcegraph"] = sourcegraphWebsocket
-			endpoints["https: sourcegraph"] = fmt.Sprint(*gatewayEndpoint, "/.api/gateway")
+			endpoints["ws(s): sourcegraph"] = sourcegraphWebsocket
+			endpoints["http(s): sourcegraph"] = fmt.Sprint(*gatewayEndpoint, "/.api/gateway")
 		}
 
 		fmt.Printf("Starting benchmark with %d requests per endpoint...\n", *requestCount)
@@ -182,6 +182,15 @@ func benchmarkEndpointHTTP(client *http.Client, url string) time.Duration {
 		fmt.Printf("non-200 response: %v\n", resp.Status)
 		return 0
 	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		fmt.Printf("Error reading response body: %v\n", err)
+		return 0
+	}
+	if string(body) != "pong" {
+		fmt.Printf("Expected 'pong' response, got: %q\n", string(body))
+		return 0
+	}
 
 	return time.Since(start)
 }
@@ -199,7 +208,7 @@ func benchmarkEndpointWebSocket(conn *websocket.Conn) time.Duration {
 		return 0
 	}
 	if string(message) != "pong" {
-		fmt.Printf("Expected 'pong' response, got: %s\n", string(message))
+		fmt.Printf("Expected 'pong' response, got: %q\n", string(message))
 		return 0
 	}
 	return time.Since(start)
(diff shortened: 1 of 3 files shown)


cmd/src/gateway_benchmark.go
```diff
@@ -55,26 +58,49 @@ Examples:
...
-		client := &http.Client{Transport: &http.Transport{}}
-
-		endpoints := map[string]string{
-			"HTTP":                fmt.Sprintf("%s/gateway", cfg.Endpoint),
-			"HTTP then WebSocket": fmt.Sprintf("%s/gateway/http-then-websocket", cfg.Endpoint),
+		var (
+			gatewayWebsocket, sourcegraphWebsocket *websocket.Conn
+			err                                    error
+			httpClient                             = &http.Client{}
+			endpoints                              = map[string]any{}
```
vdavid, line 65: I think this would help clarify that `any`.

```suggestion
			endpoints                              = map[string]any{} // Values: URL `string`s or `*websocket.Conn`s
```

cmd/src/gateway_benchmark.go
```diff
@@ -55,26 +58,49 @@ Examples:
...
+		)
+		if *gatewayEndpoint != "" {
+			wsURL := strings.Replace(fmt.Sprint(*gatewayEndpoint, "/v2/websocket"), "http", "ws", 1)
```
vdavid, line 68: Ah, the `http` → `ws` replacement is nice and elegant to keep that optional `s`. Self-explanatory once I saw it, but I missed it in my original version. Thanks! 👏

cmd/src/gateway_benchmark.go
```diff
@@ -55,26 +58,49 @@ Examples:
...
+			gatewayWebsocket, _, err = websocket.DefaultDialer.Dial(wsURL, nil)
+			if err != nil {
+				return fmt.Errorf("WebSocket dial(%s): %v", wsURL, err)
+			}
+			endpoints["wss: gateway"] = gatewayWebsocket
+			endpoints["https: gateway"] = fmt.Sprint(*gatewayEndpoint, "/v2")
```
vdavid, line 74: Nit, maybe this would be more precise? We output these names, so might be worth adjusting.

```suggestion
			endpoints["ws(s): gateway"] = gatewayWebsocket
			endpoints["http(s): gateway"] = fmt.Sprint(*gatewayEndpoint, "/v2")
```
(same for sourcegraph endpoints below)

go.sum
vdavid, line 1: I think this update messed up something, see the check annotations below. Also, would be nice to mention in the PR description that you updated packages, and state the reason (incl. whether it was related or unrelated to the main purpose of the PR).

cmd/src/gateway_benchmark.go
```diff
@@ -133,15 +159,15 @@ type endpointResult struct {
...
 	start := time.Now()
 	resp, err := client.Get(url)
 	if err != nil {
 		fmt.Printf("Error calling %s: %v\n", url, err)
 		return 0
 	}
-	defer func(Body io.ReadCloser) {
-		err := Body.Close()
+	defer func(body io.ReadCloser) {
+		err := body.Close()
```
vdavid, line 170: Ah, thanks! This was IntelliJ's autogenerated code. Will watch out for this in the future.

cmd/src/gateway_benchmark.go
```diff
@@ -152,7 +178,30 @@ func benchmarkEndpoint(client *http.Client, url string) time.Duration {
 		fmt.Printf("Error reading response body: %v\n", err)
 		return 0
 	}
+	if resp.StatusCode != http.StatusOK {
+		fmt.Printf("non-200 response: %v\n", resp.Status)
+		return 0
+	}
```
vdavid, line 184: Maybe check the body as well? I think it's a great call that you checked for response body == "pong" for WebSockets. I think we should do the same here. I'll adjust the endpoints to respond with "pong"

cmd/src/gateway_benchmark.go
```diff
@@ -152,7 +178,30 @@ func benchmarkEndpoint(client *http.Client, url string) time.Duration {
...
+
+	return time.Since(start)
+}
 
+func benchmarkEndpointWebSocket(conn *websocket.Conn) time.Duration {
```
vdavid, line 189: Naming: I'd have called this `benchmarkEndpointWebsocket` with a non-capital "s". I looked up the Go docs and your version seems more idiomatic with Go.
If we agree that "WebSocket" is the correct capitalization for entity names in Go then please also rename `gatewayWebsocket` and `sourcegraphWebsocket` accordingly. (I'd do it to fix my mistake but I don't want to mess or conflict with your PR.)

vdavid (approved):
Approved to unblock and because it's generally 👌. Also left a few suggestions to improve it further before merging.

cmd/src/gateway_benchmark.go
vdavid: I've changed it to "WebSocket" everywhere in https://github.com/sourcegraph/sourcegraph/pull/1526

cmd/src/gateway_benchmark.go
```diff
@@ -152,7 +178,30 @@ func benchmarkEndpoint(client *http.Client, url string) time.Duration {
...
+	if err != nil {
+		fmt.Printf("WebSocket write error: %v\n", err)
+		return 0
+	}
+	_, message, err := conn.ReadMessage()
+	if err != nil {
+		fmt.Printf("WebSocket read error: %v\n", err)
+		return 0
+	}
+	if string(message) != "pong" {
```
vdavid, line 201: I've changed my other PR https://github.com/sourcegraph/sourcegraph/pull/1526 to respond with "pong" on all three endpoints.

cmd/src/gateway_benchmark.go
slimsag: wikipedia says WebSocket so yea let's go with that capitalization

cmd/src/gateway_benchmark.go
slimsag: Fantastic!

cmd/src/gateway_benchmark.go
vdavid: I was aware of that, I just wasn't aware how to translate MixedCase entity names into Go entity names. But yeah, all clear, WebSocket it is!

go.sum
slimsag: Yeah, so I think we're going to run into some challenges here with getting this PR merged:

* Currently, I can't develop src-cli at all without modifying the go.mod file to have a replace directive pointing to the private monorepo: https://sourcegraph.slack.com/archives/C07KZF47K/p1731627487087909
* With the replace directive, the Go version of this repository must be updated.. which requires updating the Go packages.
* I think(?) this repository now uses some Bazel shenanigans and that will need to be udpated to reflect these two bullet points - or else I'm guessing this PR will fail CI.

In short, I don't think we can currently merge changes to src-cli repository at all :/ will need to sort this out seperately before we can merge this PR.

vdavid: Crap, what an unexpected skeleton falling out of that closet. Thanks for handling it!

camdencheek (approved):
Didn't review the changes to `src gateway benchmark` closely, but stamping on behalf of the code search team for the version bumps. Thanks!

PR #1131, opened 2024-12-06
gateway: benchmark: fix 'all websocket requests fail' case

slimsag:
The first commit here replaces the authentication with a simpler / more correct implementation (no transport) and improves usage doc commands slightly.

The second commit fixes a major issue: we would connect the websocket at the very start, which meant that if the tests which run _before_ we use the websocket take >60s.. by the time we use the websocket, the server would already consider the connection dead due to inactivity. This fixes the issue by only connecting the first time we want to use the connection, and also adding reconnect logic.

### Test plan

Manually tested

### DIFF
--- a/cmd/src/gateway_benchmark.go
+++ b/cmd/src/gateway_benchmark.go
@@ -62,21 +62,15 @@ Examples:
 		}
 
 		var (
-			gatewayWebsocket, sourcegraphWebsocket *websocket.Conn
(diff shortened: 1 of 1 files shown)


vdavid (approved):
Both changes make sense. I'm AFK but went through all the code on mobile, and they look good. Approving to unblock.

PR #1137, opened 2024-12-16
Use 50, 200, 500 max token counts (PRIME-654)

vdavid:
This PR aims to investigate https://linear.app/sourcegraph/issue/PRIME-654/investigate-why-the-fastpathnon-fastpath-diff-decreases-with-longer, and answer the question of whether the back end seems to contribute to our observations there (increased latency diff between SG→CG and direct CG with _shorter_ completions).

It uses the somewhat arbitrary token counts 50, 200, and 500 to simulate single-line, multiline, and long code completions. In all cases, it uses the messages

```
{"role": "user", "content": "def bubble_sort(arr):"},
{"role": "assistant", "content": "Here is a bubble sort:"}
```

talking to Claude 3 Haiku (for cost efficiency) at temperature 0.0 and streaming ON.

To fully understand the changes in this PR, see also https://github.com/sourcegraph/src-cli/pull/1138 which this PR builds on.

### Test plan

(I don't intend to merge this.)

Caveat: I was not extremely rigorous in looking into the exact differences between the requests assembled by SG and by the `src-cli` script. I trusted that we successfully converted the same arguments passed to SG to the format CG needs them. Whitespace in the body and some HTTP headers are probably different, though. This can be tested and improved if needed.

### DIFF
--- a/cmd/src/gateway_benchmark_stream.go
+++ b/cmd/src/gateway_benchmark_stream.go
@@ -66,18 +66,20 @@ Examples:
 		fmt.Printf("Starting benchmark with %d requests per endpoint...\n", *requestCount)
 		if *gatewayEndpoint != "" {
 			fmt.Println("Benchmarking Cody Gateway instance:", *gatewayEndpoint)
-			endpoint := buildGatewayHttpEndpoint(*gatewayEndpoint, *sgdToken, 200)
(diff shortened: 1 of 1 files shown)


vdavid:
Closing this in favor of embedded functionality in https://github.com/sourcegraph/src-cli/pull/1138. It needs three separate CLI runs for the three token counts, and it'll work.

PR #1143, opened 2025-01-28
feature/security: Implement image signature verification

willdollman:
Implement verification of container signatures for release images. This allows customers to easily verify the signatures we started publishing in [SEC-2445](https://linear.app/sourcegraph/issue/SEC-2445/sign-public-releases).

The implementation shares several helper functions with the `src sbom fetch` command, as both fetch information about releases, talk to our container registries, and use `cosign` behind the scenes.

The verification process is:
- Fetch list of images in $release from GCS bucket
- Fetch our public key from GCS bucket
- Verify signature of each image at the release tag using `cosign verify` with our public key
- Output image digests once they're verified

Digests can then be used or checked against our deployment configs by customers. Customers could also use the sigstore admission controller on their cluster to automate this, but allowing signature verification via `src-cli` is convenient and a small code change.

Signature verification of a subset of images:

CleanShot 2025-01-28 at 20 40 47

### Test plan

- Tested signature verification locally
- Verified SBOM fetching still works after changes to helper funcs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ All notable changes to `src-cli` are documented in this file.
 
 ## Unreleased
 
+## 6.0.0
+
+- Container signature verification support: Container signatures can now be verified for Sourcegraph releases after 5.11.4013 using `src signature verify -v <release>` [#1143](https://github.com/sourcegraph/src-cli/pull/1143)
+
 ## 5.11.1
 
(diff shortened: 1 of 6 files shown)


willdollman:
golangci-lint setup seems to be failing in CI, but I've run locally and no issues flagged in these changes.

camdencheek (approved):
Looks good to me!

PR #1146, opened 2025-02-20
Support pre-compressed uploads in code-intel upload, deprecate LSIF 

antonsviridov-src:
Closes GRAPH-1046

Contains a lot of code vendored in from the public snapshot – the main upload methods were unfortunately private so I had to unravel a bunch of dependencies.

### Test plan

- The CLI is now built from source and used to upload the scip-go index for this repository – both in compressed and raw form


varungandhi-src:
> As an alternative proposal, I can modify the scip-go Github Action we have to build this CLI from source and use it for upload. WDYT @varungandhi-src?

You mean the one in this repo itself? Yeah, sure, that seems like a good idea. 👍🏽

.github/workflows/scip.yml
```diff
@@ -11,11 +11,14 @@ jobs:
...
-        run: curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o /usr/local/bin/src;
-          chmod +x /usr/local/bin/src
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: 1.23.2
       - name: Set directory to safe for git
         run: git config --global --add safe.directory $GITHUB_WORKSPACE
+      - name: Build src-cli
+        run: go build -o /usr/local/bin/src ./cmd/src
```
varungandhi-src, line 21: Looks like you're missing a compression step in the CI workflow? Instead of writing to a global directory and assuming that it'll be on path, can we invoke the binary below as `./src-cli`?

antonsviridov-src: Sorry, I intentionally kept it uncompressed as per https://github.com/sourcegraph/src-cli/pull/1146#issuecomment-2678470749 – my main concern with this PR is not breaking the most commonly present workflow used by customers.

I guess I can do uploads both compressed and uncompressed.

cmd/src/code_intel_upload_transfer.go
```diff
@@ -0,0 +1,704 @@
...
+			float64(originalSize)/1000/1000,
+			float64(compressedSize)/1000/1000,
+		))
+	}
+
+	if compressedSize <= opts.MaxPayloadSizeBytes {
+		return uploadIndex(ctx, httpClient, opts, compressedReader, compressedSize, &originalSize)
+	}
+
+	return uploadMultipartIndex(ctx, httpClient, opts, compressedReader, compressedSize, &originalSize)
```
varungandhi-src, line 74: Let's call `UploadCompressedIndex` here since this is conceptually the same? Or do you think the two code paths may need to diverge later?

cmd/src/code_intel_upload_transfer.go
```diff
@@ -0,0 +1,704 @@
...
+
+/*
+	NOTE:
+
+	All the definitions below have been vendored in from the Sourcegraph repository public snapshot at
+	this commit: https://github.com/sourcegraph/sourcegraph-public-snapshot/commit/1af563b61442c255af7b07a526efd71b3b0bad0d
+
+	All necessary definitions were vendored in without changes until the code built successfully.
+*/
+
```
varungandhi-src, line 102: Let's move this + below code to a separate file with a `_vendored.go` suffix for clarity? That way, in case we need to vendor something else in the future, there is a clear pattern.

It would be good to create a small follow-up PR in the monorepo mentioning this vendored file in a doc comment for the `UploadHandler` type (or maybe some more appropriate place if you have a suggestion).

varungandhi-src (approved):
I have not reviewed the vendored code, but let's make sure we add some comments here and in the monorepo to make it clear which code is coupled to which other code (e.g. using `NOTE(id: blah)` which can be used for searching)

Mostly non-blocking/stylistic comments. Please @ mention me if you want me to take a deeper/second look at something specific.

cmd/src/code_intel_upload_transfer.go
antonsviridov-src: Let me check – there are some _minor_ modifications for me to use uncompressed size as a pointer to indicate optionality – I can remove that and rely on zero value instead (which is exactly what's currently happening), which would mean that the vendored code is 100% carbon copy of the code on that commit.

.github/workflows/scip.yml
```diff
@@ -11,11 +11,14 @@ jobs:
     container: sourcegraph/scip-go
     steps:
       - uses: actions/checkout@v4
-      - name: Get src-cli
-        run: curl -L https://sourcegraph.com/.api/src-cli/src_linux_amd64 -o /usr/local/bin/src;
-          chmod +x /usr/local/bin/src
+      - name: Set up Go
```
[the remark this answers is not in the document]
antonsviridov-src: 30c9f716b9be6d4d3465a4c5ded85eaf4c3f9559

cmd/src/code_intel_upload_flags.go
```diff
@@ -307,3 +367,12 @@ func validateCodeIntelUploadFlags() error {
...
 }
+
+func checkIsGzipped(r io.Reader) error {
+	gz, err := gzip.NewReader(r)
+	if err != nil {
+		return err
+	}
+	defer gz.Close()
+	return nil
+}
```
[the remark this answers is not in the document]
antonsviridov-src: 30c9f716b9be6d4d3465a4c5ded85eaf4c3f9559

PR #1147, opened 2025-02-21
Remove LSIF support from src-cli

antonsviridov-src:
Also uses streaming parser to infer metdata about index, to avoid loading the whole thing in memory

### Test plan

- Added a test for indexer name/version inference
- Removed all LSIF tests 
- There are no other tests 😅


bahrmichael (approved):
LGTM

cmd/src/code_intel_upload_flags.go
```diff
@@ -122,56 +109,52 @@ func parseAndValidateCodeIntelUploadFlags(args []string) (*output.Output, bool,
...
 		}
 
-		return nil, false, errors.Newf("file %q does not exist", codeintelUploadFlags.file)
+		return nil, errors.Newf("file %q does not exist", codeintelUploadFlags.file)
 	}
 
 	isSCIPAvailable, err := isSCIPAvailable()
 	if err != nil {
-		return nil, false, err
+		return nil, err
```
varungandhi-src, line 134: Let's remove the `isSCIPAvailable()` check here? I don't think we particularly care about what error message someone gets on having a large divergence between src-cli and Sourcegraph.

We also have clear docs on how to get a compatible src-cli version. https://github.com/sourcegraph/src-cli?tab=readme-ov-file#version-compatible-with-your-sourcegraph-instance

The logic for the `isSCIPAvailable` function is a bit out-dated with this PR anyways (it still references `lsif`).

cmd/src/code_intel_upload_flags.go
```diff
@@ -484,14 +301,31 @@ func inferIndexRoot() (string, error) {
...
+	// convert file to io.Reader
+	if err := visitor.ParseStreaming(file); err != nil {
+		return "", "", err
+	}
+
+	if metadata == nil {
+		return "", "", errors.New("index file does not contain metadata")
+	}
+
+	return metadata.ToolInfo.Name, metadata.ToolInfo.Version, nil
```
varungandhi-src, line 328: You need an extra `nil` check for `ToolInfo` since the field has type `*scip.ToolInfo`, so `metadata.ToolInfo` could panic if that's not set.

cmd/src/code_intel_upload_flags.go
```diff
@@ -320,28 +208,12 @@ func inferDefaultFile() (string, error) {
...
-		return "", errors.Newf("both %s and %s exists - cannot determine unambiguous choice", scipFilename, lsifFilename)
-	}
-	if !(hasSCIP || hasLSIF) {
-		return "", formatInferenceError(argumentInferenceError{"file", errors.Newf("neither %s nor %s exists", scipFilename, lsifFilename)})
-	}
-
-	if hasSCIP {
+	if !hasSCIP {
+		return "", formatInferenceError(argumentInferenceError{"file", errors.Newf("%s does not exist", scipFilename)})
+	} else {
```
varungandhi-src, line 213: I think you could simplify this function a bit to avoid the `hasSCIP` boolean.

```
const scipFilename = "index.scip"
_, err := os.Stat(scipFilename)
if err == nil {
    return scipFilename, nil
}
if os.IsNotExist(err) {
    return "", <construct special error here>
}
return "", err
```

varungandhi-src (approved):

PR #1151, opened 2025-03-05
Bike-shedding on search jobs

peterguy:
What started as a simple linter placating ended up with adding support for numeric search job ids instead of needing to use the base64-encoded values.

Not sure how helpful that is.

### Test plan

TBD


trly (approved):
Using the non-encoded ids is a great change. I'll be sure to apply this pattern to future changes I make as well

PR #1154, opened 2025-03-19
bahrmichael/fail-fast

bahrmichael:

### DIFF
--- a/internal/batches/executor/executor.go
+++ b/internal/batches/executor/executor.go
@@ -3,10 +3,8 @@ package executor
 import (
 	"context"
 	"fmt"
-	"sync"
+	"github.com/sourcegraph/conc/pool"
 	"time"
-
-	"github.com/neelance/parallel"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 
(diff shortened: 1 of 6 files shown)


internal/batches/executor/executor.go
```diff
@@ -69,83 +68,72 @@ type NewExecutorOpts struct {
...
-			case <-ctx.Done():
-				return
-			default:
-				err := x.do(ctx, task, ui)
-				if err != nil {
-					x.par.Error(err)
-				}
-			}
-		}(task, ui)
+		t := task
```
camdencheek, line 109: Why create a new variable here?

internal/batches/executor/executor.go
```diff
@@ -69,83 +68,72 @@ type NewExecutorOpts struct {
...
 		doneEnqueuing: make(chan struct{}),
-		par:           parallel.NewRun(opts.Parallelism),
+		workPool:      p,
 	}
 }
 
 // Start starts the execution of the given Tasks in goroutines, calling the
 // given taskStatusHandler to update the progress of the tasks.
-func (x *executor) Start(ctx context.Context, tasks []*Task, ui TaskExecutionUI) {
+func (x *executor) Start(tasks []*Task, ui TaskExecutionUI) {
```
camdencheek, line 99: I'd suggest creating the work pool in `Start`, then saving it on the `executor` struct so you can `Wait()` it later. It's a bit of an antipattern to save a context on a struct without really good reason

camdencheek (approved):
Couple small comments, but all the concurrency stuff looks great to me!

internal/batches/executor/executor.go
bahrmichael: I followed the example from the sourcegraph/conc readme, and slightly recalled something about go syntax and requiring variables to be in the right scope. Not too confident if the var duplication in this example is because of the stream.

```
func process(stream chan int) {
    p := pool.New().WithMaxGoroutines(10)
    for elem := range stream {
        elem := elem
        p.Go(func() {
            handle(elem)
        })
    }
    p.Wait()
}
```

internal/batches/executor/executor.go
bahrmichael: Sounds like a good change! Will tackle that before merging.

internal/batches/executor/executor.go
camdencheek: Oh, I should fix that! That's no longer necessary after [go 1.22](https://go.dev/blog/loopvar-preview)

bahrmichael: Nice! Thank you :)

PR #1157, opened 2025-05-07
fix: update documentation URLs to use sourcegraph.com

trly:
This commit updates several documentation URLs throughout the codebase to point to the `sourcegraph.com` domain instead of `docs.sourcegraph.com`. This ensures that users are directed to the most up-to-date and accurate documentation.

docs.sourcegraph.com links that were broken after updating were left as-is

#### Manual testing
Validated methods with changed URLs are still functional as expected. No application logic was changed, only the URL string.


trly:
https://github.com/sourcegraph/docs/pull/1122#issuecomment-2858892447

PR #1159, opened 2025-05-09
Bahrmichael/dropped edits 2

bahrmichael:


bahrmichael:
After cleaning up a bunch of "AI dodging the actual problem" code, we can still reproduce the error, meaning that this approach might not have worked.

```
github.com/mibali/massive-repository: Done!

❌ Error:
   {
     "message": "2 errors occurred:\n\t* Must validate one and only one schema (oneOf)\n\t* commits.0: diff is required",
     "path": [
       "createChangesetSpec"
     ]
   }

💡 The troubleshooting documentation can help to narrow down the cause of the errors:
   https://docs.sourcegraph.com/batch_changes/references/troubleshooting
exit status 1
```

PR #1177, opened 2025-06-24
Add semgrep CI scan step

willdollman:
Implement SAST scanning as part of [SEC-3159(https://linear.app/sourcegraph/issue/SEC-3159)

### Test plan

Check scan succeeds

### DIFF
--- a/.github/workflows/semgrep.yml
+++ b/.github/workflows/semgrep.yml
@@ -1,0 +1,37 @@
+name: Semgrep - SAST Scan
+
+on:
+  pull_request_target:
+    types: [ closed, edited, opened, synchronize, ready_for_review ]
+
+jobs:
+  semgrep:
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+      actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
+    runs-on: ubuntu-latest
+    container:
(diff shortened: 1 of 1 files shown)


.github/workflows/semgrep.yml
```diff
@@ -0,0 +1,37 @@
...
+
+jobs:
+  semgrep:
+    permissions:
+      contents: read # for actions/checkout to fetch code
+      security-events: write # for github/codeql-action/upload-sarif to upload SARIF results
+      actions: read # only required for a private repository by github/codeql-action/upload-sarif to get the Action run status
+    runs-on: ubuntu-latest
+    container:
+      image: returntocorp/semgrep
```
evict, line 15: no need for a tag here?

willdollman: Just a copypasta from our [other semgrep.ymls](https://sourcegraph.sourcegraph.com/search?q=context:sourcegraph+returntocorp/semgrep&patternType=keyword&sm=0). Looks like GHA uses :latest by default. We should probably pin to a specific version, but will need to do that everywhere.

evict (approved):

.github/workflows/semgrep.yml
evict: > but will need to do that everywhere

we can start here, hehe.

willdollman: We also need to overhaul all our SAST, i'll leave it with @andreeleuterio 😉

PR #1186, opened 2025-11-01
Chore: Fix snapshot upload filter for incompatible SQL statements

marcleblanc2:
Expanded the filtering mechanism beyond commenting on extensions, to address new features in Postgres v17 which are not available in our SG Cloud pg v16 databases, as well as SQL queries / commands which break Google Cloud SQL's very fragile import. 

Also renamed "primary.sql" to "pgsql.sql" to be consistent, as this is the only place across our entire codebase that this database got yet-another-name. 

### Test plan

Tested on customer Cloud migration

Updated unit tests 

Updated process docs for IE team performing these migrations

### DIFF
--- a/cmd/src/snapshot_upload.go
+++ b/cmd/src/snapshot_upload.go
@@ -152,30 +152,30 @@ func copyDumpToBucket(ctx context.Context, src io.ReadSeeker, stat fs.FileInfo, 
 	// To assert against actual file size
 	var totalWritten int64
 
-	// Do a partial copy that trims out unwanted statements
+	// Do a partial copy, that filters out incompatible statements
 	if trimExtensions {
-		written, err := pgdump.CommentOutInvalidLines(object, src, progressFn)
+		written, err := pgdump.FilterInvalidLines(object, src, progressFn)
 		if err != nil {
-			return errors.Wrap(err, "trim extensions and upload")
+			return errors.Wrap(err, "filter out incompatible statements and upload")
 		}
 		totalWritten += written
 	}
 
-	// io.Copy is the best way to copy from a reader to writer in Go, and storage.Writer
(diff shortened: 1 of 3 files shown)


cmd/src/snapshot_databases.go
```diff
@@ -19,13 +19,13 @@ func init() {
...
 
 USAGE
-	src [-v] snapshot databases [--targets=<docker|k8s|"targets.yaml">] [--run] <pg_dump|docker|kubectl> 
+	src [-v] snapshot databases [--targets=<docker|k8s|"targets.yaml">] [--run] <pg_dump|docker|kubectl>
 
 TARGETS FILES
 	Predefined targets are available based on default Sourcegraph configurations ('docker', 'k8s').
 	Custom targets configuration can be provided in YAML format with '--targets=target.yaml', e.g.
 
-		primary:
```
bobheadxi, line 28: I don't think we should rename this in this change, as it's unrelated to the main thing being fixed, and is also a configuration change

I also don't think this is correct? If anything it should be called `frontend` https://sourcegraph.sourcegraph.com/github.com/sourcegraph/sourcegraph/-/blob/migrations/README.md but either way, I dont think we should change that here

bobheadxi: Hmm, I suppose we call the container `pgsql` in some places... I'm not sure that's what we want to align on though, and either way, separate change 🙏

marcleblanc2: Noted, have removed it from this PR, will submit it in a new PR. From the self-hosted customer's perspective, this database is known as the pgsql database, as it's the container / pod they deal with.

bobheadxi (approved):
thank you!

PR #1187, opened 2025-11-03
Chore: Rename 'primary' to 'pgsql' in the src-cli commands for database uploads

marcleblanc2:
The `pgsql` / `frontend` database already has too much confusion over its name when TS teams work with self-hosted customer admins. As far as I'm aware, the `src snapshot databases` commands for taking and uploading database snapshots as part of the migration from self-hosted to our Cloud is the only place our customers see the same database get a bonus third name `primary`, adding unnecessary confusion to a process which is already a little overwhelming for our customers, specifically the types of customers we're encouraging to migrate from self-hosted to Cloud.

I propose renaming `primary` -> `pgsql` in these src-cli commands, because that's the name of the pod / container which our self-hosted customer admins and TS teams already work with on a day-to-day basis, and it matches the database's name in the SG Cloud infra, and the rest of the commands throughout the migration process:

Customer commands:
```shell
# pgsql / frontend database is the only one that doesn't match from container name to snapshot name
docker exec -i pgsql sh -c 'PGPASSWORD=sg pg_dump ... --username=sg --dbname=sg' > src-snapshot/primary.sql

# codeintel and codeinsights get a snapshot name which match the container names
docker exec -i codeintel-db sh -c 'PGPASSWORD=sg pg_dump ... --username=sg --dbname=sg' > src-snapshot/codeintel.sql
docker exec -i codeinsights-db sh -c 'PGPASSWORD=password pg_dump ... --username=postgres --dbname=postgres' > src-snapshot/codeinsights.sql
```

IE commands: 
```sql
DROP DATABASE IF EXISTS "pgsql";
CREATE DATABASE "pgsql";
```

```shell
gcloud sql import sql $CLOUD_SQL_INSTANCE  \
  gs://$BUCKET/primary.sql  \ # Confusing primary.sql
  --database=pgsql            # Database is named pgsql inside our Cloud instances
```

This database is referred to as `frontend` in database schema code, however, our self-hosted customer admins are never exposed to the database name `frontend`; it is the pgsql container that they work with / are taking the snapshot out of. The database name inside the container is also `sg`, but renaming the src-cli commands to `sg` wouldn't be any clearer.

### Test plan

Naming change

### DIFF
--- a/internal/pgdump/pgdump.go
+++ b/internal/pgdump/pgdump.go
@@ -9,7 +9,7 @@ import (
 
 // Targets represents configuration for each of Sourcegraph's databases.
 type Targets struct {
-	Primary      Target `yaml:"primary"`
+	Pgsql        Target `yaml:"pgsql"`
 	CodeIntel    Target `yaml:"codeintel"`
 	CodeInsights Target `yaml:"codeinsights"`
 }
@@ -59,8 +59,8 @@ type Output struct {
 // path. It can be provided a zero-value Targets to just generate the output paths.
 func Outputs(dir string, targets Targets) []Output {
 	return []Output{{
-		Output: filepath.Join(dir, "primary.sql"),
-		Target: targets.Primary,
+		Output: filepath.Join(dir, "pgsql.sql"),
+		Target: targets.Pgsql,
 	}, {
 		Output: filepath.Join(dir, "codeintel.sql"),
 		Target: targets.CodeIntel,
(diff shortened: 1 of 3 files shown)


internal/pgdump/pgdump.go
```diff
@@ -59,8 +59,8 @@ type Output struct {
 // path. It can be provided a zero-value Targets to just generate the output paths.
 func Outputs(dir string, targets Targets) []Output {
 	return []Output{{
-		Output: filepath.Join(dir, "primary.sql"),
-		Target: targets.Primary,
+		Output: filepath.Join(dir, "pgsql.sql"),
+		Target: targets.Pgsql,
```
bobheadxi, line 63: It's been a long time since I've been here, but are there any other process docs we need to update that may reference this DB by its old name?

PR #1188, opened 2025-11-04
Feature:  Allow re-uploading subsets of files

marcleblanc2:
It happens from time to time that when migrating from self-hosted to Cloud, a customer admin may need to re-upload a database export. These database files can be very large, and may be uploaded from the other side of the planet from the GCS bucket, and some admins may be working from home with slow upload speeds. So, when the IE team needs to ask a customer to re-upload a database file, maybe with a different set of args, or a different format, it can be quite a problem to have to re-upload all 4 files when maybe only 1 is needed. 

- Added the `--file` arg, where customer admins can specify all / any / none of the database dump files, and / or the summary.json file, to upload to the GCS bucket
- Decomposed for readability, maintainability, and testability
- Documented for clarity
- Added tests

### Test plan

Tested on marc-dev cloud migration and bucket

### DIFF
--- a/cmd/src/snapshot_upload.go
+++ b/cmd/src/snapshot_upload.go
@@ -67,11 +67,11 @@ Usage:
 
 Examples:
 
-	src snapshot upload -bucket=sourcegraph-cloud-data-migration-example-bucket -credentials=path/to/migration_private_key.json
+	src snapshot upload -bucket=example-bucket-name -credentials=path/to/migration_private_key.json
 
-	src snapshot upload -bucket=sourcegraph-cloud-data-migration-example-bucket -credentials=./migration_private_key.json -file=pgsql.sql
+	src snapshot upload -bucket=example-bucket-name -credentials=./migration_private_key.json -file=pgsql.sql
 
-	src snapshot upload -bucket=sourcegraph-cloud-data-migration-example-bucket -credentials=./migration_private_key.json -file="codeinsights.sql, codeintel.sql, pgsql.sql"
+	src snapshot upload -bucket=example-bucket-name -credentials=./migration_private_key.json -file="codeinsights.sql, codeintel.sql, pgsql.sql"
 
 Args:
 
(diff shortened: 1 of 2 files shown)


cmd/src/snapshot_upload.go
```diff
@@ -18,168 +19,317 @@ import (
...
+	filterSQL bool // Whether to filter incompatible SQL statements during upload, true for database files, false for summary file
+}
 
 func init() {
-	usage := `'src snapshot upload' uploads instance snapshot contents generated by 'src snapshot databases' and 'src snapshot summary' to the designated bucket.
+	usage := fmt.Sprintf(`'src snapshot upload' uploads instance snapshot contents generated by 'src snapshot databases' and 'src snapshot summary' to the designated bucket.
 
 USAGE
-	src snapshot upload -bucket=$BUCKET -credentials=$CREDENTIALS_FILE
+	src snapshot upload -bucket=$MIGRATION_BUCKET_NAME -credentials=./migration_private_key.json [-file=$FILE]
```
bobheadxi: I think I like the credentials file better when it was `$CREDENTIALS_FILE`, lets you decide where to put it instead of current working dir where it may be mishandled 👀

cmd/src/snapshot_upload.go
```diff
@@ -18,168 +19,317 @@ import (
...
+	Optional: Specify files to upload as a comma-delimited list (with file extensions). Valid values: %s. Default: All files.
+	Examples: -file=summary.json,pgsql.sql or -file=codeintel.sql
+`, strings.Join(listOfValidFiles, ", "))
+
 	flagSet := flag.NewFlagSet("upload", flag.ExitOnError)
 	bucketName := flagSet.String("bucket", "", "destination Cloud Storage bucket name")
 	credentialsPath := flagSet.String("credentials", "", "JSON credentials file for Google Cloud service account")
-	trimExtensions := flagSet.Bool("trim-extensions", true, "trim EXTENSION statements from database dumps for import to Google Cloud SQL")
+	fileArg := flagSet.String("file", strings.Join(listOfValidFiles, ","), "comma-delimited list of files to upload")
+	filterSQL := flagSet.Bool("filter-sql", true, "filter incompatible SQL statements from database dumps for import to Google Cloud SQL")
```
bobheadxi: I assume you're updating all docs that exist elsewhere to use renamed flags 🙏

bobheadxi (commented):
Thank you! just one minor suggested tweak to the docstring

cmd/src/snapshot_upload.go
marcleblanc2: Searching for the `trim-extensions` flag, it appears to not be documented [anywhere else in our code](https://sourcegraph.sourcegraph.com/search?q=repo:%5Egithub.com/sourcegraph+trim-extensions&patternType=keyword&sm=0), or in Notion. 

The helper text in this repo appears to be the only documentation, sub-command helper text doesn't seem to get printed -- separate issue to resolve.

cmd/src/snapshot_upload.go
marcleblanc2: I agree customers should decide where to put it. I'm not clear on the mishandling part.

The CLI doesn't make use of the env vars, and I felt it useful to clarify in the helper text that the `-bucket` arg is a string, whereas `-credentials` is a file path.

I've rewritten the helper text to improve it from both angles.

cmd/src/snapshot_upload.go
marcleblanc2: The only documentation / reference of this is in [our internal process doc](https://www.notion.so/sourcegraph/IE-Cloud-Migrating-self-hosted-databases-to-Sourcegraph-Cloud-6bff28ede18b455da604217bf2c1c475?source=copy_link#124a8e11265880788360c1db090c18f7), and thus the commands we copy / paste / send to the customer admin, so I'll make the changes to match in this doc once this PR is merged and released.

marcleblanc2:
Related fix, so now the subcommand help / usage text now actually prints: https://github.com/sourcegraph/src-cli/pull/1190

PR #1190, opened 2025-11-09
Fix: Handling of --help arg for sub commands

marcleblanc2:
Usage / helper text was only printed for either `src` (top-level) or `src <command>` (command-level), ignoring any of the usage functions defined by subcommands. 

For example, `src snapshot upload --help` would only output the usageFunc for `src snapshot`

With this change:

Any of the following will output the usageFunc for `src snapshot upload`

```shell
src snapshot upload --help
src snapshot --help upload
src --help snapshot upload

src snapshot upload -h
src snapshot -h upload
src -h snapshot upload

src snapshot upload help
src snapshot help upload
src help snapshot upload
```

Any of the following will output the usageFunc for `src snapshot`

```shell
src snapshot --help
src --help snapshot

src snapshot -h
src -h snapshot

src snapshot help
src help snapshot
```

Any of the following will output the usageFunc for `src`

```shell
src --help

src -h

src help
```

### Test plan

- Manual testing via Bash script
- Added unit tests
- Amp's pretty happy with itself
- The issues remediated in https://github.com/sourcegraph/src-cli/pull/1117 are still remediated; commands that do not require args still run
- The functionality change from https://github.com/sourcegraph/src-cli/pull/1105 is still behaving as needed; helper text is printed to stdout when expected / when the user runs a command with a --help / -h / help arg, but prints to stderr when printed in response to invalid user input

### DIFF
--- a/cmd/src/cmd_test.go
+++ b/cmd/src/cmd_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"bytes"
 	"flag"
+	"fmt"
 	"os"
 	"os/exec"
 	"testing"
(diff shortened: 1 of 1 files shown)


marcleblanc2:
Hey @bobheadxi or @keegancsmith, I'd like this to make it into the 6.10 release on Wednesday, if either of you has a moment to review, or lmk who else I should ask 🙏

cmd/src/cmd_test.go
```diff
@@ -0,0 +1,346 @@
...
+			arg:          "help",
+			contains:     "usage text",
+			expectedExit: 0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			tempDir := t.TempDir()
+			cmd := exec.Command(os.Args[0], "-test.run=^TestCommander_Run_HelpFlag$")
```
keegancsmith, line 183: the pattern in this file of using subprocesses is very complicated for what you want to check. I'm guessing you did it because of the os.Exit in the code under test. Rather just make os.Exit something you can inject?

cmd/src/cmd.go
```diff
@@ -42,86 +42,147 @@ func (c *command) matches(name string) bool {
...
 			panic(fmt.Sprintf("all registered commands should use flag.ExitOnError: error: %s", err))
 		}
 
-		// Execute the subcommand.
-		if err := cmd.handler(flagSet.Args()[1:]); err != nil {
+		// Execute the subcommand
+		// Handle any errors returned
+		if err := cmd.handler(args); err != nil {
+
+			// If the returned error is of type UsageError
```
keegancsmith, line 142: note: the introduction of all these comments does not make the code more readable and makes the PR much harder to review.

marcleblanc2: Noted. I felt the code was rather cryptic to read.

cmd/src/cmd_test.go
marcleblanc2: I'm not aware of the difference, but Amp had picked this route so that it could validate the exit code.

`src snapshot upload -h` would return 0, whereas `src snapshot upload` would return 2 (input error).

Validating the return code is not the most material change, so if it's more complex than it's worth, it's fine to remove.

PR #1192, opened 2025-11-10
go 1.25 + upgrade containers

burmudar:
* Update to go 1.25
* Use latest contianers

### Test plan

`docker build .` + `go build ./cmd/...`


.golangci.yml
burmudar, line 1: this was done by running `golangci-lint migrate`

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/sourcegraph/src-cli
 
-go 1.24
-
-toolchain go1.24.1
+go 1.25
```
keegancsmith: until we start using features that require go 1.25 we should actually make this the minimal possible version (even lower than 1.24). That is because the higher this version makes installing this by downstream users a bit more annoying unless they are on latest go.

but if this makes your life easy, then keep it. I suspect most downstream users consume our published artifacts (except for nixpkgs :P )

Dockerfile
```diff
@@ -4,14 +4,14 @@
 #
 # Nothing fancy here: we copy in the source code and build on the Alpine Go
 # image. Refer to .dockerignore to get a sense of what we're not going to copy.
-FROM golang:1.24.1-alpine@sha256:43c094ad24b6ac0546c62193baeb3e6e49ce14d3250845d166c77c25f64b0386 AS builder
+FROM golang:1.25.2-alpine@sha256:d3f0cf7723f3429e3f9ed846243970b20a2de7bae6a5b66fc5914e228d831bbb AS builder
```
keegancsmith: don't we have a 1.25.4?

burmudar: ah updated the hash but somehow got the tag wrong 🤷🏼 fixing

go.mod
burmudar: I got Amp to check what the lowest version we can support and it's apparently 1.23 so going to use that!

burmudar:
I will fix the lint issues separately

go.mod
```diff
@@ -1,8 +1,6 @@
 module github.com/sourcegraph/src-cli
 
-go 1.24
-
-toolchain go1.24.1
+go 1.23.12
```
keegancsmith, line 3: aah we probably wanted to bump the toolchain. But you know what I am being a bit silly. We should be the only people consuming this so lets just use the latest go version?

PR #1195, opened 2025-11-11
support --help anywhere on command line

keegancsmith:
Previously we special cased --help only for the first level of subcommands. However, this broke for example if you did "src snapshot upload --help". So instead we normalize our command line arguments so that if a user does "--help" we see "-help". We can then rely on go's stdlib flag parser to do the right thing w.r.t. sub commands/etc.

Test Plan: ran with --help in a few different places and observed good behaviour.

Alternative to https://github.com/sourcegraph/src-cli/pull/1190

### DIFF
--- a/cmd/src/cmd.go
+++ b/cmd/src/cmd.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"log"
 	"os"
-	"slices"
 
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
 )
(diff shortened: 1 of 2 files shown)


cmd/src/cmd.go
```diff
@@ -86,16 +85,6 @@ func (c commander) run(flagSet *flag.FlagSet, cmdName, usageText string, args []
...
-		// Print help to stdout if requested
-		if slices.IndexFunc(args, func(s string) bool {
-			return s == "--help"
-		}) >= 0 {
-			cmd.flagSet.SetOutput(os.Stdout)
-			flag.CommandLine.SetOutput(os.Stdout)
-			cmd.flagSet.Usage()
-			os.Exit(0)
-		}
-
```
marcleblanc2: Removing this block (from an OSS contributor) results in usage text getting output to stderr instead of stdout, even when the user provided the `-h` / `--help` arg. 

Personally, I'm not too attached to which stream it comes out, but thought I'd mention it.

Removing this block also fixes the issue which I was trying to fix in https://github.com/sourcegraph/src-cli/pull/1190, where `src snapshot upload -h` was printing the usage text for `src snapshot` and not `upload`, so if we're alright with the usage text going to stderr, then removing this block is the simplest solution.

cmd/src/main.go
```diff
@@ -89,7 +90,21 @@ func main() {
...
+	for i, arg := range args {
+		if arg == "--" {
+			break
+		}
+		if arg == "--help" {
+			args[i] = "-help"
+		}
+	}
+	return args
 }
```
marcleblanc2: Much simpler than the changes Amp steered me down <3

marcleblanc2:
Whoops, accidentally merged main into this branch, fixed by reverting the merge commit and rebased your commit on top of main.

PR #1198, opened 2025-11-11
chore(ci): always use go version in .tool-versions

keegancsmith:
setup-go@v6 supports reading in .tool-versions which means we will now always run the same version we run in local dev. This is also a better version to use than go.mod since that is a minimum version.

Test Plan: CI


burmudar (approved):

keegancsmith:
Well it seems this feature wasn't really tested lol. Was committed two weeks ago and it fails for us and check out this issue https://github.com/actions/setup-go/issues/680

PR #1207, opened 2025-11-25
use mise action with cache

burmudar:
Use mise github action together with github cache by specifying `cache: true`

### Test plan

actions should run on this branch


burmudar:
This change is part of the following stack:

- #1206
    - #1207 ◀

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

PR #1211, opened 2025-11-27
feat(mcp): add script to dump mcp tool list

burmudar:
Add a script which curls s2 mcp endpoint and dumps the list-tools json

Part of SRCH-2167

### Test plan
executed script locally and committed the json

### DIFF
--- a/cmd/src/config_list.go
+++ b/cmd/src/config_list.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 
 	"context"
+
 	"github.com/sourcegraph/src-cli/internal/api"
 )
 


internal/mcp/mcp_tools.json
burmudar, line 1: Note that in subsequent PRs this file is moved to `internal/mcp`

eseliger:
Just a random question: should the src-cli surface depend on the MCP specification, or should it just have similar tools such as `src repo file <path>` or so, and leave us the freedom to tailor it more towards the humans _and_ angets use-case, unlike MCP? 

(I genuinely don't know, just a conversation starter!)

PR #1212, opened 2025-11-27
feat(mcp): parse mcp tool json

burmudar:
Define structures to parse the mcp tool json

- define general Schema which can be used for Input and Out schema definitions
- define SchemaValue interface to unitfy types under one umbrella
- specilzied schema types: SchemaObject, SchemaArray, SchemaPrimitve
- special handling for when `items: true`

Part of SRCH-2167

### Test plan
Added testcase and parsed the dumped json

### DIFF
--- a/internal/mcp/mcp_parse.go
+++ b/internal/mcp/mcp_parse.go
@@ -18,14 +18,12 @@ type ToolDef struct {
 	OutputSchema SchemaObject `json:"outputSchema"`
 }
 
-type RawSchema struct {
-	Type                 string                     `json:"type"`
-	Description          string                     `json:"description"`
-	SchemaVersion        string                     `json:"$schema"`
-	Required             []string                   `json:"required,omitempty"`
-	AdditionalProperties bool                       `json:"additionalProperties"`
-	Properties           map[string]json.RawMessage `json:"properties"`
-	Items                json.RawMessage            `json:"items"`
+type rawSchema struct {
+	Type        string                     `json:"type"`
+	Description string                     `json:"description"`
+	Required    []string                   `json:"required,omitempty"`
+	Properties  map[string]json.RawMessage `json:"properties"`
+	Items       json.RawMessage            `json:"items"`
 }
 
 type SchemaValue interface {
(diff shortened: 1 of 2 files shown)


internal/mcp/mcp_parse.go
```diff
@@ -0,0 +1,167 @@
...
+import (
+	_ "embed"
+	"encoding/json"
+	"fmt"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+//go:embed mcp_tools.json
+var _ []byte
```
burmudar, line 12: temporarily ignored

internal/mcp/mcp_parse.go
```diff
@@ -0,0 +1,159 @@
...
+
+type ToolDef struct {
+	Name         string       `json:"name"`
+	Description  string       `json:"description"`
+	InputSchema  SchemaObject `json:"inputSchema"`
+	OutputSchema SchemaObject `json:"outputSchema"`
+}
+
+type RawSchema struct {
```
keegancsmith: minor: you might as well only define the fields on these structs you will actually end up reading.

And my favourite nit: probably not used outside of this package so don't export.

internal/mcp/mcp_parse.go
```diff
@@ -0,0 +1,159 @@
...
+			InputSchema  RawSchema `json:"inputSchema"`
+			OutputSchema RawSchema `json:"outputSchema"`
+		} `json:"tools"`
+	}{}
+
+	if err := json.Unmarshal(data, &defs); err != nil {
+		return nil, err
+	}
+
+	tools := map[string]*ToolDef{}
```
keegancsmith, line 79: minor: is there any chance the tool order matters? I think in the backend we also use a map and just sort by name when serializing for determinism, but I can't remember.

internal/mcp/mcp_parse.go
```diff
@@ -0,0 +1,159 @@
...
+	SchemaVersion        string                     `json:"$schema"`
+	Required             []string                   `json:"required,omitempty"`
+	AdditionalProperties bool                       `json:"additionalProperties"`
+	Properties           map[string]json.RawMessage `json:"properties"`
+	Items                json.RawMessage            `json:"items"`
+}
+
+type SchemaValue interface {
+	ValueType() string
+}
```
keegancsmith, line 31: can't we remove SchemaValue and just use a normal go interface? Then we just have some sort of standard set of go types we could marshal into (same as how the json library works in go)

internal/mcp/mcp_parse_test.go
```diff
@@ -0,0 +1,100 @@
+package mcp
+
+import (
+	"testing"
+)
+
+func TestLoadToolDefinitions(t *testing.T) {
```
keegancsmith, line 7: a nicer test here would probably just load in mcp_tools.json and then assert on some stable tool's properties. eg the read file tool likely won't change.

internal/mcp/mcp_parse.go
burmudar: yeah, I had grander plans but in the end didn't need this. Removing it

PR #1215, opened 2025-11-27
feat(mcp): build flagset from mcp input schema to parse args

burmudar:
We build a flagset from the input schema. This allows the input schema to be the
source of truth of the flags on the mcp command and ensures we are up to date with
the schema stored in the repository.

Part of SRCH-2167

### Test plan
Added a test case and also executed locally with `commit-search`

### DIFF
--- a/cmd/src/mcp.go
+++ b/cmd/src/mcp.go
@@ -1,31 +1,22 @@
 package main
 
 import (
-	"context"
-	"encoding/json"
 	"flag"
 	"fmt"
-	"os"
-	"strings"
 
-	"github.com/sourcegraph/src-cli/internal/api"
 	"github.com/sourcegraph/src-cli/internal/mcp"
-
-	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
 func init() {
-	if os.Getenv("SRC_EXPERIMENT_MCP") == "true" {
-		flagSet := flag.NewFlagSet("mcp", flag.ExitOnError)
-		commands = append(commands, &command{
-			flagSet: flagSet,
-			handler: mcpMain,
-		})
-	}
+	flagSet := flag.NewFlagSet("mcp", flag.ExitOnError)
+	commands = append(commands, &command{
+		flagSet: flagSet,
+		handler: mcpMain,
+	})
 }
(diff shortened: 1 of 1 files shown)


internal/mcp/mcp_args.go
```diff
@@ -0,0 +1,76 @@
...
+				flagVars[name] = dst
+			default:
+				return nil, nil, fmt.Errorf("unknown schema primitive kind %q", pv.Type)
+
+			}
+		case *SchemaArray:
+			strSlice := new(strSliceFlag)
+			fs.Var(strSlice, name, pv.Description)
+			flagVars[name] = strSlice
+		case *SchemaObject:
```
keegancsmith, line 86: do we have nay uses of that in our tools?

internal/mcp/mcp_args.go
```diff
@@ -0,0 +1,76 @@
...
+			// TODO(burmudar): we can support SchemaObject as part of stdin echo '{ stuff }' | sg mcp commit-search
```
keegancsmith, line 87: why not just pass in json as a string? that way you can have more than one arg.

cmd/src/mcp.go
```diff
@@ -47,6 +48,33 @@ func mcpMain(args []string) error {
...
+	if len(args) < len(inputSchema.Required) {
+		return fmt.Errorf("not enough arguments provided - the following flags are required:\n%s", strings.Join(inputSchema.Required, "\n"))
+	}
+
+	mcp.DerefFlagValues(vars)
+
+	fmt.Println("Flags")
+	for name, val := range vars {
+		fmt.Printf("--%s=%v\n", name, val)
+	}
```
keegancsmith: debug output?

burmudar: doh yeah. This should definitely not be there

internal/mcp/mcp_args.go
burmudar: Yip - that is part of the plan in my head!

cmd/src/mcp.go
burmudar: Removed

PR #1219, opened 2025-12-01
fix: upgrade cross-spawn to 7.0.6 (CVE-2024-21538)

burmudar:
ran `trivy fs .`

[Amp thread](https://ampcode.com/threads/T-e8c505f6-e23d-45ad-bfba-ba18d84c117e)
### Test plan
CI

### DIFF
--- a/npm-distribution/package.json
+++ b/npm-distribution/package.json
@@ -19,6 +19,9 @@
   "dependencies": {
     "tar": "^7.1.0"
   },
+  "overrides": {
+    "cross-spawn": "^7.0.5"
+  },
   "devDependencies": {
     "@types/tar": "6.1.13"
   },


npm-distribution/yarn.lock
```diff
@@ -253,28 +253,28 @@ string-width@^5.0.1, string-width@^5.1.2:
 
 "strip-ansi-cjs@npm:strip-ansi@^6.0.1":
   version "6.0.1"
-  resolved "https://registry.yarnpkg.com/strip-ansi/-/strip-ansi-6.0.1.tgz#9e26c63d30f53443e9489495b2105d37b67a85d9"
```
keegancsmith: should you be using yarn instead of npm here? Or is this just a change in the yarn tool?

burmudar: I think you're right, even though this works still. Looking closer I don't think we even need yarn so going to switch this to use npm

PR #1222, opened 2025-12-01
security: use traversal safe os.Root for FS operations

evict:
This implement a path traversal safe file system to be used with `serve-git`
### Test plan
Tested cloning repositories with `serve-git`, works fine. Doesn't download symlinks.

### DIFF
--- a/cmd/src/main.go
+++ b/cmd/src/main.go
@@ -25,23 +25,22 @@ Usage:
 	src [options] command [command options]
 
 Environment variables
-	SRC_ACCESS_TOKEN        Sourcegraph access token
-	SRC_ENDPOINT            Endpoint to use, if unset will default to "https://sourcegraph.com"
-	SRC_PROXY               A proxy to use for proxying requests to the Sourcegraph endpoint.
-	                        Supports HTTP(S), SOCKS5/5h, and UNIX Domain Socket proxies.
-					        If a UNIX Domain Socket, the path can be either an absolute path,
-					        or can start with ~/ or %USERPROFILE%\ for a path in the user's home directory.
-					        Examples:
-					         - https://localhost:3080
-					         - https://<user>:<password>localhost:8080
-					         - socks5h://localhost:1080
-					         - socks5://<username>:<password>@localhost:1080
-					         - unix://~/src-proxy.sock
-					         - unix://%USERPROFILE%\src-proxy.sock
-					         - ~/src-proxy.sock
-					         - %USERPROFILE%\src-proxy.sock
-					         - C:\some\path\src-proxy.sock
-    SRC_MAX_REQUESTSIZE_MB   Maximum request body size in MB for git operations (default: 10)
+	SRC_ACCESS_TOKEN  Sourcegraph access token
+	SRC_ENDPOINT      endpoint to use, if unset will default to "https://sourcegraph.com"
+	SRC_PROXY         A proxy to use for proxying requests to the Sourcegraph endpoint.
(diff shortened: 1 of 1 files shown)


internal/servegit/gitservice.go
```diff
@@ -16,6 +17,17 @@ import (
...
+
+func maxRequestSize() int64 {
+	if v := os.Getenv("SRC_MAX_REQUESTSIZE_MB"); v != "" {
+		if mb, err := strconv.ParseInt(v, 10, 64); err == nil && mb > 0 {
+			return mb * 1024 * 1024
+		}
+	}
+	return defaultMaxFileSizeMB * 1024 * 1024
+}
+
```
evict: We can leave this out. It's a best practice but for `serve-git` not a big deal.

keegancsmith: yeah the reason I didn't review this faster was this change is trickier to make sure it doesn't break on large repos/git requests/etc. Can you take it out and maybe submit it seperately.

I agree this isn't actually important since this should not be exposed to the internet / bad actors.

internal/servegit/gitservice.go
```diff
@@ -93,15 +109,18 @@ func (s *Handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if _, err = os.Stat(dir); os.IsNotExist(err) {
+	// os.Root only accepts relative paths from it's root. So we trim the
+	// prefix.
+	relDir := strings.TrimPrefix(dir, s.RootFS.Name()+string(os.PathSeparator))
```
keegancsmith: use filepath.Rel. Less tricky code needed then

internal/servegit/serve_test.go
```diff
@@ -132,6 +140,12 @@ func gitInit(t *testing.T, path string) {
 
 func gitInitRepos(t *testing.T, names ...string) string {
 	root := t.TempDir()
+
+	// We cannot os.OpenRoot on a non-existent dir so we return tmpdir
```
keegancsmith, line 144: could this be an issue? previously if the dir didn't exist this program would just return nothing if I am not mistaken. Is it possible it fails at startup now. Likely a fine, but a change in behaviour.

evict: Good point. On main when you open `serve-git` with a non-existing directory, it still works as `repos` will then list cwd. I think it's odd that this works this way as the argument is technically invalid, but I can make sure this behavior remains the same with the new rootFS. 

In this branch it will fail with a non-existent path error.

evict: Hm, actually I think that's a bug. All else is doesn't work when one does that. 

```
$ src serve-git ./foo
serve-git: 2025/12/04 15:00:35 listening on http://[::]:3434
serve-git: 2025/12/04 15:00:35 serving git repositories from ./foo
serve-git: 2025/12/04 15:01:08 WARN: ignoring error searching : lstat foo: no such file or directory
```

When browsing, it indicates nothing is found for everything (my version of `src` was from last year, so after updating previous screenshot didn't apply).

I think the behavior on this branch is better:

```
4/12/25 15:02:50 > go run ./cmd/src serve-git ./foo                                                                                                                                                 (vr/servegit/use-traversal-safe-api) 
open ./foo: no such file or directory
exit status 1
```
Happy to address this post merge if you disagree.

keegancsmith: ouch falling back to cwd is bad. cool

PR #1223, opened 2025-12-03
feat(oauth): use oauth device flow to authenticate with predefined src-cli OAuth client 

burmudar:
This adds the flag `--oauth` to `login` command which then starts the OAuth device authentication flow. `gh` does the same flow when you authenticate from the cli with `gh auth login`.

- add package `internal/oauth`
- discover oauth configuration via path `.well-known/openid-configuration`
- add client to handle flow, in particular to discover the endpoints it should use and ultimately poll for the token once the user as authorized the application.

In https://github.com/sourcegraph/src-cli/pull/1228 we add a keyring to store the oauth credentials.

> [!IMPORTANT]
> A Follow up will remove the `--oauth` flag and make it dynamically decide to start the flow or not based on whether the access token is set.

### Test plan

- Unit tests
```
go test -v ./internal/oauthdevice/...
=== RUN   TestDiscover_Success
--- PASS: TestDiscover_Success (0.00s)
=== RUN   TestDiscover_Caching
--- PASS: TestDiscover_Caching (0.00s)
=== RUN   TestDiscover_Error
--- PASS: TestDiscover_Error (0.00s)
=== RUN   TestStart_Success
--- PASS: TestStart_Success (0.00s)
=== RUN   TestStart_WithScopes
--- PASS: TestStart_WithScopes (0.00s)
=== RUN   TestStart_Error
--- PASS: TestStart_Error (0.00s)
=== RUN   TestStart_NoDeviceEndpoint
--- PASS: TestStart_NoDeviceEndpoint (0.00s)
=== RUN   TestPoll_Success
--- PASS: TestPoll_Success (0.00s)
=== RUN   TestPoll_AuthorizationPending
--- PASS: TestPoll_AuthorizationPending (0.00s)
=== RUN   TestPoll_SlowDown
--- PASS: TestPoll_SlowDown (0.00s)
=== RUN   TestPoll_ExpiredToken
--- PASS: TestPoll_ExpiredToken (0.00s)
=== RUN   TestPoll_AccessDenied
--- PASS: TestPoll_AccessDenied (0.00s)
=== RUN   TestPoll_Timeout
--- PASS: TestPoll_Timeout (0.00s)
=== RUN   TestPoll_ContextCancellation
--- PASS: TestPoll_ContextCancellation (0.00s)
PASS
ok      github.com/sourcegraph/src-cli/internal/oauthdevice     0.625s
```
- Tested manually against my local SG

[Amp thread](https://ampcode.com/threads/T-1e7c987a-e92d-4f67-8be5-be08cfe4048a)

### DIFF
--- a/cmd/src/login.go
+++ b/cmd/src/login.go
@@ -112,7 +112,9 @@ func loginCmd(ctx context.Context, p loginParams) error {
    export SRC_ACCESS_TOKEN=(your access token)
 
    To verify that it's working, run the login command again.
-`, endpointArg, endpointArg)
+
+   Alternatively, you can try logging in using OAuth by running: src login --oauth %s
+`, endpointArg, endpointArg, endpointArg)
 
 	if cfg.ConfigFilePath != "" {
 		fmt.Fprintln(out)
@@ -121,19 +123,7 @@ func loginCmd(ctx context.Context, p loginParams) error {
 
 	noToken := cfg.AccessToken == ""
 	endpointConflict := endpointArg != cfg.Endpoint
-
-	if p.useOAuth {
-		token, err := runOAuthDeviceFlow(ctx, endpointArg, out, p.deviceFlowClient)
-		if err != nil {
-			printProblem(fmt.Sprintf("OAuth Device flow authentication failed: %s", err))
-			fmt.Fprintln(out, createAccessTokenMessage)
-			return cmderrors.ExitCode1
-		}
-
-		cfg.AccessToken = token
-		cfg.Endpoint = endpointArg
-		client = cfg.apiClient(p.apiFlags, out)
-	} else if noToken || endpointConflict {
+	if !p.useOAuth && (noToken || endpointConflict) {
 		fmt.Fprintln(out)
 		switch {
 		case noToken:
@@ -143,6 +133,30 @@ func loginCmd(ctx context.Context, p loginParams) error {
 		}
 		fmt.Fprintln(out, createAccessTokenMessage)
 		return cmderrors.ExitCode1
+	}
+
+	if p.useOAuth {
+		token, err := runOAuthDeviceFlow(ctx, endpointArg, out, p.deviceFlowClient)
+		if err != nil {
+			printProblem(fmt.Sprintf("OAuth Device flow authentication failed: %s", err))
+			fmt.Fprintln(out, createAccessTokenMessage)
+			return cmderrors.ExitCode1
+		}
+
+		if err := oauth.StoreToken(ctx, token); err != nil {
+			fmt.Fprintln(out)
+			fmt.Fprintf(out, "⚠️  Warning: Failed to store token in keyring store: %q. Continuing with this session only.\n", err)
+		}
+
+		client = api.NewClient(api.ClientOpts{
+			Endpoint:          cfg.Endpoint,
+			AdditionalHeaders: cfg.AdditionalHeaders,
+			Flags:             p.apiFlags,
+			Out:               out,
+			ProxyURL:          cfg.ProxyURL,
+			ProxyPath:         cfg.ProxyPath,
+			OAuthToken:        token,
+		})
 	}
 
 	// See if the user is already authenticated.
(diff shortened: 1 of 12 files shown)


cmd/src/login.go
```diff
@@ -122,6 +162,43 @@ func loginCmd(ctx context.Context, cfg *config, client api.Client, endpointArg s
...
 }
+
+func runDeviceFlow(ctx context.Context, endpoint string, out io.Writer, client oauthdevice.Client) (string, error) {
+	authResp, err := client.Start(ctx, endpoint, nil)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Fprintln(out)
+	fmt.Fprintf(out, "🔐 To authenticate, visit %s and enter the code: %s\n", authResp.VerificationURI, authResp.UserCode)
```
eseliger: no emojis in UI please 😬 since claude, this has an AI vibe slop feeling to it :(

cmd/src/login.go
```diff
@@ -122,6 +162,43 @@ func loginCmd(ctx context.Context, cfg *config, client api.Client, endpointArg s
 	}
 	fmt.Fprintln(out)
 	fmt.Fprintf(out, "✔️  Authenticated as %s on %s\n", result.CurrentUser.Username, endpointArg)
+
+	if p.useDeviceFlow {
+		fmt.Fprintln(out)
+		fmt.Fprintf(out, "To use this access token, set the following environment variables in your terminal:\n\n")
+		fmt.Fprintf(out, "   export SRC_ENDPOINT=%s\n", endpointArg)
+		fmt.Fprintf(out, "   export SRC_ACCESS_TOKEN=%s\n", cfg.AccessToken)
```
eseliger: what you get here is not a SG access token, it's an oauth token and it comes with an access token and refresh token (and expiry) and needs to regularly be refreshed.

I think we need to store the accesstoken/refreshtoken pair in secure storage and add some http Transport that refreshes the credential as needed.

cmd/src/login.go
burmudar: valid :) I'll remove it

burmudar: Removed. We need to do a larger sweep of the other emojis too :|

cmd/src/login.go
burmudar: Will look at using https://github.com/99designs/keyring. We already use it with `sg` to store some secrets. It uses your OS keychain

eseliger: oh nice, cross OS too :)

cmd/src/login.go
```diff
@@ -122,6 +172,43 @@ func loginCmd(ctx context.Context, cfg *config, client api.Client, endpointArg s
...
+	authResp, err := client.Start(ctx, endpoint, nil)
+	if err != nil {
+		return "", err
+	}
+
+	fmt.Fprintln(out)
+	fmt.Fprintf(out, "To authenticate, visit %s and enter the code: %s\n", authResp.VerificationURI, authResp.UserCode)
+	if authResp.VerificationURIComplete != "" {
+		fmt.Fprintln(out)
+		fmt.Fprintf(out, "Alternatively, you can open: %s\n", authResp.VerificationURIComplete)
```
burmudar: Should:
* always try to open the browser
* always print the url

cmd/src/login.go
```diff
@@ -28,6 +30,15 @@ Examples:
   Authenticate to Sourcegraph.com:
 
     $ src login https://sourcegraph.com
+
+  Use OAuth device flow to authenticate:
+
+    $ src login --device-flow https://sourcegraph.com
```
keegancsmith: Do other CLI's require a flag for this? If I remember they are normally interactive right? You could still interactively decide between creating an access token vs oauth flows right?

cmd/src/login.go
```diff
@@ -28,6 +30,15 @@ Examples:
...
+
+
+  Override the default client id used during device flow when authenticating:
+
+    $ src login --device-flow https://sourcegraph.com --client-id sgo_my_own_client_id
```
keegancsmith: doesn't seem worth supporting a custom client-id given you shipped a predefined one. If a user still hasn't upgraded sourcegraph just fallback to the old flow?

cmd/src/login.go
burmudar: No they don't. It just happens. The plan it to make it part of the normal flow if you don't have `SRC_ACCESS_TOKEN` set

cmd/src/login.go
burmudar: agreed. It's already removed just have to update

cmd/src/login.go
eseliger: `src login` is a misnomer currently haha, it should probably be renamed to `src whoami` and then `src login` is the interactive flow always 😬

cmd/src/login.go
burmudar: Fixed

cmd/src/login.go
burmudar: as a follow up I will make it interactive if you don't have a SRC_ACCESS_TOKEN set

PR #1225, opened 2025-12-05
refactor(api): add buildTransport that uses ProxyTransport

burmudar:
The customTransport boolean made it a bit awkward to add new things, especially it you want to just wrap a transport. So here we refactor:

- applyProxy -> NewProxyTransport
- `buildTransport` handles all the different configurations to build a transport

### Test plan

- [x] Added minimal unit test

[Tested by making Amp write a minimal proxy](https://ampcode.com/threads/T-a94a90e6-4e14-4c3f-91ad-8f7fc96da8ec) for unix / http and then running `go run ./cmd/src login`. For the socks5 proxy I ran `ssh -D :5050 media` where media is another machine on my network.

- [x] Integration tests
  - [x] Unix Socket 
  - [x] SOCKS5 Proxy
  - [x] HTTP Proxy


internal/api/api_test.go
```diff
@@ -1,3 +1,40 @@
...
+		}
+	})
+
+	t.Run("socks5 proxy sets Proxy", func(t *testing.T) {
+		transport := buildTransport(ClientOpts{ProxyURL: &url.URL{Scheme: "socks5", Host: "proxy:1080"}}, defaultFlags())
+		if transport.Proxy == nil {
+			t.Error("expected Proxy to be set")
+		}
+	})
+}
```
keegancsmith: tbh this test seems pretty low value?

internal/api/proxy.go
```diff
@@ -11,10 +11,9 @@ import (
 	"net/url"
 )
 
-func applyProxy(transport *http.Transport, proxyURL *url.URL, proxyPath string) (applied bool) {
-	if proxyURL == nil && proxyPath == "" {
-		return false
-	}
+func NewProxyTransport(base *http.Transport, proxyURL *url.URL, proxyPath string) *http.Transport {
```
keegancsmith: nit: don't export.

internal/api/proxy.go
```diff
@@ -11,10 +11,9 @@ import (
...
+	// Clone so that we don't change the original transport
+	transport := base.Clone()
```
keegancsmith: very minor, but we literally just cloned the transport before calling this function.

burmudar: Yeah I felt the same about it tbh

internal/api/api_test.go
burmudar: yeah looking at it now - I agree. I initially added it because it felt weird that there were none

PR #1227, opened 2025-12-08
feat(oauth): Add refresh to oauthdevice.Client

burmudar:
add refresh to oauthdevice.Client

oauthdevice: add RefreshToken field and Refresh method

### DIFF
--- a/internal/oauth/flow.go
+++ b/internal/oauth/flow.go
@@ -3,6 +3,7 @@
 package oauth
 
 import (
+	"cmp"
 	"context"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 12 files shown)


internal/oauth/flow.go
```diff
@@ -307,3 +308,55 @@ func (c *httpClient) pollOnce(ctx context.Context, tokenEndpoint, deviceCode str
 
 	return &tokenResp, nil
 }
+
+// Refresh exchanges a refresh token for a new access token.
+func (c *httpClient) Refresh(ctx context.Context, endpoint, refreshToken string) (*TokenResponse, error) {
```
keegancsmith: does this just magically get used to refresh in our client?

burmudar: Not yet! There is some code I still need to land which adds a transport which will do automatic refresh _if you use api.Client_

burmudar: Fixed and addressed in https://github.com/sourcegraph/src-cli/pull/1228

PR #1228, opened 2025-12-08
feat(oauth): Use keyring to store oauth token

burmudar:
This adds a keyring which is used to store OAuth credentials. The OAuth credentials are retrieved from the keyring whenever there is no `SRC_ACCESS_TOKEN` defined.

When we use an OAuth token, we also add a OAuth transport which will automatically refresh the access token using the refresh token when it is near to expiring.

Credentials are stored in the keyring with under the service "Sourcegraph CLI" and with key "src:oauth:<endpoint>". The reason for the endpoint being in the key name is that some clients might access multiple sourcegraph instances like dev and prod, and we can't reuse the same credentials for the different servers.

###  Test plan
* Authorize by running `src login --oauth https://sourcegraph.sourcegraph.com`
* run `export SRC_ENDPOINT=https://sourcegraph.sourcegraph.com; echo 'query { currentUser { username } }' | go run ./cmd/src api`
* Revoke access of `Sourcegraph CLI` application and rerun previous command
```
The OAuth token is invalid. Please check that the Sourcegraph CLI client is still authorized.

To re-authorize, run: src login

Learn more at https://github.com/sourcegraph/src-cli#readme

error: 401 Unauthorized

Invalid OAuth token.
exit status 1
```

### Platforms tested

- [x] Mac
- [ ] Windows
- [x] Linux

### DIFF
--- a/internal/oauth/http_transport.go
+++ b/internal/oauth/http_transport.go
@@ -3,6 +3,7 @@ package oauth
 import (
 	"context"
 	"net/http"
+	"sync"
 	"time"
 )
 
@@ -13,6 +14,8 @@ var _ http.RoundTripper = (*Transport)(nil)
 type Transport struct {
 	Base  http.RoundTripper
 	Token *Token
+
+	mu sync.Mutex
 }
 
 // storeRefreshedTokenFn is the function the transport should use to persist the token - mainly used during
@@ -22,15 +25,9 @@ var storeRefreshedTokenFn = StoreToken
 // RoundTrip implements http.RoundTripper.
 func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	ctx := req.Context()
-	prevToken := t.Token
-	token, err := maybeRefresh(ctx, t.Token)
-	if err != nil {
+
+	if err := t.refreshToken(ctx); err != nil {
 		return nil, err
-	}
-	t.Token = token
-	if token != prevToken {
-		// try to save the token if we fail let the request continue with in memory token
-		_ = storeRefreshedTokenFn(ctx, token)
 	}
 
 	req2 := req.Clone(req.Context())
@@ -42,8 +39,32 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return http.DefaultTransport.RoundTrip(req2)
 }
 
+// refreshToken checks if the token has expired or expiring soon and refreshes it. Once the token is
+// refreshed, the in-memory token is updated and a best effort is made to store the token.
+// If storing the token fails, no error is returned.
+func (t *Transport) refreshToken(ctx context.Context) error {
+	t.mu.Lock()
+	defer t.mu.Unlock()
+
+	prevToken := t.Token
+	token, err := maybeRefresh(ctx, t.Token)
+	if err != nil {
+		return err
+	}
+	t.Token = token
+	if token != prevToken {
+		// try to save the token if we fail let the request continue with in memory token
+		_ = storeRefreshedTokenFn(ctx, token)
+	}
+
+	return nil
+}
+
+// maybeRefresh conditionally refreshes the token. If the token has expired or is expriing in the next 30s
+// it will be refreshed and the updated token will be returned. Otherwise, no refresh occurs and the original
+// token is returned.
 func maybeRefresh(ctx context.Context, token *Token) (*Token, error) {
-	// token has NOT expired or NOT about to expire in 30s
+	// token has NOT expired and is NOT about to expire in 30s
 	if !(token.HasExpired() || token.ExpiringIn(time.Duration(30)*time.Second)) {
 		return token, nil
 	}
@@ -59,6 +80,7 @@ func maybeRefresh(ctx context.Context, token *Token) (*Token, error) {
 	return next, nil
 }
 
+// IsOAuthTransport checks wether the underlying type of the given RoundTripper is a OAuthTransport
 func IsOAuthTransport(trp http.RoundTripper) bool {
 	_, ok := trp.(*Transport)
 	return ok


cmd/src/login.go
```diff
@@ -125,6 +126,13 @@ func loginCmd(ctx context.Context, p loginParams) error {
 	noToken := cfg.AccessToken == ""
 	endpointConflict := endpointArg != cfg.Endpoint
 
+	secretStore, err := keyring.Open()
```
keegancsmith: you only open this if doing deviceflow. If we are not doing deviceflow don't open secret storage. IE avoid interacting with secret storage unless we actually need to.

Maybe you can add a wrapper around secretstore which lazily opens?

burmudar: Stop looking over my shoulder :P _busy addressing this atm_ 😄

cmd/src/login.go
```diff
@@ -130,19 +143,20 @@ func loginCmd(ctx context.Context, p loginParams) error {
...
-		cfg.Endpoint = endpointArg
-		client = cfg.apiClient(p.apiFlags, out)
-	} else if noToken || endpointConflict {
-		fmt.Fprintln(out)
-		switch {
-		case noToken:
-			printProblem("No access token is configured.")
-		case endpointConflict:
-			printProblem(fmt.Sprintf("The configured endpoint is %s, not %s.", cfg.Endpoint, endpointArg))
+		if err := oauth.StoreToken(ctx, token); err != nil {
```
burmudar, line 146: opted to not store anything when failing to store in the keyring. This means if we fail storing the user would have to login everytime - not exactly ideal.

We could write the details to disk and warn the user? And everytime we load it we warn them again?

internal/oauth/http_transport.go
```diff
@@ -0,0 +1,65 @@
...
+	prevToken := t.Token
+	token, err := maybeRefresh(ctx, t.Token)
+	if err != nil {
+		return nil, err
+	}
+	t.Token = token
+	if token != prevToken {
+		// try to save the token if we fail let the request continue with in memory token
+		_ = storeRefreshedTokenFn(ctx, token)
+	}
```
keegancsmith: This should all be behind a mutex to prevent concurrent refresh attempts. Additionally then when you read t.Token below that should have come from this critical section.

To make the code easy to read I suggest pulling this out into a method on *Transport like `getToken(ctx)` then you can just have a simple mu.Lock defer mu.Unlock at the top.

internal/secrets/keyring.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/zalando/go-keyring"
+)
+
+var ErrSecretNotFound = errors.New("secret not found")
+
+// Store provides secure credential storage operations.
+type Store interface {
+	Get(key string) ([]byte, error)
+	Put(key string, data []byte) error
+	Delete(key string) error
```
keegancsmith: no one calls this

internal/secrets/keyring.go
```diff
@@ -0,0 +1,86 @@
...
+	"context"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+	"github.com/zalando/go-keyring"
+)
+
+var ErrSecretNotFound = errors.New("secret not found")
+
+// Store provides secure credential storage operations.
+type Store interface {
```
keegancsmith: where do you use this interface?

internal/secrets/keyring.go
```diff
@@ -0,0 +1,86 @@
...
+	case <-ctx.Done():
+		var zero T
+		return zero, ctx.Err()
+	case r := <-ch:
+		return r.val, r.err
+	}
+}
+
+// Put stores a key-value pair in the keyring.
+func (k *keyringStore) Put(key string, data []byte) error {
```
keegancsmith, line 54: given the underlying library is all just global function calls, why don't we just do the same and wrap them with the service name we use?

keegancsmith (commented):
My main concern is hitting libsecret/etc when we don't need to. But the underlying library looks good + we only open the keyring when doing oauth. So LGTM

internal/secrets/keyring.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+type keyringStore struct {
+	ctx         context.Context
+	serviceName string
+}
+
+// Open opens the system keyring for the Sourcegraph CLI.
+func Open(ctx context.Context) (Store, error) {
+	return &keyringStore{ctx: ctx, serviceName: "Sourcegraph CLI"}, nil
```
keegancsmith: I think our service name needs to include the endpoint we are authenticating against. eg this is what it looks like when I look at some of my secrets

burmudar: Agreed - I'll add it

internal/secrets/keyring.go
burmudar: I don't - I'll remove

internal/secrets/keyring.go
burmudar: My main motivation to keep a struct is to avoid making Get / Put param lists be "a lot" e.g.
`Put(ctx context.Context, endpoint, key string, data []byte)`
`Get(ctx context.Context, endpoint, key string)`

We also centralize the store name to only happen once Open is called. Once it is open, the user doesn't have to keep extra context about the endpoint and can just use the store.

internal/secrets/keyring.go
burmudar: Updated

Screenshot 2026-03-09 at 11 26 20

internal/secrets/keyring.go
burmudar: removed

internal/oauth/http_transport.go
burmudar: pulled it out into `refreshToken` with a mutex

keegancsmith: There is a very minor race condition here. You write to Token behind a mutex, but then you read it without a mutex. That is why I suggested `getToken` since refreshing and getting is all the same critical section.

burmudar: Ah man. Sorry misunderstood. I'll submit a fix

PR #1237, opened 2025-12-10
feat(mcp): add tool registry to handle tools from different endpoints

burmudar:
Add a tool registry which keeps tracks of tools and from which endpoints they originated from. We need to know the endpoint so that we a tool call is made it is made against the correct endpoint.

### Test plan
`go run ./cmd/src deepsearch -question 'which repo uses bazel?'`

### DIFF
--- a/internal/mcp/mcp_request.go
+++ b/internal/mcp/mcp_request.go
@@ -112,6 +112,7 @@ func decodeToolResponse(resp *http.Response) (map[string]json.RawMessage, error)
 		Result struct {
 			Content           []json.RawMessage          `json:"content"`
 			StructuredContent map[string]json.RawMessage `json:"structuredContent"`
+			IsError           bool                       `json:"isError"`
 		} `json:"result"`
 		Error *struct {
 			Code    int    `json:"code"`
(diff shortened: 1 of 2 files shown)


internal/mcp/mcp_request.go
```diff
@@ -14,9 +14,10 @@ import (
 )
 
 const McpURLPath = ".api/mcp/v1"
+const McpDeepSearchURLPath = ".api/mcp/deepsearch"
```
keegancsmith: nit: It is MCP not Mcp.

keegancsmith (approved):
Should we just update the server now to include deep search tools on v1? That way we can avoid making this more complicated. Otherwise we can move forward with this.

internal/mcp/mcp_request.go
burmudar: fixed

burmudar:
> Should we just update the server now to include deep search tools on v1? That way we can avoid making this more complicated. Otherwise we can move forward with this.

A part of me think we should. I'd still keep this registry though but remove the endpoint tracking. 

Site note: should it be deep-search instead of deepsearch?

PR #1242, opened 2025-12-11
chore(logs): add verbose logging for batch changes steps

BolajiOlajide:
Setting `-v` to true didn't include information that can be used for debugging. This PR improves that adding detailed logging to several points of the Batch Changes pre-execution and execution steps.

Closes CP-4

### Test plan

Manually running a batch spec with the `src-cli`.


internal/batches/ui/task_exec_tui.go
```diff
@@ -427,50 +428,102 @@ func diffStatDiagram(stat diff.Stat) string {
...
+func (ui stepsExecTUI) WorkspaceInitializationFinished() {
+	ui.out.Verbosef("[%s] Workspace initialization completed", ui.task.Repository.Name)
 }
-func (ui stepsExecTUI) WorkspaceInitializationFinished() {}
+
 func (ui stepsExecTUI) SkippingStepsUpto(startStep int) {
 	switch startStep {
 	case 1:
 		ui.updateStatusBar("Skipping step 1. Found cached result.")
+		ui.out.Verbosef("[%s] Skipping step 1 (cached result found)", ui.task.Repository.Name)
```
burmudar, line 462: Won't we have duplicate output when using `-v`?

```
[              Skipping step 1. Found cached result.             ]
[sourcegraph/sourcegraph] Skipping step 1 (cached result found)
```

or can you provide a screenshot of how this looks? with and without the verbose flag

BolajiOlajide: Yes, however, the statusbar text changes depending on the steps so the text isn't preserved for one to see if they aren't looking.

PR #1243, opened 2025-12-23
feat(lsp): expose Sourcegraph Precise code navigation through LSP

jupblb:
## Foreword (by me)
Today's HackerNews page about [Claude Code getting native LSP support](https://news.ycombinator.com/item?id=46355165) made me think that we should experiment with having a more traditional API of interacting with Precise data than just using MCP or raw GraphQL queries.

So here's something: a regular LSP server that is easy to run with just `src lsp` command. The implementation is surprisingly straightforward. It works great with Neovim (other editors should be fine too).

To test it with Neovim use the following configuration (no plugin needed):
``` lua
-- Remember to build src binary from this PR
-- $ go build -o src ./cmd/src
vim.lsp.config['src-lsp'] = {
  cmd = { '/path/to/src', 'lsp' },
  root_markers = { '.git' },
  filetypes = { 'go', 'typescript', 'javascript', 'python', 'lua', 'rust', 'java', 'c', 'cpp' },
}
vim.lsp.enable('src-lsp')
```

Caveats:
- It will not take into consideration changes made locally.
- Sourcegraph endpoint must have the codebase indexed with SCIP.
- Doesn't handle cross-repo navigation (I'm unsure how this should be handled).
- When searching for references, the code doesn't handle pagination so we get at most 100 results.

## Amp description

This PR adds a new `src lsp` command that implements a Language Server Protocol (LSP) server, proxying requests to Sourcegraph's code intelligence backend via GraphQL. The server communicates over stdio and is designed for use with editors like Neovim.

### Features

- **textDocument/hover** - Get hover information for symbols
- **textDocument/definition** - Go-to-definition support  
- **textDocument/references** - Find all references to a symbol
- **textDocument/documentHighlight** - Highlight all occurrences of a symbol in the current file

### Implementation Details

- Infers repository name and commit from git at startup
- Uses `git merge-base` to find a commit that exists on remote Sourcegraph
- Converts `file://` URIs to repo-relative paths for GraphQL queries

### New Files

- `cmd/src/lsp.go` - Command registration and help
- `internal/lsp/server.go` - LSP server implementation using glsp
- `internal/lsp/query.go` - GraphQL queries for hover, definitions, and references

## Test plan
Added unit tests

### DIFF
--- a/cmd/src/lsp.go
+++ b/cmd/src/lsp.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 
+	"github.com/sourcegraph/src-cli/internal/api"
 	"github.com/sourcegraph/src-cli/internal/lsp"
 )
 
@@ -38,10 +39,12 @@ Example Neovim configuration (0.11+):
 `
 
 	flagSet := flag.NewFlagSet("lsp", flag.ExitOnError)
+	apiFlags := api.NewFlags(flagSet)
+
 	usageFunc := func() {
+		fmt.Println(usage)
 		fmt.Fprintf(flag.CommandLine.Output(), "Usage of 'src %s':\n", flagSet.Name())
 		flagSet.PrintDefaults()
-		fmt.Println(usage)
 	}
 
 	handler := func(args []string) error {
@@ -49,7 +52,7 @@ Example Neovim configuration (0.11+):
 			return err
 		}
 
-		client := cfg.apiClient(nil, io.Discard)
+		client := cfg.apiClient(apiFlags, io.Discard)
 
 		srv, err := lsp.NewServer(client)
 		if err != nil {


peterguy (approved):
Love the idea here, and the code looks solid. I suppose `InferRoot` could now use `GitRoot` if we want.

I'll approve to unblock; @burmudar or @keegancsmith have more familiarity with `src-cli` and may have some particular commentary, but it looks good to me!

cmd/src/lsp.go
```diff
@@ -0,0 +1,67 @@
...
+		flagSet.PrintDefaults()
+		fmt.Println(usage)
+	}
+
+	handler := func(args []string) error {
+		if err := flagSet.Parse(args); err != nil {
+			return err
+		}
+
+		client := cfg.apiClient(nil, io.Discard)
```
burmudar: I think we need to pass flagSet in here?

keegancsmith (approved):
Love the code here. Damn, go ahead and just land it. Nice, simple and self contained.

Given we don't advertise the existance of this command in the help output this is perfect for landing and getting feedback on.

cmd/src/lsp.go
jupblb: Done.

PR #1244, opened 2026-01-05
feat: add blueprints command

trly:
### Test plan
All implemented blueprint sub-commands have been tested locally. This is intended to remain undocumented/internal only for now. Mainly intended for the FE team to work with customers to quickly import pre-built solutions.

### DIFF
--- a/cmd/src/blueprint_import.go
+++ b/cmd/src/blueprint_import.go
@@ -169,7 +169,7 @@ func runBlueprintImportAll(ctx context.Context, opts blueprintImportOpts, rootDi
 		return nil
 	}
 
-	fmt.Fprintf(opts.out, "Found %d blueprint(s) in repository\n\n", len(found))
+	fmt.Fprintf(opts.out, "Found %d blueprint(s) in repository\n", len(found))
 
 	exec := blueprint.NewExecutor(blueprint.ExecutorOpts{
 		Client:          opts.client,


cmd/src/blueprint_import.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func runBlueprintImportSingle(ctx context.Context, opts blueprintImportOpts, blueprintDir string) error {
+	bp, err := blueprint.Load(blueprintDir)
+	if err != nil {
+		return err
+	}
+
+	fmt.Fprintf(opts.out, "Loaded blueprint: %s\n", bp.Name)
+	if bp.Title != "" {
```
burmudar, line 217: I think you can roll all this `fmt.Fprintf` to be part of `bp.String()`

```go
fmt.Fprintln(opts.out, bp.String())
```

internal/blueprint/find.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"io/fs"
+	"path/filepath"
+)
+
+// FindBlueprints recursively searches rootDir for blueprint.yaml files and
+// returns the successfully loaded blueprints. Invalid or malformed blueprints
+// are silently skipped to allow discovery to continue.
+func FindBlueprints(rootDir string) ([]*Blueprint, error) {
```
burmudar, line 11: I think this should  be just `Find`, because we're already in the blueprint package, otherwise we repeat ourselves with `blueprint.FindBlueprints`

internal/blueprint/source.go
```diff
@@ -0,0 +1,112 @@
...
+	"context"
+	"net/url"
+	"path/filepath"
+	"strings"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+// BlueprintSource provides access to blueprint files from various sources.
+type BlueprintSource interface {
```
burmudar, line 13: Maybe just `Source` since outside of the package it will be `blueprint.Source` ?

internal/blueprint/source.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+	return false
+}
+
+// ResolveRootSource returns a BlueprintSource that provides the repository root
+// without requiring a blueprint.yaml file. Used for listing all blueprints in a repository.
+func ResolveRootSource(rawRepo, rev string) (BlueprintSource, error) {
+	if isLocalPath(rawRepo) {
+		path := rawRepo
+		if !filepath.IsAbs(path) {
```
burmudar: after `filepath.IsLocal` you can use `filepath.Locallize` to get a local path

cmd/src/blueprint_import.go
```diff
@@ -0,0 +1,245 @@
...
+		handler:   handler,
+		usageFunc: usageFunc,
+	})
+}
+
+func runBlueprintImport(ctx context.Context, opts blueprintImportOpts) error {
+	var src blueprint.BlueprintSource
+	var err error
+
+	if opts.subdir == "" {
```
burmudar, line 137: nit: maybe this distinction should be made in a top level `blueprint.NewSource` / `blueprint.ResolveSource` method? We then unexport `ResolveRootSource` / `ResolveBlueprintSource` and just have a single `Resolve` method exposed that returns the correct source for us.

cmd/src/blueprint_import.go
```diff
@@ -0,0 +1,245 @@
...
+}
+
+func runBlueprintImportAll(ctx context.Context, opts blueprintImportOpts, rootDir string) error {
+	found, err := blueprint.FindBlueprints(rootDir)
+	if err != nil {
+		return err
+	}
+
+	if len(found) == 0 {
+		fmt.Fprintf(opts.out, "No blueprints found in repository\n")
```
burmudar:
```suggestion
		fmt.Fprintf(opts.out, "No blueprints found in %s\n", rootDir)
```

cmd/src/blueprint_import.go
```diff
@@ -0,0 +1,245 @@
...
+		return nil
+	}
+
+	fmt.Fprintf(opts.out, "Found %d blueprint(s) in repository\n\n", len(found))
```
burmudar:
```suggestion
	fmt.Fprintf(opts.out, "Found %d blueprint(s) in repository\n", len(found))
```
unless the extra `\n` was intentional?

internal/blueprint/executor.go
```diff
@@ -0,0 +1,460 @@
...
+
+// Executor executes a blueprint against a Sourcegraph instance.
+type Executor struct {
+	client          api.Client
+	out             io.Writer
+	vars            map[string]string
+	dryRun          bool
+	continueOnError bool
+	currentUserID   *string
+	dashboardIDs    map[string]string
```
burmudar, line 74: I don't think the executor should keep this state (same with insightIDs). I think this should be returned as part of ExecutionSummary.

internal/blueprint/executor.go
```diff
@@ -0,0 +1,460 @@
...
+type ResourceResult struct {
+	Kind      ResourceKind
+	Name      string
+	Succeeded bool
+	Skipped   bool
+	Err       error
+}
+
+// ExecutionSummary contains the results of executing all resources in a blueprint.
+type ExecutionSummary struct {
```
burmudar, line 36: This should be `ExecutionResult`. It should also contain `dashboardIDs` and `insightIDs` which is now part of `Executor`

internal/blueprint/executor.go
```diff
@@ -0,0 +1,460 @@
...
+const (
+	KindDashboard ResourceKind = "dashboard"
+	KindMonitor   ResourceKind = "monitor"
+	KindInsight   ResourceKind = "insight"
+	KindBatchSpec ResourceKind = "batch-spec"
+	KindLink      ResourceKind = "link"
+)
+
+// ResourceResult captures the outcome of executing a single resource within a blueprint.
+type ResourceResult struct {
```
burmudar, line 27: You might want to add a `Value map[string]string` member here to store the dashboardIDs / InsightIDs from the result, which you can then extract in exec and to `Summary`

internal/blueprint/executor.go
```diff
@@ -0,0 +1,460 @@
...
+		insightIDs:      make(map[string]string),
+	}
+}
+
+// Execute executes all resources defined in the given blueprint.
+func (e *Executor) Execute(ctx context.Context, bp *Blueprint, blueprintDir string) (*ExecutionSummary, error) {
+	summary := &ExecutionSummary{}
+
+	if len(bp.Dashboards) > 0 {
+		fmt.Fprintf(e.out, "\nExecuting dashboard mutations...\n")
```
burmudar, line 100: nit: I'd rather make this method _not_ print anything. It's whomever calls this job to print the result. I can understand the need to see "as it executes" but ultimately this can be done after this method executes.

```go
for _, r := range summary.Resources[KindDashboard] { ... }
```

internal/blueprint/executor.go
```diff
@@ -0,0 +1,460 @@
...
+	Kind      ResourceKind
+	Name      string
+	Succeeded bool
+	Skipped   bool
+	Err       error
+}
+
+// ExecutionSummary contains the results of executing all resources in a blueprint.
+type ExecutionSummary struct {
+	Resources []ResourceResult
```
burmudar, line 37: I think it would be nicer to have `Resources map[ResourceKind]ResourceResult`

internal/blueprint/executor.go
```diff
@@ -0,0 +1,460 @@
...
+		for _, ge := range gqlErrs {
+			msgs = append(msgs, ge.Error())
+		}
+		return errors.Newf("%s %q failed: %s", kind, name, strings.Join(msgs, "; "))
+	}
+	return errors.Wrapf(err, "%s %q failed", kind, name)
+}
+
+// PrintExecutionSummary writes a human-readable summary of execution results.
+func PrintExecutionSummary(out io.Writer, s *ExecutionSummary, dryRun bool) {
```
burmudar, line 431: nit: think this should be `func (s *ExecutionSummary) String() string`

internal/blueprint/source.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+// BlueprintSource provides access to blueprint files from various sources.
+type BlueprintSource interface {
+	// Prepare makes the blueprint directory available and returns its path.
+	// The cleanup function MUST be called when done (may be nil for sources
+	// that don't require cleanup).
+	Prepare(ctx context.Context) (blueprintDir string, cleanup func() error, err error)
```
burmudar, line 17: nit: Would a better name be `Init` or `Setup` ?

cmd/src/blueprint_import.go
```diff
@@ -0,0 +1,245 @@
...
+		src, err = blueprint.ResolveRootSource(opts.repo, opts.rev)
+	} else {
+		src, err = blueprint.ResolveBlueprintSource(opts.repo, opts.rev, opts.subdir)
+	}
+	if err != nil {
+		return err
+	}
+
+	blueprintDir, cleanup, err := src.Prepare(ctx)
+	if cleanup != nil {
```
burmudar, line 147: why not always return a valid cleanup? then you don't have to check it the whole time and can just do `defer cleanup()` with confidence

internal/blueprint/find.go
trly: Good call, I was waffling back and forth on naming for a number of things in this package.

trly:
I'll leave this in draft and ping you and @burmudar once I've done a bit more refactoring.

PR #1247, opened 2026-01-07
fix(api): set content-type json on API requests

evict:
This just adds the `Content-Type: application/json` header when making API requests.

### Test plan
Tested making request to Sourcegraph instance

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -174,6 +174,10 @@ func (c *client) createHTTPRequest(ctx context.Context, method, p string, body i
 	if *c.opts.Flags.trace {
 		req.Header.Set("X-Sourcegraph-Should-Trace", "true")
 	}
+
+	// Set before additional headers, in case of an override.
+	req.Header.Set("Content-Type", "application/json")
+
 	for k, v := range c.opts.AdditionalHeaders {
 		req.Header.Set(k, v)
 	}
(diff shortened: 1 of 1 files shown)


internal/api/api.go
```diff
@@ -174,6 +174,10 @@ func (c *client) createHTTPRequest(ctx context.Context, method, p string, body i
 	if *c.opts.Flags.trace {
 		req.Header.Set("X-Sourcegraph-Should-Trace", "true")
 	}
+
+	// Set before additional headers, in case of an override.
+	req.Header.Set("Content-Type", "application/json")
```
eseliger, line 179: maybe check that we don't already set this header explicitly? I think we do some file upload operations in batch changes, not sure if that confuses it?

eseliger (approved):

internal/api/api.go
evict: The one for code intel gets added to additional headers, that will override ours (intended):

cmd/src/code_intel_upload.go:140
```
	cfg.AdditionalHeaders["Content-Type"] = "application/x-protobuf+scip"
```

for batch changes they set the header after creating the client, which also overrides ours:

internal/batches/service/remote.go:181
```
	request.Header.Set("Content-Type", multipartWriter.FormDataContentType())

	resp, err := svc.client.Do(request)
	if err != nil {
		// Errors passed to pipeWriter.CloseWithError come through here.
		return err
	}
	defer resp.Body.Close()
```

`Header.Set` will replace the existing values (and is case-insensitive).

eseliger: ah great!

PR #1248, opened 2026-01-12
mcp: add README for current state of src mcp experiment

burmudar:
Add README document to explain that we're not continuing further development on the `src mcp` experiment.


burmudar:
> We should rip it out, it will always exist in the history. Either now or in a month or two.

I'll just rip the bandaid off now

PR #1251, opened 2026-01-23
chore: use sourcegraph/actions/go-setup and upgrade mise go to 1.25.6

burmudar:
* use `sourcegraph/actions/go-setup` for go setup
* set go to 1.26 in mise

### DIFF
--- a/cmd/src/code_intel_upload.go
+++ b/cmd/src/code_intel_upload.go
@@ -18,8 +18,6 @@ import (
 	"github.com/sourcegraph/sourcegraph/lib/codeintel/upload"
 	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/sourcegraph/lib/output"
-
-	"github.com/sourcegraph/src-cli/internal/api"
 )
 
 func init() {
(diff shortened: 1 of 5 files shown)


.github/workflows/go-lint.yml
```diff
@@ -17,9 +17,8 @@ jobs:
         id: mise
         run: echo "go-version=$(grep 'golang' mise.toml | cut -d'"' -f2)" >> $GITHUB_OUTPUT
       - name: Set up Go
-        uses: actions/setup-go@v6
+        uses: sourcegraph/actions/go-setup@main
         with:
           go-version: ${{ steps.mise.outputs.go-version }}
-      - name: Enable pulling Go modules from private sourcegraph/sourcegraph
-        run: git config --global url."https://${PRIVATE_TOKEN}@github.com/sourcegraph/".insteadOf "https://github.com/sourcegraph/"
+          private-token: ${{ secrets.PRIVATE_SG_ACCESS_TOKEN }}
```
keegancsmith: why do we need this? We shouldn't need any private modules?

burmudar: ugh good point!

.github/workflows/go-ci.yml
```diff
@@ -21,15 +17,9 @@ jobs:
         run: echo "go-version=$(grep 'golang' mise.toml | cut -d'"' -f2)" >> $GITHUB_OUTPUT
         shell: bash
       - name: Set up Go
-        uses: actions/setup-go@v6
+        uses: sourcegraph/actions/go-setup@main
```
keegancsmith, line 20: why do we have our own action?

burmudar: Mainly for easier private go module setup (even though we do not use it here), but it also makes it easier to update setup-go / checkout at a central place instead of across repositories.

PR #1260, opened 2026-02-25
feature: Add support for HTTP_PROXY, HTTPS_PROXY, and NO_PROXY

peterguy:
Adds support for the standard environment variables `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY`.

`SRC_PROXY` still takes precedence, including over `NO_PROXY`.

### Test plan

Added CI tests validating proxy connection behavior.

#### Manual testing

Install `mitmproxy`, `socat` and `squid` (`brew install mitmproxy socat squid` for me).

Create an SSL cert for `squid`:
```
openssl req -x509 -newkey rsa:2048 \
  -keyout ${HOME}/squid-key.pem -out ${HOME}/squid-cert.pem \
  -days 365 -nodes -subj "/CN=localhost"
cat ${HOME}/squid-key.pem ${HOME}/squid-cert.pem > ${HOME}/squid-proxy.pem
```

Create a config for `squid`:
```
cat << EOF >${HOME}/squid.conf
https_port 8445 cert=${HOME}/squid-proxy.pem
sslproxy_session_cache_size 0
acl localnet src 127.0.0.1/32
http_access allow localhost
acl SSL_ports port 443
acl Safe_ports port 80
acl Safe_ports port 443
acl Safe_ports port 1025-65535
http_access deny !Safe_ports
http_access deny CONNECT !SSL_ports
http_access deny all
http_port 3128
EOF
```

Launch three terminals to run the proxies:
- run `socat -d -d unix-listen:${HOME}/src-proxy.sock,fork tcp:localhost:8080` in one.
- run `mitmproxy -v` in another.
- run squid - `squid -f ${HOME}/squid.conf -N -d 1` - in the third.

In a fourth terminal, in the src-cli repo, build the binary: `go build -o ./src-cli ./cmd/src`

Ensure you have `SRC_ENDPOINT` and `SRC_ACCESS_TOKEN` set in your environment, and unset `SRC_PROXY`. `SRC_ENDPOINT` needs to be `https://`.

Connect with each proxy (`mitmproxy` is 8080, `squid` is 8445) and with no proxy:

```
for x in http://localhost:8080 http://localhost:3128 https://localhost:8445
do
HTTPS_PROXY=${x} ./src-cli login --insecure-skip-verify=true
SRC_PROXY=${x} ./src-cli login --insecure-skip-verify=true
HTTPS_PROXY=${x} NO_PROXY=${${SRC_ENDPOINT#*://}%%/*} ./src-cli login --insecure-skip-verify=true
SRC_PROXY=${x} NO_PROXY=${${SRC_ENDPOINT#*://}%%/*} ./src-cli login --insecure-skip-verify=true
done
HTTPS_PROXY="" SRC_PROXY="" ./src-cli login --insecure-skip-verify=true
SRC_PROXY=${HOME}/src-proxy.sock ./src-cli login --insecure-skip-verify=true
```

You should get 14 "Authenticated as [USER] on [ENDPOINT]" messages in your terminal. You should see four connections in `mitmproxy`, six in `squid`, and several lines in `socat` indicating one connection.

This test ensures:
- all of the proxy environment variables are supported (`SRC_PROXY`, `HTTPS_PROXY`, `NO_PROXY`)
-  `NO_PROXY` trumps `HTTPS_PROXY`
- `SRC_PROXY` trumps `HTTPS_PROXY` and `NO_PROXY`
- both plain HTTP and TLS-enabled proxies are supported
- the ability to connect to UNIX Domain Sockets is preserved
- communication with no proxy at all is successful

Cleanup from the test:
- terminate all running proxy processes.
- `rm -f ${HOME}/squid-proxy.pem ${HOME}/squid-key.pem ${HOME}/squid-cert.pem ${HOME}/squid.conf ${HOME}/src-proxy.sock`

Closes https://linear.app/sourcegraph/issue/CPL-236/src-cli-support-standard-http-proxy


internal/api/api.go
```diff
@@ -99,12 +99,37 @@ func buildTransport(opts ClientOpts, flags *Flags) *http.Transport {
...
 
+// envProxyURL resolves the proxy URL
+// from standard HTTP_PROXY/HTTPS_PROXY/NO_PROXY
+// environment variables for the given endpoint.
+// Returns nil if the endpoint is not a valid URL,
+// no proxy is configured, or the endpoint is excluded.
+func envProxyURL(endpoint string) *url.URL {
+	u, err := url.Parse(endpoint)
+	if err != nil || u.Scheme == "" || u.Host == "" {
+		return nil
```
bahrmichael: What do you think about logging the error here?

internal/api/api.go
```diff
@@ -99,12 +99,37 @@ func buildTransport(opts ClientOpts, flags *Flags) *http.Transport {
...
+	}
+	proxyURL, err := http.ProxyFromEnvironment(&http.Request{URL: u})
+	if err != nil || proxyURL == nil {
+		return nil
```
bahrmichael: What do you think about logging the error here?

internal/api/api.go
```diff
@@ -99,12 +99,37 @@ func buildTransport(opts ClientOpts, flags *Flags) *http.Transport {
 	}
 
 	if opts.ProxyURL != nil || opts.ProxyPath != "" {
+		// Explicit SRC_PROXY configuration takes precedence.
 		transport = withProxyTransport(transport, opts.ProxyURL, opts.ProxyPath)
+	} else if proxyURL := envProxyURL(opts.Endpoint); proxyURL != nil && proxyURL.Scheme == "https" {
```
keegancsmith: why does https:// not work without intervention? (as documented below for http and socks5)

internal/api/proxy.go
```diff
@@ -6,30 +6,58 @@ import (
...
+
+func (c *connWithBufferedReader) Read(p []byte) (int, error) {
+	return c.r.Read(p)
+}
+
+// proxyDialAddr returns proxyURL.Host with a default port appended if one is
+// not already present (443 for https, 80 for http).
+func proxyDialAddr(proxyURL *url.URL) string {
+	addr := proxyURL.Host
+	if _, _, err := net.SplitHostPort(addr); err != nil {
```
keegancsmith: why are you ignoring the host here and instead assuming addr is just the host? I think this could be a bug right? IE if your proxy url has a port specified won't this break?

internal/api/proxy.go
```diff
@@ -6,30 +6,58 @@ import (
...
+}
+
+func (c *connWithBufferedReader) Read(p []byte) (int, error) {
+	return c.r.Read(p)
+}
+
+// proxyDialAddr returns proxyURL.Host with a default port appended if one is
+// not already present (443 for https, 80 for http).
+func proxyDialAddr(proxyURL *url.URL) string {
+	addr := proxyURL.Host
```
keegancsmith: I think you should be using the Hostname and Port methods instead of net.SplitHostPort. For example, I think ipv6 addresses would break since the way they are represented in a URL are different to net.SplitHostPort expects.

internal/api/proxy.go
```diff
@@ -59,62 +87,74 @@ func withProxyTransport(baseTransport *http.Transport, proxyURL *url.URL, proxyP
...
+				// Dial the proxy. For https:// proxies, we TLS-connect to the
+				// proxy itself and force ALPN to HTTP/1.1 to prevent Go from
+				// negotiating HTTP/2 for the CONNECT tunnel. Many proxy servers
+				// don't support HTTP/2 CONNECT, and Go's default Transport.Proxy
+				// would negotiate h2 via ALPN when TLS-connecting to an https://
+				// proxy, causing "bogus greeting" errors. For http:// proxies,
+				// CONNECT is always HTTP/1.1 over plain TCP so this isn't needed.
+				// The target connection (e.g. to sourcegraph.com) still negotiates
+				// HTTP/2 normally through the established tunnel.
+				proxyAddr := proxyDialAddr(proxyURL)
```
keegancsmith: neat!

internal/api/proxy_test.go
```diff
@@ -0,0 +1,432 @@
...
+
+	// Wait for the server to be ready
+	waitForServerReady(t, ln.Addr().String(), useTLS, 5*time.Second)
+
+	scheme := "http"
+	if useTLS {
+		scheme = "https"
+	}
+	pURL, _ := url.Parse(fmt.Sprintf("%s://%s", scheme, ln.Addr().String()))
+	return pURL, ch
```
keegancsmith: I think you can greatly simplify code here (and remove helpers like waitForServerReady and generateTestCert) by using the httptest package. It already provides functions to listen on TLS/etc and does things like wait for the server to be listening.

internal/api/proxy_test.go
```diff
@@ -0,0 +1,432 @@
...
+func TestProxyDialAddr(t *testing.T) {
+	tests := []struct {
+		name string
+		url  string
+		want string
+	}{
+		{"https with port", "https://proxy.example.com:8443", "proxy.example.com:8443"},
+		{"https without port", "https://proxy.example.com", "proxy.example.com:443"},
+		{"http with port", "http://proxy.example.com:8080", "proxy.example.com:8080"},
+		{"http without port", "http://proxy.example.com", "proxy.example.com:80"},
```
keegancsmith: get our favourite AI's to generate some example with ipv4, ipv6 and localhost.

keegancsmith (approved):
Nice! Inline comments, but otherwise LGTM

keegancsmith:
You might wanna rebase, I think william just did some CI updates which include things like go-lint/etc.

internal/api/api.go
peterguy: Since we're connecting to the SG API using HTTP/2, the standard transport tries to connect to the TLS-enabled proxy using HTTP/2 also, which fails if the proxy doesn't support HTTP/2. I haven't found a proxy for local testing that does support HTTP/2; I suspect that most/all proxies that are TLS-enabled don't support HTTP/2.

internal/api/api.go
peterguy: Great idea. I re-did the URL parsing for both the endpoint and the proxy - pushing that out to `main.go` when the program loads, so that users get notified immediately if there are parsing problems for either of them.

internal/api/api.go
peterguy: Great idea. I re-did the URL parsing for both the endpoint and the proxy - pushing that out to main.go when the program loads, so that users get notified immediately if there are parsing problems for either of them.

internal/api/proxy.go
peterguy: Hmmmm, the docs for `net.SplitHostPort` mention IPV6; I'll add some test cases to see if it does work.

internal/api/proxy.go
peterguy: Yeah, that's confusing. I wonder if I goofed and deleted a line or something. All the tests were passing; guess I'd better revisit those also! 😊

cmd/src/main.go
```diff
@@ -227,11 +257,19 @@ func readConfig() (*config, error) {
 				return nil, errors.Newf("Invalid proxy configuration: %w", err)
 			}
 			if !isValidUDS {
-				return nil, errors.Newf("invalid proxy socket: %s", path)
+				return nil, errors.Newf("Invalid proxy socket: %s", path)
```
keegancsmith: you changed this is some other places as well. go convention is lowercase on error messages since they get combined. eg `My outer error: My inner error` doesn't read as nicely as `my outer error: my inner error`

internal/api/api.go
```diff
@@ -98,10 +98,17 @@ func buildTransport(opts ClientOpts, flags *Flags) *http.Transport {
...
+		// Use our custom dialer for:
+		// - unix socket proxies
+		// - TLS=enabled proxies, to force HTTP/1.1 for the CONNECT tunnel.
+		//   Many TLS-enabled proxy servers don't support HTTP/2 CONNECT,
+		//   which Go may negotiate via ALPN, resulting in connection errors.
 		transport = withProxyTransport(transport, opts.ProxyURL, opts.ProxyPath)
 	}
 
+	// For http:// and socks5:// proxies, the cloned
+	// transport's default Proxy handles them correctly without intervention.
```
keegancsmith: if I set SRC_PROXY to a http:// or socks5:// though, won't the default transport ignore it?

internal/api/proxy_test.go
keegancsmith: This advice still stands

internal/api/api.go
```diff
@@ -334,6 +341,6 @@ func (r *request) curlCmd() (string, error) {
 		s += fmt.Sprintf("   %s \\\n", shellquote.Join("-H", k+": "+v))
 	}
 	s += fmt.Sprintf("   %s \\\n", shellquote.Join("-d", string(data)))
-	s += fmt.Sprintf("   %s", shellquote.Join(r.client.opts.Endpoint+"/.api/graphql"))
+	s += fmt.Sprintf("   %s", shellquote.Join(r.client.opts.EndpointURL.String()+"/.api/graphql"))
```
keegancsmith: given you are doing TrimRight before you should probably do it here as well.

internal/api/proxy.go
```diff
@@ -6,30 +6,58 @@ import (
...
 )
 
+type connWithBufferedReader struct {
+	net.Conn
+	r *bufio.Reader
+}
+
+func (c *connWithBufferedReader) Read(p []byte) (int, error) {
+	return c.r.Read(p)
+}
```
keegancsmith: net.Conn documents all methods to be concurrency safe. However, bufio.Reader is not. You need to add in a mutex here.

\> Multiple goroutines may invoke methods on a Conn simultaneously.

internal/api/proxy.go
```diff
@@ -59,62 +87,74 @@ func withProxyTransport(baseTransport *http.Transport, proxyURL *url.URL, proxyP
...
-				// Send the CONNECT request to the proxy to establish the tunnel
-				if _, err := conn.Write([]byte(connectReq)); err != nil {
+				if err := connectReq.Write(conn); err != nil {
 					conn.Close()
 					return nil, err
 				}
 
-				// Read and check the response from the proxy
-				resp, err := http.ReadResponse(bufio.NewReader(conn), nil)
+				br := bufio.NewReader(conn)
```
keegancsmith: It isn't clear to me why you introduced the buffered reader?

cmd/src/main.go
```diff
@@ -118,12 +134,13 @@ type config struct {
 	ProxyURL          *url.URL
 	ProxyPath         string
 	ConfigFilePath    string
+	EndpointURL       *url.URL
```
keegancsmith: minor: but these fields should probably all not be exported. Looks like we don't want to marshal them / they are internally read/set.

cmd/src/main_test.go
```diff
@@ -271,7 +331,11 @@ func TestReadConfig(t *testing.T) {
 			envToken:     "abc",
 			envHeaders:   "foo:bar\nfoo-bar:bar-baz",
 			want: &config{
-				Endpoint:          "https://override.com",
+				Endpoint: "https://override.com",
+				EndpointURL: &url.URL{
+					Scheme: "https",
+					Host:   "override.com",
+				},
```
keegancsmith: is it worth updating all these test cases for this? I'd consider just not comparing this field in the comparsion in the test.

cmd/src/main.go
```diff
@@ -107,6 +109,20 @@ func normalizeDashHelp(args []string) []string {
...
+	}
+	if u.Host == "" {
+		return nil, errors.Newf("Empty host")
+	}
+	return u, nil
+}
+
 var cfg *config
 
 // config represents the config format.
```
keegancsmith: <p class='graphite__hidden'><i>[Re: line +130]</i></p>
I think you should marshal into a struct with just the json fields here. Then from your readConfig function return something which just contains the fields you want to read. Then you don't have both Endpoint and EndpointURL in this struct which is a bit weird.
<p class='graphite__hidden'>See this comment inline on <a href="https://app.graphite.com/github/pr/sourcegraph/src-cli/1260?utm_source=unchanged-line-comment">Graphite</a>.</p>

keegancsmith (changes requested):
Requesting changes since I just wanna re-review once you have addressed the comments. FYI I pushed some commits to resolve the CI issues.

internal/api/proxy.go
```diff
@@ -6,30 +6,58 @@ import (
...
+		}
+		// Preserve HTTP/2 negotiation to the origin when ForceAttemptHTTP2
+		// is enabled. Without this, the manual TLS handshake would not
+		// advertise h2 via ALPN, silently forcing HTTP/1.1.
+		if baseTransport.ForceAttemptHTTP2 && len(cfg.NextProtos) == 0 {
+			cfg.NextProtos = []string{"h2", "http/1.1"}
+		}
+		tlsConn := tls.Client(conn, cfg)
 		if err := tlsConn.HandshakeContext(ctx); err != nil {
 			return nil, err
```
keegancsmith, line 31: I think you need to call tlsConn.Close here to cleanup.

cmd/src/main.go
```diff
@@ -107,6 +109,20 @@ func normalizeDashHelp(args []string) []string {
...
+func parseEndpoint(endpoint string) (*url.URL, error) {
+	u, err := url.ParseRequestURI(strings.TrimSuffix(endpoint, "/"))
+	if err != nil {
+		return nil, err
+	}
+	if !(u.Scheme == "http" || u.Scheme == "https") {
+		return nil, errors.Newf("Invalid scheme %s. Require http or https", u.Scheme)
+	}
+	if u.Host == "" {
+		return nil, errors.Newf("Empty host")
```
bahrmichael:
```suggestion
		return nil, errors.New("empty host")
```

cmd/src/main.go
```diff
@@ -189,9 +207,21 @@ func readConfig() (*config, error) {
...
+	// Lastly, apply endpoint flag if set
+	if endpoint != nil && *endpoint != "" {
+		cfg.Endpoint = *endpoint
+	}
+
+	if endpointURL, err := parseEndpoint(cfg.Endpoint); err != nil {
+		return nil, errors.Newf("invalid endpoint: %s", cfg.Endpoint)
+	} else {
+		cfg.EndpointURL = endpointURL
+		cfg.Endpoint = endpointURL.String()
```
bahrmichael: Why do we store both EndpointURL and Endpoint?

cmd/src/main.go
```diff
@@ -227,11 +257,19 @@ func readConfig() (*config, error) {
...
 			cfg.ProxyPath = path
 		} else {
-			return nil, errors.Newf("invalid proxy endpoint: %s", cfg.Proxy)
+			return nil, errors.Newf("Invalid proxy endpoint: %s", cfg.Proxy)
+		}
+	} else {
+		// no SRC_PROXY; check for the standard proxy env variables HTTP_PROXY, HTTPS_PROXY, and NO_PROXY
+		if u, err := http.ProxyFromEnvironment(&http.Request{URL: cfg.EndpointURL}); err != nil {
+			// when there's an error, the value for the env variable is not a legit URL
+			return nil, fmt.Errorf("invalid HTTP_PROXY or HTTPS_PROXY value: %w", err)
```
bahrmichael: To stay consistent you probably want to use errors.Newf here

```suggestion
			return nil, errors.Newf("invalid HTTP_PROXY or HTTPS_PROXY value: %w", err)
```

internal/api/proxy.go
peterguy: oh, thanks!

internal/api/api.go
peterguy: I re-worked this section

internal/api/proxy.go
peterguy: Ah, yes.

internal/api/proxy.go
peterguy: Because I didn't want to lose any bytes after the headers that the reader may have buffered, which bytes could be the start of the tunneled connection data. Granted, it's probably not really a problem because the SG endpoint will in nearly every scenario also be TLS, and ["TLS server will not speak until spoken to."](https://go.dev/src/net/http/transport.go), so there will not be any bytes beyond the header. But if the SG endpoint is plain http, we could lose bytes if we discard the buffered reader.

internal/api/api.go
peterguy: Once the URL is created after trimming the terminating slash, it doesn't need to be trimmed again.

cmd/src/main_test.go
peterguy: Now that we're storing only the URL, yes, we do. :-)

internal/api/proxy.go
peterguy: After more analysis, removed the buffered reader - the CONNECT won't actually cause anything passed through to the target server, so there's no danger of extra bytes from the target server.

internal/api/proxy_test.go
peterguy: I _think_ I finally fixed up these tests?

peterguy:
Alright; I think this is cleaned up and ready for a re-review.

internal/api/proxy.go
```diff
@@ -53,82 +50,7 @@ func withProxyTransport(baseTransport *http.Transport, proxyURL *url.URL, proxyP
...
-					return nil, err
-				}
-				return handshakeTLS(ctx, conn, addr)
-			}
-			baseTransport.DialContext = dial
-			baseTransport.DialTLSContext = dialTLS
-			// clear out any system proxy settings
-			baseTransport.Proxy = nil
-		}
+		baseTransport.Proxy = http.ProxyURL(proxyURL)
```
peterguy, line 53: Turns out, the whole custom dialer was breaking the ability to connect through TLS-enabled proxies. Simply removing it and just adding the proxy to the transport cleared everything up.

internal/api/proxy.go
```diff
@@ -15,22 +12,22 @@ import (
...
-			ServerName: host,
-			// Pull InsecureSkipVerify from the target host transport
-			// so that insecure-skip-verify flag settings are honored for the proxy server
-			InsecureSkipVerify: baseTransport.TLSClientConfig.InsecureSkipVerify,
-		})
+		cfg := baseTransport.TLSClientConfig.Clone()
+		if cfg.ServerName == "" {
+			cfg.ServerName = host
+		}
+		tlsConn := tls.Client(conn, cfg)
```
peterguy, line 28: Not entirely necessary, but cloning the client config preserves all of the settings, like `InsecureSkipVerify`.

internal/api/proxy.go
```diff
@@ -15,22 +12,22 @@ import (
...
 		if err := tlsConn.HandshakeContext(ctx); err != nil {
+			tlsConn.Close()
```
peterguy, line 30: This is probably important - closing the connection on error. Should be closed when the program terminates, but explicitly closing now is safer.

bahrmichael (approved):
Thank you for simplifying this so much! I'm surprised that we can delete so much code, but as you said it still works after removing the socks5 code.

Is the PR description still up to date?

keegancsmith (approved):
Sorry didn't realise it would wait until I approve before allowing you to merge.

PR #1261, opened 2026-02-26
chore: set go version to 1.26

burmudar:
set go version to 1.26 in mise

>[!NOTE]
> There are no changes to github workflows since they all reference the mise.toml file.


burmudar:
Closing in favor of #1251

PR #1267, opened 2026-03-07
refactor: validate and parse the endpoint and proxy at program load

peterguy:
## Summary

Parse the endpoint URL at program load instead of passing around an unparsed string and needing to check/parse it in multiple places.

### `config` struct (`cmd/src/main.go`)
- Split out a `configFromFile` struct containing the JSON elements
- Un-export fields (all consumers are in `package main`)
- Use `endpointURL` - `Endpoint` and `Proxy` are now in `configFromFile`
- Parse `endpointURL` in `readConfig` - consumers use the parsed URL

### `api.ClientOpts` (`internal/api/api.go`)
- `Endpoint string` → `EndpointURL *url.URL`
- Use `JoinPath` for URL construction instead of string concatenation with `TrimRight`

### `login` command (`cmd/src/login.go`)
- `loginCmd` now uses the config's `endpointURL`
- Endpoint conflict detection moved from `loginCmd` into the handler, where the CLI arg is parsed and compared before entering the command

### `code_intel_upload` (`cmd/src/code_intel_upload.go`)
- `makeCodeIntelUploadURL` uses a stack copy (`u := *cfg.endpointURL`) instead of re-parsing the URL
- Error return removed since there's nothing that can fail

### Tests
- Updated all test call sites to pass `*url.URL`

## Testing

All affected packages pass:
```
ok  github.com/sourcegraph/src-cli/cmd/src
ok  github.com/sourcegraph/src-cli/internal/api
ok  github.com/sourcegraph/src-cli/internal/batches/repozip
ok  github.com/sourcegraph/src-cli/internal/batches/executor
```

### DIFF
--- a/internal/oauth/flow.go
+++ b/internal/oauth/flow.go
@@ -328,7 +328,7 @@ func (c *httpClient) pollOnce(ctx context.Context, tokenEndpoint, deviceCode str
 func (c *httpClient) Refresh(ctx context.Context, token *Token) (*TokenResponse, error) {
 	endpointURL, err := token.EndpointURL()
 	if err != nil {
-		return nil, errors.Wrap(err, "invlaid token endpoint")
+		return nil, errors.Wrap(err, "invalid token endpoint")
 	}
 
 	config, err := c.Discover(ctx, endpointURL)


cmd/src/code_intel_upload.go
```diff
@@ -191,16 +187,12 @@ func printInferredArguments(out *output.Output) {
...
 // makeCodeIntelUploadURL constructs a URL to the upload with the given internal identifier.
 // The base of the URL is constructed from the configured Sourcegraph instance.
-func makeCodeIntelUploadURL(uploadID int) (string, error) {
-	url, err := url.Parse(cfg.Endpoint)
-	if err != nil {
-		return "", err
-	}
-
+func makeCodeIntelUploadURL(uploadID int) string {
+	u := *cfg.endpointURL
```
peterguy: copy by dereference makes a shallow copy - affects User mostly/only - should we include a comment to that effect here?

cmd/src/login.go
```diff
@@ -44,17 +44,21 @@ Examples:
...
-		if endpoint == "" {
-			return cmderrors.Usage("expected exactly one argument: the Sourcegraph URL, or SRC_ENDPOINT to be set")
+			arg := flagSet.Arg(0)
+			parsed, err := parseEndpoint(arg)
+			if err != nil {
+				return cmderrors.Usage(fmt.Sprintf("invalid endpoint URL: %s", arg))
+			}
+			if parsed.String() != cfg.endpointURL.String() {
+				return cmderrors.Usage(fmt.Sprintf("endpoint argument %s conflicts with configured endpoint %s", parsed, cfg.endpointURL))
+			}
```
peterguy: The existence of this check calls into question the command-line parameter - `SRC_ENDPOINT` (or the value from the config file) is the source[1] of truth, forcing the user to put the same value on the command line.
Adding to the confusion is that if there is nothing in the config file or in `SRC_ENDPOINT`, `cfg.endpointURL` defaults to `https://sourcegraph.com`, which is probably wrong pretty much all of the time, and still invalidates anything different on the command line.
I kept the error condition and message (although I did change the semantics by changing from an exit error to a usage error) because I didn't want to rip that rug out from under anyone at this point.
Thoughts?

[1] no pun intended, but I do like it

cmd/src/login.go
```diff
@@ -77,23 +79,16 @@ func loginCmd(ctx context.Context, cfg *config, client api.Client, endpointArg s
...
 
    To verify that it's working, run the login command again.
-`, endpointArg, endpointArg)
+`, cfg.endpointURL, cfg.endpointURL)
 
-	if cfg.ConfigFilePath != "" {
+	if cfg.configFilePath != "" {
 		fmt.Fprintln(out)
-		fmt.Fprintf(out, "⚠️  Warning: Configuring src with a JSON file is deprecated. Please migrate to using the env vars SRC_ENDPOINT, SRC_ACCESS_TOKEN, and SRC_PROXY instead, and then remove %s. See https://github.com/sourcegraph/src-cli#readme for more information.\n", cfg.ConfigFilePath)
+		fmt.Fprintf(out, "⚠️  Warning: Configuring src with a JSON file is deprecated. Please migrate to using the env vars SRC_ENDPOINT, SRC_ACCESS_TOKEN, and SRC_PROXY instead, and then remove %s. See https://github.com/sourcegraph/src-cli#readme for more information.\n", cfg.configFilePath)
```
peterguy: The config file has been deprecated for a long time, but if it ever is finally removed, the ability to set additional headers will go also. Do we need to keep the ability to set additional headers around? If so, we should probably introduce yet another environment variable for those and delay the config file removal. If not, let's complete the config file deprecation!

peterguy:
This change is part of the following stack:

- #1267 ◀
    - #1260

<sub>Change managed by [git-spice](https://abhinav.github.io/git-spice/).</sub>

cmd/src/code_intel_upload.go
bahrmichael: Anything that surprised you or had you dig deeper to understand, is probably worth a comment helping us in the future :)

cmd/src/login.go
bahrmichael: I think it's reasonable to keep backwards maintainability in this PR, but follow up with one to make the breaking changes. That makes it easier to isolate and revert if necessary.

cmd/src/login.go
bahrmichael: See https://github.com/sourcegraph/src-cli/pull/1267/changes#r2903847768

internal/api/api.go
```diff
@@ -159,7 +158,7 @@ func (c *client) NewHTTPRequest(ctx context.Context, method, p string, body io.R
 }
 
 func (c *client) createHTTPRequest(ctx context.Context, method, p string, body io.Reader) (*http.Request, error) {
-	req, err := http.NewRequestWithContext(ctx, method, strings.TrimRight(c.opts.Endpoint, "/")+"/"+p, body)
+	req, err := http.NewRequestWithContext(ctx, method, c.opts.EndpointURL.JoinPath(p).String(), body)
```
bahrmichael: Could you verify whether this turns `.api/search/stream?q=repo:foo` into `https://sourcegraph.com/.api/search/stream%3Fq=repo:foo` (probably wrong) or `https://sourcegraph.com/.api/search/stream?q=repo:foo` (probably correct)?

cmd/src/code_intel_upload.go
keegancsmith: You can use `cfg.endpointURL.JoinPath`. If the first path element starts with `/` it will make an absolute path. But even just using something relative should be safe here.

Edit: I see you already are using JoinPath in other parts of this refactor

cmd/src/batch_remote.go
```diff
@@ -157,7 +157,7 @@ Examples:
 
 		executionURL := fmt.Sprintf(
 			"%s/%s/batch-changes/%s/executions/%s",
-			strings.TrimSuffix(cfg.Endpoint, "/"),
+			cfg.endpointURL,
```
keegancsmith: switch to using JoinPath?

cmd/src/login.go
keegancsmith: I believe you are changing the behaviour here. The old behaviour ignored SRC_ENDPOINT for login if the argument was set.  The old error message was misleading

Edit: I see in the loginCmd function they checked if cfg.Endpoint is the same. I imagine that was a mistake. I agree with you about how this should work. But for now lets keep the same behaviour.

internal/api/api.go
keegancsmith: Yeah this is a breaking change. Unfortunetly you need to use the old string concat here.

cmd/src/batch_remote.go
peterguy: Maybe? It is a pretty complex concatenation. Here are a few options for comparison:
Current:
```
		executionURL := fmt.Sprintf(
			"%s/%s/batch-changes/%s/executions/%s",
			cfg.endpointURL,
			strings.TrimPrefix(namespace.URL, "/"),
			batchChangeName,
			batchSpecID,
		)
```

Proposal A:
```
		executionURL := cfg.endpointURL.JoinPath(
			fmt.Sprintf(
				"%s/batch-changes/%s/executions/%s", 
				namespace.URL,
				batchChangeName,
				batchSpecID,
			),
		).String()
```

Proposal B:
```
		executionURL := endpointURL.
			JoinPath(namespace.URL).
			JoinPath("batch-changes").
			JoinPath(batchChangeName).
			JoinPath("executions").
			JoinPath(batchSpecID).
			String()
```

I think proposal A is the easiest to read, but also it's nearly the same as the original `fmt.Sprintf`.

internal/api/api.go
peterguy: Drat. Thanks for catching that.

cmd/src/login.go
peterguy: For illustration, here's how it works currently (before these changes):

```
▶ SRC_ENDPOINT=https://sourcegraph.coursegraph.com src login https://sourcegraph.com

❌ Problem: The configured endpoint is https://sourcegraph.coursegraph.com, not https://sourcegraph.com.

🛠  To fix: Create an access token by going to https://sourcegraph.com/user/settings/tokens, then set the following environment variables in your terminal:

   export SRC_ENDPOINT=https://sourcegraph.com
   export SRC_ACCESS_TOKEN=(your access token)

   To verify that it's working, run the login command again.
```

And here's how it works after these changes:
```
▶ SRC_ENDPOINT=https://sourcegraph.coursegraph.com ./src login https://sourcegraph.com
error: The configured endpoint is https://sourcegraph.coursegraph.com, not https://sourcegraph.com

'src login' helps you authenticate 'src' to access a Sourcegraph instance with your user credentials.

Usage:

    src login SOURCEGRAPH_URL

Examples:
...
```

cmd/src/code_intel_upload.go
peterguy: Added a comment warning about the shallow copy

cmd/src/batch_remote.go
peterguy: I'm going to use proposal A unless someone has a strong opinion otherwise.

cmd/src/login.go
```diff
@@ -107,15 +111,13 @@ func loginCmd(ctx context.Context, p loginParams) error {
...
+	if p.loginEndpointURL != nil && p.loginEndpointURL.String() != p.cfg.endpointURL.String() {
+		return loginFlowEndpointConflict, runEndpointConflictLogin
+	}
 	switch p.cfg.AuthMode() {
 	case AuthModeOAuth:
 		return loginFlowOAuth, runOAuthLogin
 	case AuthModeAccessToken:
-		if endpointArg != p.cfg.Endpoint {
-			return loginFlowEndpointConflict, runEndpointConflictLogin
-		}
```
peterguy, line 118: @burmudar Did I do ok when I I moved the conflicting endpoint handling out of the PAT case?
I figured it should be handled no matter the auth type.
Note that this conflicting endpoint handling should be short-lived: I'm planning a follow-up PR to remove the support for a command-line argument - didn't want to introduce breaking changes to this "refactor"

cmd/src/login.go
```diff
@@ -48,23 +49,26 @@ Examples:
...
-			endpoint:    endpoint,
-			out:         os.Stdout,
-			apiFlags:    apiFlags,
-			oauthClient: oauth.NewClient(oauth.DefaultClientID),
+			cfg:              cfg,
+			client:           client,
+			out:              os.Stdout,
+			apiFlags:         apiFlags,
+			oauthClient:      oauth.NewClient(oauth.DefaultClientID),
+			loginEndpointURL: loginEndpointURL,
```
burmudar, line 71:
```suggestion
			endpointURL: loginEndpointURL,
```

cmd/src/login.go
```diff
@@ -76,12 +80,12 @@ Examples:
...
-	endpoint    string
-	out         io.Writer
-	apiFlags    *api.Flags
-	oauthClient oauth.Client
+	cfg              *config
+	client           api.Client
+	out              io.Writer
+	apiFlags         *api.Flags
+	oauthClient      oauth.Client
+	loginEndpointURL *url.URL
```
burmudar, line 88:
```suggestion
	endpointURL *url.URL
```

cmd/src/login.go
burmudar: I think it's fine but we have to then establish a precedence for flag vars and config vars. I don't think we should remove the arg, but more just give a warning to the user that we are picking a particular value.

internal/oauth/flow.go
```diff
@@ -326,7 +326,12 @@ func (c *httpClient) pollOnce(ctx context.Context, tokenEndpoint, deviceCode str
 
 // Refresh exchanges a refresh token for a new access token.
 func (c *httpClient) Refresh(ctx context.Context, token *Token) (*TokenResponse, error) {
-	config, err := c.Discover(ctx, token.Endpoint)
+	endpointURL, err := token.EndpointURL()
+	if err != nil {
+		return nil, errors.Wrap(err, "invlaid token endpoint")
```
burmudar:
```suggestion
		return nil, errors.Wrap(err, "invalid token endpoint")
```

cmd/src/main.go
```diff
@@ -110,17 +110,42 @@ func normalizeDashHelp(args []string) []string {
...
 
-// config represents the config format.
+// config holds the resolved configuration used at runtime.
 type config struct {
+	accessToken       string
+	additionalHeaders map[string]string
+	proxyURL          *url.URL
+	proxyPath         string
+	configFilePath    string
+	endpointURL       *url.URL // always non-nil; defaults to https://sourcegraph.com via readConfig
```
burmudar, line 139: if we are sure it will never be nil, then maybe we should just make it url.URL and not a pointer?

cmd/src/login.go
```diff
@@ -107,15 +111,13 @@ func loginCmd(ctx context.Context, p loginParams) error {
 
 // selectLoginFlow decides what login flow to run based on configured AuthMode.
 func selectLoginFlow(p loginParams) (loginFlowKind, loginFlow) {
-	endpointArg := cleanEndpoint(p.endpoint)
-
+	if p.loginEndpointURL != nil && p.loginEndpointURL.String() != p.cfg.endpointURL.String() {
```
burmudar, line 114: if we just make cfg the source of truth and the cli endpoint arg just overrides it we don't have to do this check at all.

cmd/src/login_oauth.go
```diff
@@ -18,15 +18,14 @@ import (
 var loadStoredOAuthToken = oauth.LoadToken
 
 func runOAuthLogin(ctx context.Context, p loginParams) error {
-	endpointArg := cleanEndpoint(p.endpoint)
-	client, err := oauthLoginClient(ctx, p, endpointArg)
+	client, err := oauthLoginClient(ctx, p)
 	if err != nil {
 		printLoginProblem(p.out, fmt.Sprintf("OAuth Device flow authentication failed: %s", err))
-		fmt.Fprintln(p.out, loginAccessTokenMessage(endpointArg))
+		fmt.Fprintln(p.out, loginAccessTokenMessage(p.cfg.endpointURL))
```
burmudar, line 24: Note that we are preferring the configured endpointURL over the provided argument endpoint here

cmd/src/login.go
```diff
@@ -48,23 +49,26 @@ Examples:
 		if err := flagSet.Parse(args); err != nil {
 			return err
 		}
-		endpoint := cfg.Endpoint
+
```
burmudar, line 52: I think after successful parse, we should override whatever is configured in cfg for the endpoint. 

So cfg should be source of truth hence forth.

cmd/src/login.go
burmudar, line 88: We're using `cfg.EndpointURL` and never use this endpointURL anymore - so we can remove it

cmd/src/main.go
peterguy: Yeah, I considered that. I do like the signal it sends: using a value says it's a constant, do not modify, owned by the config/program, immutable, non-optional, no nil dereferencing panics.

Propagating it through the code introduces enough cognitive friction that I have hesitated to do it:
- `.String()` usage scattered everywhere that are currently implicit via `%s` on `*url.URL`
- The url parse functions all deal in pointers, so there is inevitable dereferencing friction 

I just don't think we gain enough from making it a value to warrant the widespread semantic awkwardness we'd have to deal with afterward.

[diff](https://gist.github.com/peterguy/d8da35d43430b29e4e9ee47fca701dbd)

WDYT?

cmd/src/login.go
peterguy: Yes, but that would break current behavior. As confusing as the current behavior is, I was trying to keep that behavior, changing it in a followup PR.

cmd/src/login.go
peterguy: Indeed it should, and after a followup PR, it will be!

cmd/src/login.go
peterguy: That'll be the intent of a follow up PR

PR #1269, opened 2026-03-09
fix: oauth token race in http_transport

burmudar:
Fix a read race mentioned [here](https://github.com/sourcegraph/src-cli/pull/1228#discussion_r2905365147) by @keegancsmith

### Test plan
Unit tests

### DIFF
--- a/internal/oauth/http_transport.go
+++ b/internal/oauth/http_transport.go
@@ -12,9 +12,12 @@ var _ http.Transport
 var _ http.RoundTripper = (*Transport)(nil)
 
 type Transport struct {
-	Base  http.RoundTripper
+	Base http.RoundTripper
+	//Token is a OAuth token (which has a refresh token) that should be used during roundtrip to automatically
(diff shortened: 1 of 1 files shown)


internal/oauth/http_transport.go
```diff
@@ -29,16 +32,25 @@ func (t *Transport) RoundTrip(req *http.Request) (*http.Response, error) {
 	if err := t.refreshToken(ctx); err != nil {
 		return nil, err
 	}
+	token := t.getToken()
```
keegancsmith, line 35: very minor: you lock in refresh and in getToken, and this is the only place you call both. Maybe instead you can refactor it to read like this
```suggestion
	token, err := t.getToken(ctx)
	if err != nil {
		return nil, err
	}
```

burmudar: yeah I tend to agree. It did feel bit 💩 to do it - only thing that stopped me was `getToken` does more than what it says, although it's minor in the long run.

I'll update it in a bit.

PR #1272, opened 2026-03-10
fix/security: prevent command injection on windows

evict:
We should verify the result from the server, and make sure it's a URL. Additionally, the `start` pattern was dangerous and allows command injection, so we use `OpenURL` from `url.dll` directly.


cmd/src/login_oauth.go
```diff
@@ -103,20 +104,39 @@ func runOAuthDeviceFlow(ctx context.Context, endpoint string, out io.Writer, cli
...
+
 	var cmd *exec.Cmd
 	switch runtime.GOOS {
 	case "darwin":
-		cmd = exec.Command("open", url)
+		cmd = exec.Command("open", rawURL)
 	case "windows":
-		// "start" is a cmd.exe built-in; the empty string is the window title.
-		cmd = exec.Command("cmd", "/c", "start", "", url)
+		cmd = exec.Command("rundll32", "url.dll,OpenURL", rawURL)
```
burmudar, line 138: I feel this would trigger more alarms, why not `cmd /c start`? Directly executing a DLL seems odd - even though it works

burmudar: Ok re-read your PR description. `rundll32` still seems more dangerous than `cmd /c start` to my unsecure brain 😆

Have we tested that `rundll32` doesnt trigger a Admin priviliges window popup?

burmudar: Aight. Just my dll spider sense that triggered for nothing

cmd/src/login_oauth.go
```diff
@@ -103,20 +104,39 @@ func runOAuthDeviceFlow(ctx context.Context, endpoint string, out io.Writer, cli
...
 }
 
-func openInBrowser(url string) error {
-	if url == "" {
+// validateBrowserURL checks that rawURL is a valid HTTP(S) URL to prevent
+// command injection via malicious URLs returned by an OAuth server.
+func validateBrowserURL(rawURL string) error {
+	u, err := url.Parse(rawURL)
+	if err != nil {
+		return fmt.Errorf("invalid URL: %w", err)
```
burmudar: Think these have to use lib/error

evict: oh woops, yes !

cmd/src/login_oauth.go
cbrnrd: This doesn't really solve the problem, just avoids it. The issue, as i understand it, was that `url` could include escape characters and therefore lead to command injection. Why do this and not just shellquote the `url` value?

evict: Because now we have an `exec` call without any sort of expansion it's now just a parameter. This is more robust, rather than sanitizing a call to what's essentially a [system](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/system-wsystem) call. We already validate the URL now, but we don't have to worry about it being expanded or any edge cases with quoting.

PR #1274, opened 2026-03-17
feat: no longer accept the endpoint on the command line

peterguy:
The endpoint on the command line always had to match `SRC_ENDPOINT`, which defaults to `https://sourcegraph.com`; there was no point to supporting the command-line argument.

This could be considered a "breaking change" because the error `The configured endpoint is [SRC_ENDPOINT], not [command-line-arg].` Will no longer show up.

It has been replaced by a usage error: `src login no longer accepts a URL argument; set SRC_ENDPOINT instead`

Before:
CleanShot 2026-03-16 at 20 50 07@2x

After:
CleanShot 2026-03-16 at 20 50 36@2x

### Test plan

CI tests updated.

### DIFF
--- a/cmd/src/login.go
+++ b/cmd/src/login.go
@@ -9,6 +9,7 @@ import (
 	"os"
 
 	"github.com/sourcegraph/src-cli/internal/api"
+	"github.com/sourcegraph/src-cli/internal/cmderrors"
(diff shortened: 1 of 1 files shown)


keegancsmith (changes requested):
As mentioned in the comments, lets not remove it but rather change what we do with the argument.

PR #1275, opened 2026-03-17
feat/auth: add `src auth token`

burmudar:
This adds the command `src auth token`, which will print either your access token or your oauth token. This is primarily useful for scenarios like `curl -H "Authorization: Bearer $(src auth token)" https://sourcegraph.sourcegraph.com/.api/something`.

If no SRC_ACCESS_TOKEN is specified and we use the OAuth token and the token is close to expiring or has expired it will be automatically refreshed

* Centralized the refreshing logic to `TokenRefresher` which the oauth transport now uses as well as the OAuth token
* Create a constructor `NewTransport` so that TokenRefresher is properly initialized for the OAuth.Transport

Closes CPL-274

### Future work

- support `src auth status` and tell you if we're using `SRC_ACCESS_TOKEN` or `OAuth` and with which endpoint

### Test plan
```
src-cli on  wb/src-auth [$?] via 🐹 v1.25.5
❯ export SRC_ENDPOINT=https://sourcegraph.sourcegraph.com; go run ./cmd/src/... auth token
sgo_at_DEADBEEF

src-cli on  wb/src-auth [$?] via 🐹 v1.25.5 took 2s
❯ export SRC_ENDPOINT=https://sourcegraph.sourcegraph.com; export SRC_ACCESS_TOKEN='hello-reviewer'; go run ./cmd/src/... auth token
hello-reviewer
```
When there is no OAuth Token and no SRC_ACCESS_TOKEN
```
export SRC_ENDPOINT=https://sourcegraph.sourcegraph.com; go run ./cmd/src/... auth token
error loading OAuth token; set SRC_ACCESS_TOKEN or run `src login`: secret not found
```

### DIFF
--- a/internal/oauth/http_transport.go
+++ b/internal/oauth/http_transport.go
@@ -13,7 +13,7 @@ var _ http.Transport
 
 var _ http.RoundTripper = (*Transport)(nil)
 
-const defaultRefreshWindow = 30 * time.Second
+const defaultRefreshWindow = 5 * time.Minute
 
 type Transport struct {
 	Base      http.RoundTripper


internal/oauth/http_transport.go
```diff
@@ -13,14 +13,24 @@ var _ http.Transport
 
 var _ http.RoundTripper = (*Transport)(nil)
 
+const defaultRefreshWindow = 30 * time.Second
```
eseliger: I think 30s is quite optimistic, if I pass this to a bash script that iterates a few pages 30s will quickly give me an expired token. In sg/sg we refresh tokens with less than 5m of validity, should we mirror that?

burmudar: Best to be consistent! Will update it

PR #1277, opened 2026-03-23
chore: bump Go and vulnerable dependencies

evict:
Updates the Go toolchain and a few dependencies to resolve vulnerabilities.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/sourcegraph/src-cli
 
-go 1.26.1
+go 1.25.1
 
 require (
 	cloud.google.com/go/storage v1.50.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/sourcegraph/src-cli
 
-go 1.24.1
```
keegancsmith, line 3: actually can you keep this behind by 1 major version. This is declaring what the minimum version of go is needed to build this project.  
  
For example we have linux package managers that auto-update src-cli. They aren't as quick as updating go -> we no longer get auto-updates of src-cli in them.

The important thing to bump is in .tool-versions for our CI which does the releasing

PR #1278, opened 2026-03-25
bug: only require CI=true and AccessToken when making API requests

burmudar:
In https://github.com/sourcegraph/sourcegraph/actions/runs/23518169374/job/68455311352?pr=11118 the action failed on command `src version -client-only`, where it should not have failed because it's a client only check and no API interaction thus not requiring a Access Token to be set.

In this PR we update the config:
* record whether we are in CI

Update APIClient:
* add a check to see if we're in CI and our AuthMode is "AccessToken"
* fail on any request if we require a CI AccessToken and it is not set.

Closes CPL-291
### Test plan
unit tests + manual tests

### DIFF
--- a/cmd/src/main_test.go
+++ b/cmd/src/main_test.go
@@ -3,7 +3,6 @@ package main
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"io"
 	"net/url"
 	"os"
@@ -13,6 +12,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 	"github.com/sourcegraph/src-cli/internal/api"
 )
 
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -6,7 +6,6 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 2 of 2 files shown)


cmd/src/main.go
```diff
@@ -162,16 +163,29 @@ func (c *config) AuthMode() AuthMode {
 	return AuthModeOAuth
 }
 
+func (c *config) InCI() bool {
+	return c.inCI
+}
+
+func (c *config) requireCIAccessToken() error {
```
keegancsmith, line 170: this needs a doc string. Additionally at callsites this reads kinda weird. Maybe it should be called `checkIfCIAccessTokenRequired` (matching up with the err it returns)

cmd/src/main.go
```diff
@@ -162,16 +163,29 @@ func (c *config) AuthMode() AuthMode {
...
+	if c.InCI() && c.AuthMode() != AuthModeAccessToken {
+		return errCIAccessTokenRequired
```
keegancsmith, line 175: there might be an access token set but it is oauth instead? Maybe we need to include a little more context when this fails to help debug weird issues. Or does oauth + CI seem impossible?

burmudar: I'll add some context

> Or does oauth + CI seem impossible?

Not entirely impossible but it's definitely the more difficult path. If there is no keyring the OAuth token you get will only be used for that session and then forgotten (we do warn when this happens)

cmd/src/main.go
burmudar: updated

cmd/src/main.go
burmudar: We also can't use OAuth access tokens as `SRC_ACCESS_TOKEN` values, since OAuth requires `Bearer` in the Authorization header and `SRC_ACCESS_TOKEN` requires `token`

burmudar: Updated to provide more context in the error!

PR #1279, opened 2026-03-26
security: use os.Root API for file path logic

cbrnrd:
# Problem

Mount path validation in `src batch` uses `strings.HasPrefix` on non-canonical paths and `os.Stat` which follows symlinks. This allowed two bypass techniques:

* Path prefix confusion — a path like `/tmp/specdir-123-outside/secret.txt` passes a HasPrefix check against `/tmp/specdir-123`
* Symlink traversal — a symlink inside the spec directory pointing outside it passes both `os.Stat` and the prefix check

# Solution

Replace all os.Stat/os.Open/os.ReadDir + strings.HasPrefix path validation with Go's os.OpenRoot API, which enforces directory confinement at the OS level — blocking symlink escapes, .. traversal, and prefix confusion.

### Test plan
Added new tests which pass, also verified locally.

### DIFF
--- a/cmd/src/auth_token.go
+++ b/cmd/src/auth_token.go
@@ -54,6 +54,10 @@ func init() {
 }
 
 func resolveAuthToken(ctx context.Context, cfg *config) (string, error) {
+	if err := cfg.requireCIAccessToken(); err != nil {
+		return "", err
+	}
+
 	if cfg.accessToken != "" {
 		return cfg.accessToken, nil
 	}
--- a/cmd/src/auth_token_test.go
+++ b/cmd/src/auth_token_test.go
@@ -35,6 +35,28 @@ func TestResolveAuthToken(t *testing.T) {
 		}
 	})
 
+	t.Run("requires access token in CI", func(t *testing.T) {
+		reset := stubAuthTokenDependencies(t)
+		defer reset()
+
+		loadCalled := false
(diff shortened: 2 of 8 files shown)


internal/batches/service/remote.go
```diff
@@ -129,31 +129,43 @@ func (svc *Service) UploadBatchSpecWorkspaceFiles(ctx context.Context, workingDi
 }
 
 func getFilePaths(workingDir, filePath string) ([]string, error) {
```
keegancsmith, line 131: any idea why this implementation didn't just use filepath.Walk instead of implementing its own recursion?

internal/batches/executor/run_steps.go
```diff
@@ -610,26 +610,37 @@ func createCidFile(ctx context.Context, tempDir string, repoSlug string) (string
 }
 
 func getAbsoluteMountPath(batchSpecDir string, mountPath string) (string, error) {
+	root, err := os.OpenRoot(batchSpecDir)
```
bahrmichael, line 614: I would appreciate a comment here for future devs to tell them why this should be used.

internal/batches/service/remote.go
cbrnrd: According to [deepsearch](https://sourcegraph.sourcegraph.com/deepsearch/52c90e0b-5cf7-435e-a7c2-c5dda20e364a), it's to preserve the relative path structure from a particular working directory

PR #1292, opened 2026-04-13
feat/cli: migrate version subcommand to use urface/cli

burmudar:
This introduces a compatibility layer to start migrating src-cli to use a cli framework.

Some of the benefits we get by using a cli framework:
- consistent flag handling
- automatic help rendering of commands and flags
- automatic value set of flags based on env variables
- consistent handling of exit codes
- hooks - like to make sure we have loaded the configuration before a command starts execution

If a command is within the migrated set, we then defer to executing the command within urfave/cli with `runMigrated`. If the command is **not** in the set we executing the legacy code with `runLegacy`.

### Test plan
Tested locally


BolajiOlajide (approved):

cmd/src/cmd.go
```diff
@@ -68,44 +78,98 @@ func (c commander) run(flagSet *flag.FlagSet, cmdName, usageText string, args []
...
-				}
-				os.Exit(e.Code())
-			}
+		if isMigratedCmd {
+			exitCode, err = runMigrated(flagSet)
+		} else {
+			exitCode, err = runLegacy(cmd, flagSet)
+		}
+		if err != nil {
 			log.Fatal(err)
```
keegancsmith, line 84: Shouldn't we now log.Println this since we always want to use the returned exitCode?

cmd/src/cmd.go
```diff
@@ -1,15 +1,25 @@
...
+	"sort"
 
+	"github.com/sourcegraph/src-cli/internal/clicompat"
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
+	"github.com/urfave/cli/v3"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
 )
 
+var MigratedCommands = map[string]*cli.Command{
```
keegancsmith: minor this doesn't need to be exported

cmd/src/cmd.go
```diff
@@ -68,44 +78,98 @@ func (c commander) run(flagSet *flag.FlagSet, cmdName, usageText string, args []
...
+		Name:        "src",
+		HideVersion: true,
+		Commands:    commands,
+	})
+}
+
+// runMigrated runs the command within urfave/cli framework
+func runMigrated(flagSet *flag.FlagSet) (int, error) {
+	ctx := context.Background()
+	args := append([]string{"src"}, flagSet.Args()...)
```
keegancsmith: I don't think this works since it won't include any flag arguments rights? This is just everything after command line flags?

I think the only reason it works right now is you have only miugrated version

cmd/src/cmd.go
```diff
@@ -68,44 +78,98 @@ func (c commander) run(flagSet *flag.FlagSet, cmdName, usageText string, args []
...
 		// Read global configuration now.
 		var err error
 		cfg, err = readConfig()
 		if err != nil {
 			log.Fatal("reading config: ", err)
 		}
 
-		// Parse subcommand flags.
-		args := flagSet.Args()[1:]
-		if err := cmd.flagSet.Parse(args); err != nil {
```
keegancsmith, line 85: I'm kinda confused how this all worked before. .Args is the non flag arguments, not the raw argv.

cmd/src/version.go
```diff
@@ -3,63 +3,73 @@ package main
...
-	var (
-		clientOnly = flagSet.Bool("client-only", false, "If true, only the client version will be printed.")
-		apiFlags   = api.NewFlags(flagSet)
-	)
-
-	handler := func(args []string) error {
-		fmt.Printf("Current version: %s\n", version.BuildTag)
-		if clientOnly != nil && *clientOnly {
-			return nil
+var versionCommandv2 = clicompat.Wrap(&cli.Command{
```
keegancsmith: why do you call it v2?

cmd/src/cmd.go
```diff
@@ -68,44 +78,98 @@ func (c commander) run(flagSet *flag.FlagSet, cmdName, usageText string, args []
 
 	// Find the subcommand to execute.
 	name := flagSet.Arg(0)
+
 	for _, cmd := range c {
-		if !cmd.matches(name) {
+		_, isMigratedCmd := MigratedCommands[name]
+		if !isMigratedCmd && !cmd.matches(name) {
```
keegancsmith: this is kinda funky to read and will also break when one day `c` is empty.

you don't do that much inside of this for loop. Why don't you pull out a check for MigratedCommands before this for loop and if you have a match in it, just have a simpler bit of code to run that command (with readConfig/etc duplicated)

keegancsmith (approved):
I'm kinda confused how things worked before for subcommands and flags... but yeah if your testing works (especially trying out cli flags for the subcommand), then lets do it.

Is there a wayt o migrate future commands which don't break them up so much / adjust indentation so much. That would make the PRs much easier to validate.

cmd/src/version.go
burmudar: Mostly just to distinguish it from the normal version command but I realize now it's a misnomer, since there isn't a versionCommand 🤡 

I'll update it!

cmd/src/cmd.go
burmudar: Agreed! I'll move it

keegancsmith:
@burmudar I realised how it works now and came back to comment why. golangs flagset packages will treat all arguments after the first non flag as arguments. So `src version -flag` has `Args = ['version', '-flag']`

cmd/src/cmd.go
keegancsmith: Another reason to move it, I just realised it has this issue with how its currently implemented. Any subcommand can be a global migrated command:

```
❯ go run ./cmd/src auth version
Current version: dev
Recommended version: 7.0.3 or later
```

burmudar: Well now that is awkward 🤣 Good spot! Push the fix in a few

burmudar: Thought I report back on what is happening here :D A migrated command becomes global by "accident", that is because the run we execute here, gets executed again on a child/subcommand again.

For `src auth version`
```
-- first pass ('src', 'auth', 'version') --
> subcommand is `auth`
> auth is not a migrated command and `runLegacy` gets executed
> In `runLegacy`, cmd.run (auth.run) gets called - which is the same run loop we just executed ...

-- second pass ('auth', 'version') --
> subcommand is `version`
> version is a migrated command and `runMigrated` gets executed
```

So for the fix, I will be moving `runMigrated` completely out of `commander.run`

cmd/src/cmd.go
burmudar: fixed

cmd/src/cmd.go
burmudar: correct and fixed!

cmd/src/cmd.go
burmudar: changed this to just work with os.Args

cmd/src/main.go
```diff
@@ -93,7 +93,18 @@ func main() {
...
+	ranMigratedCmd, exitCode, err := maybeRunMigratedCommand()
+	if ranMigratedCmd {
+		if err != nil {
+			log.Println(err)
+		}
+		os.Exit(exitCode)
+	}
+
+	// if we didn't run a migrated command, then lets try running the legacy version
+	if !ranMigratedCmd {
```
keegancsmith, line 105: minor: you don't need this, you always call os.Exit if its true

burmudar: Agreed - fixed in https://github.com/sourcegraph/src-cli/pull/1294

PR #1293, opened 2026-04-15
feat(abc): src abc variables command

bahrmichael:
This PR introduces a `src abc variables` command for setting and deleting variables on workflow instances using the update variables api from Sourcegraph.

```
Syntax:
src abc variables delete <workflow-instance-id> <variable-name>
src abc variables set <workflow-instance-id> <variable-name>=<variable-value>

With multiple arguments and --var:
src abc variables delete <workflow-instance-id> --var <variable-name-1> --var <variable-name-2>
src abc variables set <workflow-instance-id> --var <variable-name-1>=<variable-value-1> --var <variable-name-2>=<variable-value-2>

Examples:
src abc variables delete QWdlbnRpY1dvcmtmbG93SW5zdGFuY2U6MQ== approval
src abc variables set QWdlbnRpY1dvcmtmbG93SW5zdGFuY2U6MQ== prompt="Make some follow up changes"
```

The workflow instance id is an argument after the delete/set command, so that we can follow the existing subcommand patterns without having to introduce support for prefix args.


cmd/src/abc.go
```diff
@@ -0,0 +1,56 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/sourcegraph/src-cli/internal/cmderrors"
+)
+
+var abcCommands commander
+
+func init() {
+	usage := `'src abc' is a tool that manages agentic batch changes on a Sourcegraph instance.
```
bahrmichael:
```suggestion
	usage := `'src abc' is a tool that manages agentic batch changes.
```

cmd/src/abc_variables.go
```diff
@@ -0,0 +1,57 @@
...
+	"flag"
+	"fmt"
+
+	"github.com/sourcegraph/src-cli/internal/cmderrors"
+)
+
+var abcVariablesCommands commander
+
+func init() {
+	usage := `'src abc <workflow-instance-id> variables' is a tool that manages workflow instance variables on a Sourcegraph instance.
```
bahrmichael:
```suggestion
	usage := `'src abc <workflow-instance-id> variables' is a tool that manages workflow instance variables.
```

cmd/src/abc_variables.go
```diff
@@ -0,0 +1,57 @@
...
+
+Usage:
+
+	src abc <workflow-instance-id> variables command [command options]
+
+The commands are:
+
+	set	set workflow instance variables
+	delete	delete a workflow instance variable
```
bahrmichael:
```suggestion
	set	    set workflow instance variables
	delete	delete workflow instance variables
```

cmd/src/abc_variables_delete.go
```diff
@@ -0,0 +1,89 @@
...
+		varArgs := abcVariableArgs(cmd.StringSlice("var"))
+
+		if len(varArgs) == 0 {
+			return cmderrors.Usage("must provide at least one variable name")
+		}
+		return runABCVariablesDelete(ctx, client, instanceID, cmd.Args().Tail(), varArgs, cmd.Writer, cmd.Bool("get-curl"))
+	},
+})
+
+func parseABCVariableNames(positional []string, flagged abcVariableArgs) ([]string, error) {
```
burmudar: think we can collapse this into `runABCVariablesDelete`

cmd/src/abc_variables_delete.go
```diff
@@ -0,0 +1,89 @@
...
+	variableNames = append(variableNames, flagged...)
+
+	if slices.Contains(variableNames, "") {
+		return nil, cmderrors.Usage("variable names must not be empty")
+	}
+
+	return variableNames, nil
+}
+
+func runABCVariablesDelete(ctx context.Context, client api.Client, instanceID string, positional []string, flagged abcVariableArgs, output io.Writer, getCurl bool) error {
```
burmudar: `positional` and `flagged` can be collapsed into one array imo

cmd/src/abc_variables_set.go
```diff
@@ -0,0 +1,181 @@
...
+
+type abcVariableArgs []string
+
+func (a *abcVariableArgs) String() string {
+	return strings.Join(*a, ",")
+}
+
+func (a *abcVariableArgs) Set(value string) error {
+	*a = append(*a, value)
+	return nil
```
burmudar: never return an error  so we might as well remove it from the function signature

cmd/src/abc_variables_set.go
```diff
@@ -0,0 +1,181 @@
...
+		return abcVariable{}, err
+	}
+	if remove {
+		return abcVariable{}, cmderrors.Usagef("invalid variable assignment %q: use 'src abc variables delete <workflow-instance-id> %s' to remove a variable", raw, name)
+	}
+
+	return abcVariable{Key: name, Value: value}, nil
+}
+
+func runABCVariablesSet(ctx context.Context, client api.Client, instanceID string, positional []string, flagged abcVariableArgs, output io.Writer, getCurl bool) error {
```
burmudar: I think positional + flagged should be combined before this method is called

cmd/src/abc_variables_set.go
```diff
@@ -0,0 +1,181 @@
...
+}
+
+func parseABCVariables(positional []string, flagged abcVariableArgs) ([]abcVariable, error) {
+	rawVariables := append([]string{}, positional...)
+	rawVariables = append(rawVariables, flagged...)
+	if len(rawVariables) == 0 {
+		return nil, cmderrors.Usage("must provide at least one variable assignment")
+	}
+
+	variables := make([]abcVariable, 0, len(rawVariables))
```
burmudar: I _think_ you can use `strings.FieldsFunc`

cmd/src/abc_variables_set.go
```diff
@@ -0,0 +1,181 @@
...
+			"key":   variable.Key,
+			"value": variable.Value,
+		})
+	}
+
+	if err := updateABCWorkflowInstanceVariables(ctx, client, instanceID, graphqlVariables); err != nil {
+		return err
+	}
+
+	if getCurl {
```
burmudar: I think if you're not going to support this flag then we should:
* not pass it in
* print a warning before running this that `get-curl` is currently unsupported for this command

cmd/src/abc_variables_delete.go
```diff
@@ -0,0 +1,89 @@
...
+			"value": "null",
+		})
+	}
+
+	if err := updateABCWorkflowInstanceVariables(ctx, client, instanceID, variables); err != nil {
+		return err
+	}
+
+	if getCurl {
+		return nil
```
burmudar: see my later comment on get-curl

burmudar (approved):

cmd/src/abc_variables_set.go
bahrmichael: If I understand it correctly urfave should already give us individual args that don't need to be split by whitespace. For each arg (name=value) the `strings.Cut` method looks like a better fit to me.

bahrmichael:
List of notable changes since review:

- Added help message on UsageError (I just got "exit code 2" if I forgot the instance id; we can isolate this to just the new commands, but I thought it's helpful to have this in general)
- Implemented proper support for `-get-cli` (tested locally, If I understand correctly this flag is evaluated by the api client, and then we're supposed to just do a noop if ok=false)
- Coalesced args and flag var args and simplified the methods
- Added DisableSliceFlagSeparator so that lists won't be split into multiple args

burmudar (approved):

PR #1297, opened 2026-04-16
Wb/migrated abc

burmudar:
Migrates abc command away from legacy commander pattern to rather use `urfave/cli`

Dependencies:
- https://github.com/sourcegraph/src-cli/pull/1296

### Test plan
`src abc variables -h`
`src abc variables set-h`
`src abc variables delete -h`


bahrmichael (approved):
LGTM!

PR #1302, opened 2026-04-28
chore: update mise golang to 1.26.2

burmudar:
see title


keegancsmith (approved):
your PR contains the docgen change as well

lib/docgen/default.go
```diff
@@ -0,0 +1,19 @@
+package docgen
+
```
BolajiOlajide: Seems this branch carries over some of the changes from the previous branch alongside it.

BolajiOlajide (approved):
approving to unblock!

lib/docgen/default.go
burmudar: Fixed! thanks

PR #1304, opened 2026-04-28
feat/docs: generate docs for legacy and urfave commands

burmudar:
Update docgen to render migrated `urface/cli` commands in the expected help doc template

Closes CPL-381

> [!NOTE]
> There are some formatting issues with Commands and Subcommands which have no flags. I plan to address this in a follow up

### Test plan
`src doc -o doctest/`

`abc/index.md`
```
# `src abc`

manages agentic batch changes.



## Usage

sh
$ src abc [command options]


## Subcommands

* [`variables`](variables/index.md)

```

`abc/variables/index.md`
```
# `src abc variables`

manage workflow instance variables.



## Usage

sh
$ src abc variables [command options]


## Subcommands

* [`delete`](delete.md)
* [`set`](set.md)

```

---
`abc/variables/delete.md`
```
# `src abc variables delete`

Delete variables on a workflow instance.

Delete workflow instance variables

Examples:

  Delete a variable from a workflow instance:

	    $ src abc variables delete QWdlbnRpY1dvcmtmbG93SW5zdGFuY2U6MQ== approval

  Delete multiple variables in one request:

	    $ src abc variables delete QWdlbnRpY1dvcmtmbG93SW5zdGFuY2U6MQ== --var approval --var checkpoints

## Usage

sh
$ src abc variables delete [options] <workflow-instance-id> [<name> ...]


## Flags

| Name | Description | Default Value |
|------|-------------|---------------|
| `---dump-requests` | Log GraphQL requests and responses to stdout | `false`|
| `---get-curl` | Print the curl command for executing this query and exit (WARNING: includes printing your access token!) | `false`|
| `---insecure-skip-verify` | Skip validation of TLS certificates against trusted chains | `false`|
| `---trace` | Log the trace ID for requests. See https://docs.sourcegraph.com/admin/observability/tracing | `false`|
| `---user-agent-telemetry` | Include the operating system and architecture in the User-Agent sent with requests to Sourcegraph | `true`|
| `---var` | Variable name to delete. Repeat for multiple names. | ``|
```
---


cmd/src/doc.go
```diff
@@ -151,6 +202,29 @@ func openDocFile(base, fqcn string) (interface {
 	return os.Create(cmdPath + ".md")
 }
 
+func migratedVisibleCommands(commands []*cli.Command) []*cli.Command {
```
keegancsmith: this feedback from amp seems legit. Maybe those should be exported from lib/docgen?

> The functions `migratedVisibleCommands` and `migratedDocLink` duplicate logic that already exists in `lib/docgen/markdown.go` (`visibleCommands`, `hasVisibleCommands`, and `docPath`).

lib/docgen/markdown.go
```diff
@@ -2,163 +2,186 @@ package docgen
...
-
+		var name string
 		for _, s := range f.Names() {
 			trimmed := strings.TrimSpace(s)
-			if len(modifiedArg) > len(opener) {
-				modifiedArg += sep
-			}
 			if len(trimmed) > 1 {
-				modifiedArg += fmt.Sprintf("--%s", trimmed)
+				name += fmt.Sprintf("--%s", trimmed)
```
keegancsmith: What happens if you have multiple aliases for a flag? eg if you have names `["output", "o"]` won't you end up with a name like `--output-o`?

lib/docgen/markdown.go
```diff
@@ -174,20 +197,64 @@ func prepareUsage(command *cli.Command) string {
...
-{{ else }}
-{{ .App.Name }} [GLOBAL FLAGS] command [COMMAND FLAGS] [ARGUMENTS...]
-{{ end }}` + "```" + `
-{{ if .GlobalArgs }}
-Global flags:
+| Name | Description | Default Value |
+|------|-------------|---------------|
+{{- range .Flags -}}
+{{- "\n" -}}
+| ` + "`" + `-{{ .Name }}` + "`" + ` | {{ .Desc }} | ` + "`" + `{{ .Default }}` + "`" + `|
```
keegancsmith: your name logic I commented on above already include the `--` prefix.

keegancsmith (approved):

lib/docgen/markdown.go
burmudar: Ah Ty. Will fix

burmudar: Fixed

cmd/src/doc.go
burmudar: Yup - moved it to lib/docgen

lib/docgen/markdown.go
burmudar: Added a strings.Join, so now we'll get `--output, -o`

PR #1315, opened 2026-05-06
fix: dedupe migrated root commands

burmudar:
This changes `migratedRootCommand` to collect commands by their canonical `cli.Command.Name` before sorting them for the root command. Alias dispatch entries remain supported in `migratedCommands`, while the generated root command only includes each command once.

## Test Plan

- `go test ./cmd/src`

### DIFF
--- a/cmd/src/run_migration_compat.go
+++ b/cmd/src/run_migration_compat.go
@@ -1,12 +1,14 @@
 package main
 
 import (
+	"cmp"
 	"context"
 	"flag"
 	"fmt"
 	"log"
+	"maps"
 	"os"
-	"sort"
+	"slices"
 
 	"github.com/sourcegraph/src-cli/internal/clicompat"
 	"github.com/sourcegraph/src-cli/internal/cmderrors"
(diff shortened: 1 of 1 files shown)


cmd/src/run_migration_compat.go
```diff
@@ -43,16 +45,13 @@ func maybeRunMigratedCommand() (isMigrated bool, exitCode int, err error) {
...
-		names = append(names, name)
-	}
-	sort.Strings(names)
-
-	commands := make([]*cli.Command, 0, len(names))
-	for _, name := range names {
-		commands = append(commands, migratedCommands[name])
+	uniqueCommands := make(map[string]*cli.Command, len(migratedCommands))
+	for _, cmd := range migratedCommands {
+		uniqueCommands[cmd.Name] = cmd
```
keegancsmith, line 53: do we have commands with duplicated names? IE do we need to add a unique step?

keegancsmith (approved):

cmd/src/run_migration_compat.go
burmudar: yeah in the subsequent migration of `codeowners` and `orgs` we register the command twice in the set to support their current aliases.

PR #1317, opened 2026-05-08
chore/cli: migrate users to urfave/cli

burmudar:
The command migration is moving legacy commander commands onto urfave/cli so they share the same help, flag, and dispatch behavior as newly migrated commands.

Migrate `src users` and its `list`, `get`, `create`, `delete`, `prune`, and `tag` subcommands to `cli.Command` definitions. Register both `users` and `user` in the migrated command map, and remove `users` from the legacy docs path so generated docs come from the migrated root command.

## Test Plan

- `go test ./cmd/src`
- `go test ./...`
- `go run ./cmd/src users -h`
- `go run ./cmd/src user -h`
- `go run ./cmd/src users list -h`


keegancsmith (approved):
big PR that looks super mechanical. So trusting ya tested everything still works :D

Very minor but I ran `./dev/smoke.sh` and it worked which is good :)

PR #1320, opened 2026-05-11
chore/release: add release skill

burmudar:
Release work has a few repository-specific conventions that are easy to miss when cutting patch, minor, or major releases from the CLI. This adds a project skill so future release tasks infer deterministic release details from tags, branches, repository conventions, and the user's wording instead of asking for information the agent can safely derive.

The skill centers on deriving a release plan up front: release type, version, target release branch, source, tag format, push mode, and worktree strategy. It centralizes hard stops for unsafe operations, keeps pushes local-only unless explicitly approved, and uses a concise execution flow for patch, minor, and major releases. Patch guidance relies on the normal meaning of a patch release instead of an exhaustive include/exclude checklist, while still calling out the less obvious need to include the full remediation closure for vulnerability-driven releases.

Before tagging, the skill now requires a short release contents summary grouped into included and excluded PRs, then asks `Proceed?` so the user can confirm the final contents before the local tag is created.

## Test Plan

Used the first version of this skill to cut 7.1.3 / 7.2.2. Since then we have reduced the redundancy of the skill and it's overfitting due to cutting a patch release

### DIFF
--- a/.agents/skills/release/SKILL.md
+++ b/.agents/skills/release/SKILL.md
@@ -1,0 +1,128 @@
+---
+name: release
+description: Guides Sourcegraph CLI patch, minor, and major releases. Use when preparing release branches, selecting patch commits, running Trivy checks, creating release tags, or managing local/pushed release artifacts.
+---
+
+# Release
+
+## Infer the release plan
+
+Gather only the information that is actually missing. Do not ask questions whose answers can be derived safely from tags, release branches, repository conventions, or the user's wording.
+
+Before asking the user anything, derive the release plan:
+
+- `release_type`: from user wording or semver shape.
+- `version`: from the explicit user version, otherwise from existing tags.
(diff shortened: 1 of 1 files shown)


keegancsmith (approved):
> Used the first version of this skill to cut 7.1.3 / 7.2.2. Since then we have reduced the redundancy of the skill and it's overfitting due to cutting a patch release

Neat. Is it worth sharing the transcript of that conversation? Or better to share the next time you use it?

burmudar:
> > Used the first version of this skill to cut 7.1.3 / 7.2.2. Since then we have reduced the redundancy of the skill and it's overfitting due to cutting a patch release
> 
> Neat. Is it worth sharing the transcript of that conversation? Or better to share the next time you use it?

Definitely the next one. The current one is filled with me iterating on it and reducing the agent overfitting

PR #1322, opened 2026-05-19
batches: upload large exec artifacts

burmudar:
Batch executor jobs can produce stdout, stderr, and diffs large enough to bloat the final JSONL result stream. Upload oversized step artifacts from the existing `src batch exec` execution path instead of adding a separate command or user-facing auth flags.

`src batch exec` now discovers the executor job context from the environment, uses the existing executor job bearer token plus executor name auth model to stream large artifacts to the Batch Changes artifacts endpoint, and records returned references on `AfterStepResult`. Small outputs stay inline for compatibility, and per-step progress output is suppressed when artifact uploads are enabled so large logs are not duplicated into task-step JSONL events.

companian branch in sg: wb/proto-exec-upload

Test Plan:

- go test ./cmd/src ./internal/batches/executor ./internal/batches/ui github.com/sourcegraph/sourcegraph/lib/batches/execution

### DIFF
--- a/cmd/src/batch_exec_artifacts.go
+++ b/cmd/src/batch_exec_artifacts.go
@@ -1,0 +1,108 @@
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strconv"
+	"strings"
+
+	"github.com/sourcegraph/src-cli/internal/batches/executor"
+
+	"github.com/sourcegraph/sourcegraph/lib/batches/execution"
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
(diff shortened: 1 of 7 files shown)


cmd/src/batch_exec_artifacts.go
```diff
@@ -0,0 +1,108 @@
...
+	req.Header.Set("X-Sourcegraph-Executor-Name", u.executorName)
+
+	resp, err := u.client.Do(req)
+	if err != nil {
+		return execution.ArtifactReference{}, err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		body, _ := io.ReadAll(io.LimitReader(resp.Body, 4096))
```
burmudar, line 82: dangerous to load this into memory

PR #1323, opened 2026-05-20
feat/doc: do not render any progress in non-TTY environments

burmudar:
* We no longer use the progress bar during doc generation.
* We only print _anything_ if the verbose flag is given `-v`

### Test plan
```
$ ./src doc -o doc-test/
$ echo $?
0
```
with verbose flag
```
go run ./cmd/src -v doc -o doc-test/
<snip>
writing: orgs/list.md...
writing: orgs/members/index.md...
writing: orgs/members/add.md...
writing: orgs/members/remove.md...
writing: users/index.md...
writing: users/create.md...
writing: users/delete.md...
vusers/get.md...
writing: users/list.md...
writing: users/prune.md...
writing: users/tag.md...
writing: version.md...
✅ 62 files rendered under doc-test/
```

### DIFF
--- a/cmd/src/doc.go
+++ b/cmd/src/doc.go
@@ -65,12 +65,8 @@ Examples:
 			"repos":      &reposCommands,
 		}
 
-		pending := out.Pending(output.Line("", output.StylePending, "Rendering Markdown..."))
-		count := 0
 		rootSubcommands := map[string]string{}
-		defer func() {
-			pending.Complete(output.Linef(output.EmojiSuccess, output.StyleSuccess, "%d files rendered under %s", count, *outputFlag))
-		}()
+		count := 0
 		for groupName, cmdr := range commanders {
 			subcommands := map[string]string{}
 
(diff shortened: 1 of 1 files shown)


lib/output/pending.go
```diff
@@ -18,9 +18,12 @@ type Pending interface {
 }
 
 func newPending(message FancyLine, o *Output) Pending {
```
keegancsmith, line 20: is it worth forking our behaviour from lib/output in the monorepo?

keegancsmith (approved):

keegancsmith:
I'm not exactly sure if progress is the right thing to be using here? Isn't this command super fast? IE it should have pretty much no output, unless -v is specified. WDYT?

burmudar:
> I'm not exactly sure if progress is the right thing to be using here? Isn't this command super fast? IE it should have pretty much no output, unless -v is specified. WDYT?

yeah I agree. I think that is the better fix here. We can then stay consistent with the monorepo too.

lib/output/pending.go
burmudar: removed this in the latest commit where we also just remove the use  of the progress bar during doc generation

keegancsmith (commented):
LGTM!

PR #1325, opened 2026-05-27
feat/batch-changes: support codingAgent steps via src-cli

taras-yemets:
v1 server-side batch changes run via `src-cli`. v3 specs can declare `codingAgent` steps, so src-cli needs to know how to run them.

This PR adds that support. It pairs with sourcegraph/sourcegraph#12503, which adds the server side for v1 (v2 is rejected at admission).

## Architecture

- **codingAgent rendering.** `lib/batches/codingagent` turns a `codingAgent` step into a shell script (`lib/batches/codex` provides codex's install + invocation). The executor calls this renderer instead of treating `step.run` as a template, and writes the result into the run container as-is.
- **Auth env forwarding.** The server sets `SRC_BATCHES_MODEL_PROVIDER_TOKEN` and `SRC_BATCHES_JOB_ID` on the v1 `CliStep`. src-cli forwards them into the user container so the agent CLI can authenticate to `/.executors/model-provider/batches`. The token is redacted from src-cli's UI and log output; the server's `RedactedValues` is a backstop.
- **Soft-forked `lib/`.** `src-cli/lib/` is a soft-fork of `sourcegraph/sourcegraph`'s `lib/batches`, wired via `replace github.com/sourcegraph/sourcegraph/lib => ./lib` in `go.mod`. The new `codingagent` / `codex` packages and the changes to `batch_spec.go` / `batch_spec_stringdata.go` mirror sourcegraph#12503. src-cli's copy is what runs in the user container, so the duplication is on purpose, not drift. `batch_spec_stringdata.go` was regenerated from the canonical schema.
- **Shared wire-protocol constants.** `SRC_BATCHES_MODEL_PROVIDER_TOKEN`, `SRC_BATCHES_JOB_ID`, and `X-Sourcegraph-Job-ID` are defined in both this repo and sourcegraph#12503. Sharing them would mean either untangling the `lib/` fork or extracting a new shared module — both much more work than keeping three short strings in sync. Both sides have comments pointing at the mirror.

## Test Plan

Verified end-to-end together with sourcegraph#12503; see the test plan and demo there.


lib/batches/codex/codex.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/kballard/go-shellquote"
+
+	batcheslib "github.com/sourcegraph/sourcegraph/lib/batches"
+	"github.com/sourcegraph/sourcegraph/lib/batches/codingagent/types"
+)
+
+const (
+	model         = "gpt-5.4"
+	pinnedVersion = "0.134.0"
```
burmudar, line 15: I think these should be injected via WorkspaceInput.json

lib/batches/codex/codex.go
```diff
@@ -0,0 +1,67 @@
...
+
+case "$(uname -m)" in
+  x86_64)  _triple=x86_64-unknown-linux-musl ;;
+  aarch64) _triple=aarch64-unknown-linux-musl ;;
+  *) echo "codingAgent codex: unsupported architecture $(uname -m)" >&2; exit 1 ;;
+esac
+
+# Stage in a temp dir and mv into place so a failed retry can't leave a half-written binary behind.
+_url="https://github.com/openai/codex/releases/download/rust-v${_version}/codex-${_triple}.tar.gz"
+_tmp=$(mktemp -d "${TMPDIR:-/tmp}/sg-codex.XXXXXX")
```
burmudar, line 35: nit: `src-codex`

lib/batches/codex/codex.go
```diff
@@ -0,0 +1,67 @@
...
+
+func (Agent) Type() batcheslib.CodingAgentType { return batcheslib.CodingAgentTypeCodex }
+func (Agent) ImageRequirements() []string      { return []string{"curl", "tar"} }
+
+func (Agent) InstallScript() string {
+	return fmt.Sprintf(`_install_dir=%s
+_version=%s
+
+case "$(uname -m)" in
+  x86_64)  _triple=x86_64-unknown-linux-musl ;;
```
burmudar, line 28: since `src` is the current runner, we know the architecture so I think this entire script can be made a lot simpler and we can fail a lot sooner than to fail inside the bash script.

I think a lot of what this script is doing can actually be done natively by `src` using go.

lib/batches/codingagent/codingagent.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+	var renderedPrompt bytes.Buffer
+	if err := template.RenderStepTemplate("codingagent-prompt", agentStep.Prompt, &renderedPrompt, stepCtx); err != nil {
+		return "", errors.Wrap(err, "rendering codingAgent.prompt")
+	}
+	prefixed := strings.TrimRight(modelProviderURL, "/") + "/" + string(a.Type())
+
+	var b strings.Builder
+	for _, binary := range a.ImageRequirements() {
+		b.WriteString(failIfMissing(a.Type(), binary))
```
burmudar, line 35: we can just use `exec.LookPath`

lib/batches/codingagent/codingagent.go
```diff
@@ -0,0 +1,64 @@
...
+	)
+}
+
+var agents = func() map[batcheslib.CodingAgentType]types.Agent {
+	out := map[batcheslib.CodingAgentType]types.Agent{}
+	for _, a := range []types.Agent{
+		codex.Agent{},
+	} {
+		if _, exists := out[a.Type()]; exists {
+			panic("duplicate codingagent agent for " + a.Type())
```
burmudar, line 59: this is a programming error if there are double? we should also not panic

lib/batches/workspaces_execution_input.go
```diff
@@ -21,6 +21,9 @@ type WorkspacesExecutionInput struct {
 	CachedStepResult execution.AfterStepResult `json:"cachedStepResult"`
 	// SkippedSteps determines which steps are skipped in the execution.
 	SkippedSteps map[int]struct{} `json:"skippedSteps"`
+	// ModelProviderURL is the resolved proxy base URL for coding-agent
+	// steps; only set when the spec contains at least one codingAgent step.
+	ModelProviderURL string `json:"modelProviderURL,omitempty"`
```
burmudar, line 26: I wonder if we should make `WorkspaceInputv3` ? Not for here but this _is_ intended for use with Batches v3

taras-yemets:
Converting to draft for the time being: [Slack](https://sourcegraph.slack.com/archives/C0A7PV1600P/p1780063424377939?thread_ts=1779974813.881439&channel=C0A7PV1600P&message_ts=1780063424.377939)

PR #1326, opened 2026-05-29
fix/batch-changes: ignore per-job env vars in cache key

taras-yemets:
With sourcegraph/sourcegraph#12503, Sourcegraph desugars v3 `codingAgent` steps server-side and declares per-job env-var forwards on `step.Env`. Their resolved values change every dequeue, so including them in cache-key inputs would invalidate the cache on every run.

Strip them before hashing:

- `SRC_EXECUTOR_JOB_TOKEN` — per-job bearer token
- `SRC_EXECUTOR_JOB_ID` — numeric job ID
- `SRC_EXECUTOR_NAME` — dequeueing executor's hostname

The bare-string declarations still influence the key via the serialized `step.Env`, so spec-change invalidation still works. The list is mirrored in sourcegraph/sourcegraph; a comment on each side points at the other.

End-to-end test plan: see sourcegraph/sourcegraph#12503.

### DIFF
--- a/lib/batches/execution/cache/cache.go
+++ b/lib/batches/execution/cache/cache.go
@@ -47,6 +47,13 @@ func (key CacheKey) mountsMetadata() ([]MountMetadata, error) {
 	return nil, nil
 }
 
+// perJobEnvVars resolve to per-job values and must be stripped from
+// cache keys. Mirrored in sourcegraph/sourcegraph/lib/batches/execution/cache/cache.go.
+var perJobEnvVars = []string{
+	"SRC_BATCHES_MODEL_PROVIDER_TOKEN",
+	"SRC_BATCHES_JOB_ID",
+}
+
 // resolveStepsEnvironment returns a slice of environments for each of the steps,
 // containing only the env vars that are actually used.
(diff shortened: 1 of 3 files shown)


lib/batches/execution/cache/cache.go
```diff
@@ -47,6 +47,15 @@ func (key CacheKey) mountsMetadata() ([]MountMetadata, error) {
 	return nil, nil
 }
 
+// perRunEnvVars resolve to per-job or per-executor values that change on
+// every dequeue and must be stripped from cache keys. Mirrored in
+// sourcegraph/sourcegraph/lib/batches/execution/cache/cache.go.
+var perRunEnvVars = []string{
```
bobheadxi, line 53: Thank you for docstring!

bobheadxi (approved):

bobheadxi:
Linking for ref: could also tidy up https://github.com/sourcegraph/sourcegraph/pull/12640 to resolve

bobheadxi:
@taras-yemets how do we pull this into the src-cli version used by executor? Do we need a release?

taras-yemets:
> @taras-yemets how do we pull this into the src-cli version used by executor? Do we need a release?

I think we need a release first, then set the minimum expected src-cli version in the Sourcegraph repo.

PR #1330, opened 2026-06-01
restore lib/go.mod

keegancsmith:
The vendor-lib script needs updating to rather preserve it. But for now unblock the breakage we will get by not using our version of the yaml lib.


bahrmichael (approved):

bahrmichael:
Thank you! :)

burmudar (approved):

PR #1338, opened 2026-06-30
feat: warn when src-cli is older than the instance's recommended version

### LINKED ISSUE #930: Warn on any command if not on latest version
Many users encounter bugs from outdated versions of src-cli and raise them to support before realizing they were already fixed and just needed to update. Recently, a user hit a nasty bug when they were unintentionally running a version 16 months out of date! We should try to warn users if the version of src-cli they are running is not at or above the version recommended by their Sourcegraph instance (`/.api/src-cli/version`). We already perform this check when running the `src version`:

```
> src version
Current version: 4.3.0
Recommended version: 4.4.0 or later
```

We should perform this check before running any command and highlight a warning in the output if it's out-of-date.

Montana:
Before running any command, `src` now does a best-effort check of the running version against the version recommended by the configured Sourcegraph instance (the existing `/.api/src-cli/version` endpoint) and prints a single warning to `stderr` when it is behind.

The check is **fail-open**:

- Missing config, network errors, an unreachable instance, or an unparseable
  version all result in no warning.
- A 3s timeout ensures it never noticeably delays the command.
- It writes to **stderr**, so `--json` output is unaffected.
- It skips `version`/`help`/no-arg invocations, dev builds, and any run with
  `SRC_SKIP_VERSION_CHECK` set.
- Comparison is limited to major/minor/patch, so prereleases do not produce
  spurious warnings.

Closes #930

### DIFF
--- a/cmd/src/version_check.go
+++ b/cmd/src/version_check.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/Masterminds/semver"
+
+	"github.com/sourcegraph/src-cli/internal/api"
+	"github.com/sourcegraph/src-cli/internal/version"
+)
+
+// envSkipVersionCheck disables the automatic out-of-date warning when set to a
+// non-empty value.
+const envSkipVersionCheck = "SRC_SKIP_VERSION_CHECK"
+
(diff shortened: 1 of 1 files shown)


burmudar:
@Montana thanks for your contribution!

I think there are some missing parts here. None of the methods added here are used anywhere?

burmudar:
@Montana as mentioned in your other PR. Please add your tests and functionality to this PR. 

I will close the other two PRs.

PR #1342, opened 2026-07-01
Add version check functionality to src-cli

Montana:
Added some things per @burmudar's comment.


burmudar:
@Montana please follow the guidance of adding it to your original PR

PR #1351, opened 2026-07-07
fix/snapshot: prevent OS command injection via targets file

cbrnrd:
# Problem

`src snapshot databases/restore --run` built shell command strings by interpolating untrusted YAML fields (target, dbname, username, password) and executed them via `exec.Command("bash", "-c", c)`, allowing shell-metacharacter injection. This could allow a malicious YAML author to run commands on a user's machine.

# Solution

Don't use `bash -c` anymore!

Reworked `internal/pgdump/pgdump.go` to build commands as argv arrays instead of shell strings:

- Added a `Command`  struct (Env, Args, InputFile, OutputFile) with a `Run()` that executes directly via `exec.Command(args[0], args[1:]...)` and does file redirection in Go (ie no `bash -c`).
- `DumpCommand`/`RestoreCommand` now return `(args, env []string)`, so each untrusted value is a discrete argv element (e.g. `--host=<target>`, `--dbname=<value>`, `PGPASSWORD` in env).
- For docker/kubectl builders, the target is a separate argv element and the nested `sh -c/bash -c` inner command is safely shell-quoted (`shellQuote`/`shellQuoteEnv`), preventing injection even inside the container.

### Test plan
CI, net-new unit tests

### DIFF
--- a/cmd/src/snapshot_databases.go
+++ b/cmd/src/snapshot_databases.go
@@ -4,7 +4,6 @@ import (
 	"flag"
 	"fmt"
 	"os"
-	"os/exec"
 	"strings"
 
 	"github.com/sourcegraph/sourcegraph/lib/errors"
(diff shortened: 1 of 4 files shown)


andreeleuterio (approved):
let's get some eyes from @jac (or someone he points to)

jac (approved):
I haven't looked at this in years but it looks good to me 👍🏻

PR #1352, opened 2026-07-07
fix/batch-changes: validate files target paths in changeset hooks and executor

cbrnrd:
## Problem

`src-cli` rejects commas in top-level `steps.files` target paths because a comma breaks out of Docker `--mount` syntax and lets an attacker inject arbitrary `source=`/`target=` pairs (e.g. mounting `/var/run/docker.sock`). That same validation was missing for `changesetHooks.*.steps[].files`, so a malicious v3 batch spec could smuggle a dangerous `files` target path through a hook and reach the executor-side Docker mount sink.

The executor (`src batch exec`) makes this worse: it loads already-parsed `Step` structs from JSON and never re-runs `ParseBatchSpec`/`validateHooks`, so parse-time validation alone is a coordinator-only guard.

## Solution

- Add the missing comma check for `files` target paths in `validateHooks` (covers both `onCIFailure` and `onMergeConflict`).
- Add defense-in-depth validation at the executor sink in `createFilesToMount`, so the check holds regardless of how a step reaches the executor.

## Verification Evidence

```
$ go test ./internal/batches/service/ -run ParseBatchSpec -v
--- PASS: TestService_ParseBatchSpec/hook_files_target_path_with_comma

$ go test ./internal/batches/executor/ -run CreateFilesToMount -v
--- PASS: TestCreateFilesToMount_RejectsCommaInTargetPath

$ go test ./internal/batches/...
ok  	github.com/sourcegraph/src-cli/internal/batches/executor	2.402s
ok  	github.com/sourcegraph/src-cli/internal/batches/service	0.626s
(all batch packages pass)
```

### DIFF
--- a/internal/batches/executor/run_steps.go
+++ b/internal/batches/executor/run_steps.go
@@ -540,6 +540,16 @@ func createFilesToMount(tempDir string, step batcheslib.Step, stepContext *templ
 	// can mount them into the container.
 	filesToMount := make(map[string]*os.File, len(files))
 	for name, content := range files {
+		// Defense-in-depth: the mount target name is concatenated directly
+		// into a Docker `--mount type=bind,...,target=<name>` argument. A comma
+		// in the name would let an attacker break out of the mount spec and
+		// inject arbitrary source=/target= pairs (e.g. /var/run/docker.sock).
+		// Parse-time validation rejects this, but the executor receives
+		// pre-parsed steps via JSON and never re-runs that validation, so we
+		// re-check here at the sink.
+		if strings.Contains(name, ",") {
+			return nil, cleanup, errors.Newf("files target path %q contains invalid characters", name)
+		}
 		fp, err := os.CreateTemp(tempDir, "")
 		if err != nil {
 			return nil, cleanup, errors.Wrap(err, "creating temporary file")
--- a/lib/batches/batch_spec.go
+++ b/lib/batches/batch_spec.go
@@ -299,13 +299,6 @@ func validateHooks(spec *BatchSpec) error {
 					)))
 				}
 			}
-			for name := range step.Files {
-				if strings.Contains(name, invalidMountCharacters) {
-					errs = errors.Append(errs, NewValidationError(errors.Newf(
-						"hooks.%s step %d files target path contains invalid characters", event, i+1,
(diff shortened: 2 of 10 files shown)


lib/batches/batch_spec.go
```diff
@@ -299,6 +299,13 @@ func validateHooks(spec *BatchSpec) error {
 					)))
 				}
 			}
+			for name := range step.Files {
```
keegancsmith: FYI the code here is vendored from our monorepo.

internal/batches/executor/run_steps.go
```diff
@@ -540,6 +540,16 @@ func createFilesToMount(tempDir string, step batcheslib.Step, stepContext *templ
...
+		// Defense-in-depth: the mount target name is concatenated directly
+		// into a Docker `--mount type=bind,...,target=<name>` argument. A comma
+		// in the name would let an attacker break out of the mount spec and
+		// inject arbitrary source=/target= pairs (e.g. /var/run/docker.sock).
+		// Parse-time validation rejects this, but the executor receives
+		// pre-parsed steps via JSON and never re-runs that validation, so we
+		// re-check here at the sink.
+		if strings.Contains(name, ",") {
+			return nil, cleanup, errors.Newf("files target path %q contains invalid characters", name)
+		}
```
keegancsmith, line 552: ok so this would of failed which is what we want, but now we fail earlier? And that is better from both a UX perspective + extra defense incase the other check for some reason doesn't run?

keegancsmith (approved):

lib/batches/batch_spec.go
cbrnrd: gah right, my [previous attempts](https://github.com/sourcegraph/src-cli/pull/1332) to re-vendor this did not go so well, but I will split this off into a separate PR.

internal/batches/executor/run_steps.go
cbrnrd: correct!

lib/batches/batch_spec.go
cbrnrd: https://github.com/sourcegraph/sourcegraph/pull/13942

PR #1360, opened 2026-08-08
fix(search-jobs): add timeout to authenticated HTTP requests

bunlongheng:
fetchJobLogs and fetchJobResults both issue authenticated GETs via http.DefaultClient, which has no timeout. A slow or unresponsive Sourcegraph instance will stall the CLI indefinitely while the connection stays open with the caller's access token attached. Replaced the bare DefaultClient with a local http.Client{Timeout: 60 * time.Second} in both callers.

### DIFF
--- a/cmd/src/search_jobs_logs.go
+++ b/cmd/src/search_jobs_logs.go
@@ -6,6 +6,7 @@ import (
 	"io"
 	"net/http"
 	"os"
+	"time"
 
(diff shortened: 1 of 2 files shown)


bahrmichael:
Thank you for the suggestion @bunlongheng! We will probably move ahead with a different implementation from https://github.com/sourcegraph/src-cli/pull/1362.

PR #1371, opened 2026-09-08
fix/batches: prevent Git commondir config bypass

cbrnrd:
## Problem

A batch container can create a `.git/commondir` file that points Git to an attacker-controlled directory. The existing metadata restoration only restores `.git/config` and `.git/config.worktree`, so host-side Git commands can still read malicious configuration from the redirected directory. A malicious clean filter can then run code on the host outside the container.

This fixes [VULN-153](https://linear.app/sourcegraph/issue/VULN-153/gitcommondir-bypasses-git-config-restoration-and-achieves-host-rce).

## Solution

Snapshot `.git/commondir` with the other trusted Git control files and restore it before every host-side Git command. If the original repository did not have a `commondir` file, an injected file is removed before Git runs.

The regression test creates an attacker-controlled common directory with a malicious clean filter and verifies that diff generation uses the trusted Git configuration instead.

## Verification Evidence

- `go test ./internal/batches/workspace -count=1`
- `go test ./internal/batches/...`
- `git diff --check`

### DIFF
--- a/internal/batches/workspace/bind_workspace_test.go
+++ b/internal/batches/workspace/bind_workspace_test.go
@@ -195,7 +195,8 @@ func TestDockerBindWorkspace_DiffRestoresTrustedGitConfig(t *testing.T) {
 	}
 
 	dir := *workspace.WorkDir()
-	configPath := filepath.Join(dir, ".git", "config")
+	dotGit := filepath.Join(dir, ".git")
+	configPath := filepath.Join(dotGit, "config")
 	trustedConfig, err := os.ReadFile(configPath)
 	if err != nil {
 		t.Fatal(err)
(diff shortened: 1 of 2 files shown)


DaedalusG (approved):

BolajiOlajide (approved):

cbrnrd:
### Merge activity

* **Sep 9, 1:26 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/sourcegraph/src-cli/1371).
* **Sep 9, 1:27 PM UTC**: [Graphite](https://app.graphite.com/github/pr/sourcegraph/src-cli/1371) rebased this pull request as part of a merge.
* **Sep 9, 1:33 PM UTC**: @cbrnrd merged this pull request with [Graphite](https://app.graphite.com/github/pr/sourcegraph/src-cli/1371).

PR #1378, opened 2026-09-18
telemetry: add best-effort usage telemetry recorder

BolajiOlajide:
Closes CPL-1005

src-cli needs a reusable way to report usage without telemetry failures affecting command behavior or output. This adds an internal Telemetry V2 recorder that sends events to the authenticated Sourcegraph instance with a bounded two-second timeout.

Events use the server-required `parameters.version` and numeric, PII-free metadata shape. The recorder relies on authoritative server-side event validation, preserves the existing API client authentication and transport behavior, and uses the HTTP-level client so OAuth failures cannot print interactive guidance into command output. Network, GraphQL, timeout, and unsupported-instance failures remain best-effort and are logged only at debug level.

This PR provides the foundational package only. Command wiring and opt-out or suppression behavior will land separately.

## Test Plan

- `go test ./...`


internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+// GraphQL, timeout, and old-instance failures are silently dropped. It applies
+// its own timeout, so the caller's context need not carry a deadline.
+func (r *Recorder) Record(ctx context.Context, feature, action string, metadata map[string]float64) {
+	_ = r.record(ctx, feature, action, metadata)
+}
+
+// record does the work behind Record and returns any error, so it can be tested
+// directly. Callers outside tests should use Record.
+func (r *Recorder) record(ctx context.Context, feature, action string, metadata map[string]float64) error {
+	if r.client == nil {
```
bobheadxi: is this a real case? why would client being nil be valid?

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// Record sends an event on a best-effort basis. Feature and action identify the
+// event (for example, "srcCli.search" and "succeeded"). Metadata must contain
+// only numeric, PII-free facts. Record never returns an error or panics: network,
+// GraphQL, timeout, and old-instance failures are silently dropped. It applies
+// its own timeout, so the caller's context need not carry a deadline.
+func (r *Recorder) Record(ctx context.Context, feature, action string, metadata map[string]float64) {
+	_ = r.record(ctx, feature, action, metadata)
```
bobheadxi: we should still write errors somewhere, maybe a debug log or osmething

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+const (
+	// clientName identifies src-cli as the source of telemetry events.
+	clientName = "SRC_CLI"
+
+	// eventParametersVersion is the schema version of the metadata we attach to
+	// each event. Bump it when the shape of the metadata changes.
+	eventParametersVersion = 1
```
bobheadxi, line 31: this assumes individual `src` CLI telemetry callsites will never provide their own parameters, is that the case?

I think the telemetry SDK concern should be separated from "telemetry middleware" etc

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+	"time"
+
+	"github.com/sourcegraph/src-cli/internal/api"
+
+	"github.com/sourcegraph/sourcegraph/lib/errors"
+)
+
+const (
+	// clientName identifies src-cli as the source of telemetry events.
+	clientName = "SRC_CLI"
```
bobheadxi: does this match existing conventions for format of client name?

internal/telemetry/telemetry.go
```diff
@@ -0,0 +1,154 @@
...
+
+	// eventParametersVersion is the schema version of the metadata we attach to
+	// each event. Bump it when the shape of the metadata changes.
+	eventParametersVersion = 1
+
+	// defaultTimeout bounds how long a single Record call may spend recording.
+	// It is deliberately short: telemetry is sent synchronously right before the
+	// process exits, so it must not add meaningful latency.
+	defaultTimeout = 2 * time.Second
```
bobheadxi, line 36: can we have telemetry be recorded in the background and have process wait for completion - up to deadline - when it receives a shutdown signal? I think we do something like that in `sg`

internal/telemetry/telemetry.go
BolajiOlajide: I'll switch to `src.cli` for now, pending @akalia25's confirmation.

internal/telemetry/telemetry.go
BolajiOlajide: I mean, not right now, just added as an edge case but can't think of a reason. Removing for now.

internal/telemetry/telemetry.go
BolajiOlajide: Added.

internal/telemetry/telemetry.go
BolajiOlajide: I'll have this as a separate ticket.

internal/telemetry/telemetry.go
BolajiOlajide: This is a required field from the telemetry GraphQL contract. You can check it out here: https://github.com/sourcegraph/sourcegraph/blob/7b2c944fffa65c5e60c35685806f4b985d140637/cmd/frontend/graphqlbackend/telemetry.graphql#L205-L239

akalia25:
@BolajiOlajide could we also support adding `parameters.privateMetadata` here, this would make it consistent with the other telemetry packages we have and allow teammates to pass through approved string data?

akalia25 (approved):
LGTM! Just two notes:
1. is the two second timeout enough
2. can we extend to include sending `privateMetadata` as well?


Thanks for adding this!

internal/telemetry/telemetry.go
BolajiOlajide: https://linear.app/sourcegraph/issue/CPL-1010/recording-coordinator-with-graceful-flush-on-shutdown

BolajiOlajide:
> LGTM! Just two notes:
> 
> 1. is the two second timeout enough
> 2. can we extend to include sending `privateMetadata` as well?
> 
> Thanks for adding this!

1. Yes, the plan is to monitor and adjust, but we can't have the CLI hang on for so long. 2s seem like a good buffer and can adjust once the recorder is hooked up.

2. Yes, On it.

internal/telemetry/telemetry.go
bobheadxi: @BolajiOlajide yes, but in the SDK implementations, individual telemetry callsites provide this parameter to self-describe the callsite. In your implementation, it is a global property for all telemetry callsites

bobheadxi: The docstring on that field indicates that it is meant to be provided on a per-event, not per-integration basis:

>    Version of the event parameters, used for indicating the "shape" of this
>    event's metadata.

internal/telemetry/telemetry.go
bobheadxi: Hm, given synchronous telemetry submission will directly impact the day-to-day usage of `src` CLI as soon as we release this, should we hold a higher bar for the first implementation?

BolajiOlajide: I wanted this to focus on just the client alone. The graceful flush and shutdown are technically next and I have a draft locally alreadt. They touch the internals of the CLI framework, hence why I am pushing to a separate PR so it's easily reviewable.

BolajiOlajide: Also, the recorder isn't hooked up anywhere yet also, thats why I'm separating that also.

